PR #2, opened 2015-05-18
SM Refactor

bakura10:
Hi everyone,

This is my first PR for ZF3. It is a direct backport of my PR on ZF2 repo.

Component is not yet fully tested but coverage should be already quite good.

Let me quickly recap the major changes of this refactor.

## Motivation

The initial motivation was to dramatically increase the performance of this core component, while making it easier to use.

Performance wise, this refactor is dramatically faster, consume less memory (benchmark can be found on ZF2 repo).

### Removal of invokables

The concept of invokables no longer exist. Now, everything is a factory. This simplifies the creation code by having a single main way to create object.

A built-in factory to create invokables has been added.

ZF2 code:

```php
return [
	'invokables' => [
		MyClass::class => MyClass::class
	]
];
```

ZF3 code:

```php
return [
	'factories' => [
		MyClass::class => InvokableFactory::class
	]
]
```

### Factories are invokables

To increase the performance and reduce branching, most interfaces (FactoryInterface, DelegatorFactory, AbstractFactory and Initializer) must be callable by implementing __invoke.

This offer a unified way to create objects and dramatically increase performance be removing a lot of instanceof checks.

### Removal of aliases

Now that this component is 5.5+, we can take advantage of ::class, and we should encourage this usage.

The usage of aliases has been removed from the main service locator where ::class must be encouraged.

For plugin managers, a specialized plugin manager that supports alias has been introduced (`AbstractAliasedPluginManager`) while a more efficient, alias-free `AbstractPluginManager` is still here.

### Ability to map multiple class to same factory

This was previously possible in ZF2 but in a hacky way. Now you can map multiple class names to the same factory, and it will be supported by default in the interface signature.

### Plugin manager change

Previously, a plugin manager had a method `getServiceLocator` to retrieve the main service locator.

This was confusing for new-comers, and could lead to hard to debug error because the constructor didn't neforce the need to pass a service locator as a dependency.

Now, what you receive in the factory is ALWAYS the main service manager, not the plugin manager one.

If you create an object through a plugin manager in ZF2 and want to retrieve a dependency in the main service locator you had to do:

```php
class MyFactory implements FactoryInterface
{
	public function createService(ServiceLocatorInterface $sl)
	{
		$mainSl = $sl->getServiceLocator();
		
		// ...
	}
}
```

In ZF3:

```php
class MyFactory implements FactoryInterface
{
  public function __invoke(ServiceLocatorInterface $sl, $className, array $options = [])
  {
  	// $sl is already the main service locator.
  	// If you want a dependency coming from the same plugin manager, you'll need to fetch it first
  	
  	$pluginManager = $sl->get(MyPluginManager::class)
  }
}
```

### Unified way of having options

In ZF2 it was very hard to have options on factories. This is made easy now, because each interface have a last parameter called options.

Usage:

```php
$object = $sl->get(MyObject::class, ['key' => 'value']);

// Options are passed as part of the options
```

Thanks


GeeH:
To clarify, while the concept of `invokables` have been removed, I don't need to write a factory for classes that have no dependencies and can just be created with `new MyClass()`?

GeeH:
Also, should we take the chance to make this PSR-Container compliant, or is that a waste of time at this point?

juriansluiman:
> The usage of aliases has been removed from the main service locator where ::class must be encouraged.

The thing I really liked about aliases was using it to customize instances which have default services too. An example is [Soflomo\Mail](https://github.com/juriansluiman/Soflomo-Mail/blob/master/config/module.config.php#L53). This pattern is fairly easy to understand, but I cannot grasp how this would work out in the new situation.

Another thing is I already mentioned on Twitter: what about https://github.com/container-interop/container-interop?

Last thing: great work! Really well done :+1:

gianarb:
Perfect in my opinion.. We can remove invokables if I can make an instante of simple service without write an explicit factory class :)

bakura10:
Well not exactly. There is still the need to map it to a factory but a built in factory that basically just do new.

This is necessary if you still want to be able to have all the initializers, delegators... mechanism.

It just removed one key (invokables) and just replace it with a factory.

Envoyé de mon iPhone

test/ServiceManagerTest.php
```diff
@@ -1,1225 +1,122 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
- * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
- * @license   http://framework.zend.com/license/new-bsd New BSD License
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license.
```
samsonasik: hm..., do we change with MIT ?

bakura10:
Oh I was not aware of a PSR about that. Is it actually released ?

Envoyé de mon iPhone

src/Factory/AbstractFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager\Factory;
+
+/**
+ * Interface for an abstract factory
+ *
+ * An abstract factory extends the factory interface, but also have an additional "canCreateService" method,
```
samsonasik: "canCreateServiceWithName"

veewee:
Great improvement!
However, I am also not sure about the alias removal.
Aliases can be very usefull in custom PluginManagers. 
For example a custom TaskManager:
```
$task = $taskManager->get('clear-cache');
$task->run();
```
You can map this directly to a CLI command:
```
public/index.php tasks clear-cache
```

The pluginManager will throw an exception if the task could not be found. You can provide multiple aliases per task to prevent type errors. For example:
cler-cach

I always liked that feature :)

asgrim:
This looks much better overall - although I agree with @juriansluiman about aliases - e.g. on a project I've worked on the internal components referenced the `session` service which was not defined anywhere, but then each individual application was able to define and alias its own `session` service depending on it's own requirements, e.g.:

In a factory for a class in a component that needs to use session:
```php
public function createService(/*...*/)
{
    $session = $this->get('session');
    return new MyClassThatUsesSession($session);
}
```

Then in each application's config:
```php
'factories' => [
    'My\App\Session1' => 'My\App\Session1Factory',
],
'aliases' => [
    'session' => 'My\App\Session1',
],
```

How would we achieve similar flexibility in the "wiring up" of services without using aliases any more?

bakura10:
Actually the aliases feature is much more lightweight in this refactor and should not add too much overhead so I'd be happy to add it again by default inside the SM !

Envoyé de mon iPhone

codeliner:
great work! I agree with all points. First I was with @veewee and @asgrim. But the longer I think about use cases for aliases the more I think it is a good idea to remove the feature.
The main SM should be as fast as possible. For the case mentioned by @veewee the AbstractAliasPluginManager is available and can be used in the CLI context so no need for the main SM to provide this feature.
The session example mentioned by @asgrim can simply continue to use a non FQCN as service name. `session` can then point to `My\App\Session1Factory` or `My\Other\App\Session2Factory`. Using FQCNs as service names should be a suggestion but not a rule, otherwise things like `doctrine.entitymanager.orm_default` would no longer be possible.

stefanotorresi:
> How would we achieve similar flexibility in the "wiring up" of services without using aliases any more?

You can do that even now, using interfaces:

```php
namespace Module;
// ...
interface SessionInterface {}
// ...
$session = $sm->get(Module\SessionInterface::class);
```

```php
namespace App;
use Module;
// ...
class Session implements Module\SessionInterface {}
// ...
'factories' => [
    Module\SessionInterface::class => function(){ return new Session(); };,
],
```

this makes aliases completely useless; in fact, I've been avoiding their usage for sometime now with this very approach.

stefanotorresi:
Now that I think about it, it would be supercool if there was an opt-in check the return type of the factories in `ServiceManager::get()`, so that it either matches the requested FQCN or throws an exception. this would hep preventing configuration ambiguities and make the FQCN actually relevant rather than just a convention.

bakura10:
How would you deal with keys like "doctrine.entitymanager.orm_default" for a type validation?

stefanotorresi:
you don't, because that's not a FQCN. ;)

danizord:
I think it's ok to create another empty class just for the sake of having a type:
```php
class DefaultEntityManager extends EntityManager
{
}

class AnotherEntityManager extends EntityManager
{
}
```

```php
$serviceLocator->get(DefaultEntityManager::class);
$serviceLocator->get(AnotherEntityManager::class);
```

src/AbstractAliasedPluginManager.php
```diff
@@ -0,0 +1,72 @@
...
+namespace Zend\ServiceManager;
+
+/**
+ * Aliased abstract plugin manager
+ *
+ * This plugin manager adds support for alias. This can be useful for some plugin managers where aliases are
+ * simpler to use (like view plugin manager). However, it adds some overhead as it needs to resolve aliases, so
+ * be sure to choose this one only if absolutely needed
+ */
+abstract class AbstractAliasedPluginManager extends AbstractPluginManager
```
danizord: What about composition over inheritance?
```php
class HydratorPluginManagerFactory implements FactoryInterface
{
    public function __invoke(ServiceLocatorInterface $sl, $className, array $options = [])
    {
        return new AliasedPluginManager(new HydratorPluginManager());
    }
}
```

src/ServiceLocatorInterface.php
```diff
@@ -10,24 +10,24 @@
...
      *
-     * @param  string  $name
-     * @throws Exception\ServiceNotFoundException
-     * @return object|array
+     * @param  string $name
+     * @param  array  $options
+     * @return object
      */
-    public function get($name);
+    public function get($name, array $options = []);
```
danizord: Can you remove the optional argument from contract?

src/AbstractAliasedPluginManager.php
stefanotorresi: :+1: 

btw if the point of aliased services is to make people happy, i'd rather have it compose a `ServiceManager` rather than an `AbstractPluginManager`.

src/ServiceLocatorInterface.php
stefanotorresi: :-1: 
why do that @danizord ?

danizord:
@bakura10 Also think you can remove initializers as we can achieve same result with delegators.

asgrim:
:+1: removing initialisers! (possibly relevant: #6068)

src/ServiceLocatorInterface.php
danizord: @stefanotorresi because it makes an "optional" argument actually mandatory.

```php
class MyFactory implements FactoryInterface
{
    // Without $options argument
    public function __invoke(ServiceLocatorInterface $sl, $className)
    {
    }
}
```

Produces:
> Fatal error: Declaration of MyFactory::__invoke() must be compatible with that of FactoryInterface::__invoke()

stefanotorresi: I'm confused, are you talking about `ServiceLocatorInterface` or `FactoryInterface`?

src/Factory/FactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Create an object
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  string                  $className
+     * @param  array                   $options
+     * @return object
+     */
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $className, array $options = []);
```
stefanotorresi: I think `$className` should be renamed to `$requestedName` unless a FQCN is actually granted (which isn't).

stefanotorresi: @asgrim :+1: on `$serviceName`, or whatever, just not `$className` :)

gianarb:
I am puzzled about `$options`.. This is a shortcut to write a bad code or it's a necessary extension?
I know that it's very comfortable :+1:

stefanotorresi:
@gianarb I'm not fully convinced by that either.

since the requested service name is always passed, one could easily do this:

```php
[
    'my_service_options' => 
        SomeService::class => [
            'foo' => 'bar',
        ],
    ],
    'service_manager' => [
        'factories => [
            SomeService::class => function ($sl, $name) {
                $options = $sl->get('config')['my_service_options'][$name];
                return new $name($options);
            }
        ],
    ],
];
```

but I guess the point of @bakura10 proposal is to be able to change these options at the `ServiceManager::get()` call.

codeliner:
@gianarb Good point: $options => potentially bad practice
Whenever I thought it would be nice to pass options to the SM I came to the conclusion that it is better to define application config options and pass them via factory to the service. Passing options to the service directly couples the consumer with the service and I have no chance to later change the options by overriding them via config.
Does anyone has a good example for the use of options?

bakura10:
Yes options is absolutely necessary.

Think for instance about the validators, filters.... They need to be created with options, and if not shared, we must be able to create different instances with different options. It's context specific and that's not the kind of things that should be done through config.

Zf2 support this through an additional interface called MutableOptionsCreation or something like that, that factories must implement. But this is super hacky and I think the options should be enforced in the interface.

There are tons of places where options is vital actually.

Envoyé de mon iPhone

codeliner:
@bakura10: I wasn't aware of the MutableOptionsCreation. Like I said: never had a real use case for that. To be honest I never get validators, filters and the like from the SM only shared services like a form which makes use of them. So for me the context you are talking about is such a form which has a dedicated factory. The factory is responsible for setting up and injecting validators, filters, etc. At least this is the way I handle it, but yeah, that's why I've asked. So thank you to pointing this out.

bakura10:
No you are using it actually but you don't know it.

When you create a filter or validator for an input filter, in a view, in a controller... You are creating those through the plugin managers and it uses this internally. But it is so hacky that it is really ugly in the core code.

And there are several modules where we need options (SlmQueue for instance) where we have to use this ugly MutableCreations interface.

So yes, you are using it :).

Envoyé de mon iPhone

codeliner:
@bakura10 Ok, got it. Sounds like we are talking again about two diffrent purposes like we do with the need of the alias feature.
For shared services you don't need these options. What happens when you pass options to an already instantiated (and shared) service? The SM would need to create a new service or remember the previous used options and compare them, right?
For plugin managers which handle non shared services it is a useful feature. So why not provide a special PluginManager implementation for otptions too?

bakura10:
Nope. Basically it will depends on the shared.

For instance for validators and filters it may makes more sense to not make them shared by default so that asking two times the same service create two different instances.

We cannot really have options only for plugin managers as this is something that needs to be enforced in factories interface and I'd prefer to avoid to have everything in double (FactoryInterface and PluginFactoryInterface for instance).

Envoyé de mon iPhone

src/ServiceManager.php
```diff
@@ -9,1265 +9,266 @@
...
-        if (!($initializer instanceof InitializerInterface || is_callable($initializer))) {
-            if (is_string($initializer)) {
-                $initializer = new $initializer;
-            }
-
-            if (!($initializer instanceof InitializerInterface || is_callable($initializer))) {
-                throw new Exception\InvalidArgumentException('$initializer should be callable.');
-            }
+        // We start by checking if the service is cached (this is the fastest method)
+        if (isset($this->services[$name])) {
```
codeliner: If you want to keep $options I would suggest the following:

```php
if (empty($options) && isset($this->services[$name])) {
```
Meaning a service which relies on userland options should not be shared, otherwise it can cause a lot of trouble. You request a service with a specific option but it was already requested elsewhere with other options...

src/ServiceLocatorInterface.php
danizord: @stefanotorresi both :D

sandrokeil:
@bakura10 @gianarb @codeliner It is possible to add a [AbstractConfigurableFactory](https://github.com/sandrokeil/EasyConfig/blob/master/src/Service/AbstractConfigurableFactory.php "Example of options from config and mandatory parameters") or something similar to add the following features or is this bad?

* Load options from config depending on name (possibility to create an option class)
* Check for mandatory parameters on config loading

It would be great if I can add some features of my [ZF2 module](https://github.com/sandrokeil/EasyConfig "EasyConfig module for Zend Framework 2") to ZF3. This should make our lives easier. ;-)

src/ServiceManager.php
bakura10: Good idea. I'm going to add this.

codeliner:
@sandrokeil :+1: nice module, definitly reduces some boilerplate when working with abstract factories.

roelvanduijnhoven:
Like @juriansluiman and others I effectively use this alias pattern a lot. It makes for reusable code that is easy to understand. So I'm for keeping it!

roelvanduijnhoven:
Another thought. I never mix shared and non-shared services in the same manager. Could this be considered an antipattern? One could argue that a single manager can support only one out of two.

src/Factory/InvokableFactory.php
```diff
@@ -7,16 +7,20 @@
...
 
-namespace ZendTest\ServiceManager\TestAsset;
+namespace Zend\ServiceManager\Factory;
 
-use Zend\ServiceManager\FactoryInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
 
-class ExceptionThrowingFactory implements FactoryInterface
+/**
+ * Factory to create newable objects
```
samsonasik: "newable classes" imo, as in term, object is an instance of class

test/bootstrap.php
```diff
@@ -14,8 +13,8 @@
     $phpUnitVersion = PHPUnit_Runner_Version::id();
     if ('@package_version@' !== $phpUnitVersion && version_compare($phpUnitVersion, '4.0.0', '<')) {
```
samsonasik: based on https://github.com/zendframework/zf2/blob/master/composer.json#L20 , should be 4.6.0

test/bootstrap.php
```diff
@@ -14,8 +13,8 @@
...
         echo 'This version of PHPUnit (' . PHPUnit_Runner_Version::id() . ') is not supported'
-           . ' in the zend-service-manager unit tests. Supported is version 4.0.0 or higher.'
-           . ' See also the CONTRIBUTING.md file in the component root.' . PHP_EOL;
+            . ' in the zend-service-manager unit tests. Supported is version 4.0.0 or higher.'
```
samsonasik: 4.6.0 or higher

composer.json
```diff
@@ -13,22 +13,21 @@
         }
     },
     "require": {
-        "php": ">=5.3.23"
+        "php": ">=5.5"
     },
     "require-dev": {
-        "zendframework/zend-di": "self.version",
+        "phpunit/phpunit": "~4.4",
```
samsonasik: ~4.6 based on https://github.com/zendframework/zf2/blob/master/composer.json#L20

vaclavvanik:
@sandrokeil :+1:

BreiteSeite:
@sandrokeil As much as i like the idea of having something like this, the version you proposed has - in my opinion - some drawbacks:
* The example is wrong. You should not use inheritance for this. The factory creating the service suddenly is responsible for creating 2 objects. The requested service AND the options. This also leads to reduced reusability and duplicated code when the options are needed in other places.
* It's very opinionated in the way it expecteds the config hierachy (module -> scope -> name) which is too inflexible to upstream it in zend-servicemanager IMHO. It's probably better to use a dynamic format (like a dot-seperated string for example).

Just my 2 cents.

sandrokeil:
@BreiteSeite Thank you for your feedback.
> The example is wrong. You should not use inheritance for this. The factory creating the service suddenly is responsible for creating 2 objects. The requested service AND the options. This also leads to reduced reusability and duplicated code when the options are needed in other places.

You're right if an option class is used. This could be solved if we use a service name instead of the option class here. But we need a new kind of factory [AbstractConstructorOptionConfigFactory](https://github.com/sandrokeil/EasyConfig/blob/master/docs/ConstructorOptionConfig.md "Factory example with Constructor Options") also discussed in issue [#7077](https://github.com/zendframework/zf2/issues/7077 "New sub key to configure factories via constructor from config file").

> It's very opinionated in the way it expecteds the config hierachy (module -> scope -> name) which is too inflexible to upstream it in zend-servicemanager IMHO. It's probably better to use a dynamic format (like a dot-seperated string for example).

Yep, you're right. The advantage of this is to see the full configuration in a hierarchy in the config file. But this should not be a must have for ZF3.

The main goal of [EasyConfig](https://github.com/sandrokeil/EasyConfig "Retrieve specific module options and provides some abstract factories to create easily instances depending on configuration") is, to reduce redundant boiler plate code for factories and to have easier usability.

xtreamwayz:
In ServiceMangager::get() there is a check for options. If options are empty it won't load the cached service. 

```php
        if (empty($options) && isset($this->services[$name])) {
            return $this->services[$name];
        }
```

However later on, it does store the created factory regardless options are set.

```php
        if (($this->sharedByDefault && !isset($this->shared[$name]))
            || (isset($this->shared[$name]) && $this->shared[$name])) {
            $this->services[$name] = $object;
        }
```

I might be missing something here but shouldn't only the service without options be cached?

bakura10:
Good catch @xtreamways! I've updated the code.

Maks3w:
We had to rewrite master branch for match release-2.4.2. Could you rebase your work again?

Sorry for the inconvenience

samsonasik:
@Maks3w will we use develop branch for zf3.0 or zf 2.5 ?

weierophinney:
2.5 for now, 3 once we tag 2.5 early next week.

bakura10:
ping @sophpie for your idea we've talked about ?

bakura10:
I've tried to rebase but, as everytime, I fail at doing it. So I will likely open a new PR.

In all cases, here are the current discussion points:

## Aliases

Should we integrate back aliases in the main service locator? I'm definitely not against it, the new architecture is must lightweight so adding aliases support back should have a minimal impact on performance.

## Options

I've integrated a native support for options, because it is often needed (think of validators, filters, queues... that are created through plugin manager but needs options to be configured).

In ZF2 this was possible natively for invokables, but for factories it was complicated: you needed to implement a specific interface, and the options were not part of the interface.

## Removing of "invokables" key

In order to simplify the configuration, this PR removes the invokables key and isntead make invokables just a specialized factory using the built-in InvokableFactory factory. I didn't receive complaints about it so I suspect this do not bother anybody.

## Removal of plugin manager?

Currently, this refactor still keeps the concept of plugin manager. However, while I love the idea of plugin manager, we may start discussing about if we should keep them or not.

### Pros of plugin manager

* While it is bad practice to inject a service manager, injecting a plugin manager is considered okay. Think of injecting a QueuePluginManager as a dependency, or a FilterPluginManager.

### Cons of plugin manager

* It is confusing. People never know exactly if they should add a service into the main service manager, or into a specific other key ("controllers", "input_filters"...). The naming of those top key is not consistent neither.
* In ZF2, when in a factory, the first parameter is actually the plugin manager and not the main service locator. In this refactor, the behaviour has been changed to inject the main service locator instead, but it's still a bit confusing.
* Having plugin manager prevents to do any "automatic" injection, because we can never guess if an instance must be fetched from the main service locator or from a plugin manager.

Maks3w:
bakura create a local branch, cherry pick and force push the the pr branch

sophpie:
@bakura10 , very few time to assign to this taskt hese days. Still working on it.

juriansluiman:
@bakura10 here my opinion:

### Aliases
It is a lightweight feature which allows some useful patterns, so I'd like to keep the feature for the new EVM

### Options
Actually I am against it. Without it, you have this api:

```
$obj1 = $sm->get('foo');
$obj2 = $sm->get('foo');
```

You expect then both variables to hold the same instance.

```
$sm->config('foo')->setShared(false); // fictional API, usually by config
$obj1 = $sm->get('foo');
$obj2 = $sm->get('foo');
```

Now both variables have two different instances of `foo`.

**With options:**

```
$obj1 = $sm->get('foo', ['bar' => 'baz']);
$obj2 = $sm->get('foo', ['bar' => 'bat']);
```
When 'foo' is shared by default, are both variables having the same instance? Is the instance cached with the option config or is the same instance reconfigured for the second `get()`? For non shared services this is clear, but for shared instances this opens the door to quite some bugs or annoyances:

```
// I have access to options in the first call (a factory)?
$obj1 = $sm->get('foo', ['bar' => 'baz']);

// I don't have access to the options now
$obj2 = $sm->get('foo');
```

Say `$obj2` is used in another module, out of the options scope for foo service form the intial module: does this give me the same instance or not?

I think handling options in the API this way is not that straightforward and should be minimized as much as possible.

### Invokables
+1 to have the standard factory which can instantiate all invokables. Do you cache the factory (i.e. fetching two different invokables is done by one factory instance) or does the check/cache outweight the benefit?

### Plugin manager
The ZF2 plugin manager is a fantastic way to scope and verify encapsulated services. The container ensures the fetched objects are always of a certain kind (usually a check for an interface). I haven't seen it in other DICs but onced used, you appreciate the design.

1. The key names are a matter of naming and documentation. The naming adjusted during the zf2 life cycle so it is not consistent (we can break this for zf3) and the docs just need to get this clear right away: per component it must clarify how to declare it (for filters, use "input_filters" e.g.) and there should also be a list of all available keys for the SM

2. The SL argument in the function call should imho be the root SL and not the plugin manager itself. You never need another plugin during the creation of a plugin, but in contrast you need a service from the root SL during creation of the plugin. It avoids the repeated `getServiceLocator()` call where the factory should actually be unaware of this fact.

3. I haven't seen auto injection done yet in the SL for zf2/zf3 so I wouldn't focus on that part until then. If auto wiring will be implemented, it's a matter of configuration how to handle this. If you use FQCN for the service name every time, you could create a separation character to walk down the SL line: `My\Service\PluginManager::My\Plugin\FooBar` (just as an example, please do not discuss this method for this PR, it is just to clarify it _is_ possible).

bakura10:
Thanks @juriansluiman !

## Aliases

Okay, I think I'll likely re-introduce that back to the SM by default.

## Options

Yeah, actually it was suggested above (and I've made changes to the code). When you ask with options, the service is never cached, even if it is shared by default.

I understand this can be confusing and I'm not satisfied 100% with it myself, but we really need ot find a solution for services that should be configured.

## Invokables

For now a new InvokableFactory is created each time. I'll need to benchmark it to see if it's worth while to cache it.

## Plugin Manager

Okay, for the 2) that's how it is now :).

samsonasik:
just for note, as @weierophinney 's comment on https://github.com/zendframework/zend-authentication/pull/2#issuecomment-110043990 , for 3.0.0 we can use develop branch

sandrokeil:
@bakura10 I think we should avoid the `$options` parameter on `get($name, array $options = [])` because of [container-interop](https://github.com/zendframework/zend-servicemanager/pull/4#issuecomment-114731719). Maybe we find another way.

Ocramius:
Given that `get($name, array $options = [])` is not idempotent on different `$options`, so it should really just be something like `createWithOptions()` or such...

GeeH:
"idempotent" is a stupid work

GeeH:
s/work/word

Ocramius:
@GeeH "software development" is hard anyway.

weierophinney:
@bakura10 Could you update the synopsis to reflect the current state of the proposal? I'm getting confused trying to figure out where things are at currently. :)

weierophinney:
> To increase the performance and reduce branching, most interfaces (FactoryInterface, DelegatorFactory, AbstractFactory and Initializer) must be callable by implementing `__invoke`.

Do you mean literally `__invoke()`, or would the code be using the `callable` typehint?

Also, what will the defined behavior be if you call something that has different numbers and/or types of arguments? As an example, `DelegatorFactory` accepts:

```php
function (ServiceLocatorInterface $services, $name, $requestedName, callable $callback)
```

while `AbstractFactoryInterface` accepts the same, minus the `$callback` argument, while `FactoryInterface` accepts just a `ServiceLocatorInterface`, and `InitializerInterface` accepts:

```php
function ($instance, ServiceLocatorInterface $services)
```

Currently in PHP, calling any of these without one of the required arguments, or calling one with an argument of the wrong type will raise an `E_FATAL` (though in PHP 7, we'll be able to catch on `Error`). What's the plan for reporting errors due to mal-registered factories?

Additionally, the `AbstractFactoryInterface` defines two methods, one for creating an instance, but the other for determining if the factory is _capable_ of creating an instance for the named service. How would naming be affected in this case?

Last but not least: what's the migration plan for people who have implemented these interfaces previously? A script can take care of scanning classes implementing the interfaces and then rename the methods, but I'd like to see that script as part of the proposal. As noted in some other comment threads, one of the things I want to avoid is the migration debacle that was ZF1 -> ZF2. Incremental improvements like this proposal are fantastic, but will be all the better if they help define and automate that migration for users.

weierophinney:
> Now, what you receive in the factory is ALWAYS the main service manager, not the plugin manager one.

I like this change. However, again: how do we manage migration? Likely we can do a script that looks for calls to `getServiceLocator()` within various factory implementations, but we definitely need to address how that will happen.

weierophinney:
> In order to simplify the configuration, this PR removes the invokables key and isntead make invokables just a specialized factory using the built-in InvokableFactory factory. I didn't receive complaints about it so I suspect this do not bother anybody.

I'm bothered. :)

This sort of thing needs either (a) automation (i.e., a pre-processor that parses the values and re-maps them to the `InvokableFactory`), or (b) a migration script that developers can run that will transform the entries. One thing to note: the current configuration is a map, so this may break existing applications that were mapping the invokable class to a different service name. That will need to be addressed!

composer.json
```diff
@@ -1,28 +1,38 @@
...
-    "name": "zendframework/zend-service-manager",
-    "description": "Zend\\ServiceManager component",
+    "name": "zendframework/zend-servicemanager",
+    "description": " ",
     "license": "BSD-3-Clause",
     "keywords": [
-        "zf2",
-        "service-manager"
+        "zf3",
+        "servicemanager"
```
weierophinney: Why not have both `servicemanager` and `service-manager` tags? I think it makes sense to have both.

Also, I think we should stop using the `zf{version}` designators, and instead just use `zf` and/or `zendframework`.

composer.json
```diff
@@ -1,28 +1,38 @@
...
     ],
+    "homepage": "https://github.com/zendframework/zend-service-manager",
```
weierophinney: Should be `https://github.com/zendframework/zend-servicemanager`

src/Exception/InvalidArgumentException.php
```diff
@@ -0,0 +1,19 @@
...
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager\Exception;
+
+use InvalidArgumentException as BaseInvalidArgumentException;
```
weierophinney: Let's use `Php` or `PHP` as the prefix, instead of `Base`; that makes the origin more clear.

src/Factory/FactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Create an object
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  string                  $requestedName
+     * @param  array                   $options
+     * @return object
+     */
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $requestedName, array $options = []);
```
weierophinney: This is another reason for a migration script. Currently, factories have exactly one argument, so any current ZF2 factories cannot just have the method renamed to be valid, but must *also* add the additional arguments to their signature.

src/Factory/LazyServiceFactoryFactory.php
```diff
@@ -0,0 +1,61 @@
...
+ * Factory to create a lazy factory
+ */
+class LazyServiceFactoryFactory implements FactoryInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $requestedName, array $options = [])
+    {
+        // @TODO: in ZF3 we should have a FQCN key for config
```
weierophinney: Why? :)

Seriously, requiring a FQCN seems like a massive break for no clear benefit. The Config we currently have in ZF2 is an _array_; if we give it a FQCN, folks will look for a class matching that name. Additionally, it means *every*. *single*. *factory*. that uses configuration in ZF2 cannot be cleanly migrated to ZF3 due to the fact that the well-known Config service is now… no longer well-known.

Again: I'm fine with BC breaks for ZF3. But we really, really, really, really (I can't repeat that word enough) need to keep a constant eye on migration.

src/Exception/InvalidArgumentException.php
danizord: Maybe `Spl` is even more clear?

weierophinney: Sure. Just something to indicate its origin. `Base` is very ambiguous. :)

On Mon, Jun 29, 2015 at 1:40 PM, Daniel Gimenes <notifications@github.com>
wrote:

> In src/Exception/InvalidArgumentException.php
> <https://github.com/zendframework/zend-servicemanager/pull/2#discussion_r33497000>
> :
>
> > @@ -0,0 +1,19 @@
> > +<?php
> > +/**
> > + * Zend Framework (http://framework.zend.com/)
> > + *
> > + * @link      http://github.com/zendframework/zf2 for the canonical source repository
> > + * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
> > + * @license   http://framework.zend.com/license/new-bsd New BSD License
> > + */
> > +
> > +namespace Zend\ServiceManager\Exception;
> > +
> > +use InvalidArgumentException as BaseInvalidArgumentException;
>
> Maybe Spl is even more clear?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/zend-servicemanager/pull/2/files#r33497000>
> .
>



-- 
Matthew Weier O'Phinney
matthew@weierophinney.net
https://mwop.net/

src/Initializer/InitializerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+interface InitializerInterface
+{
+    /**
+     * Initialize the given instance
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  object                  $instance
+     * @return void
+     */
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $instance);
```
weierophinney: Another need for migration strategies: this inverses the current order of arguments.

src/ServiceManager.php
```diff
@@ -0,0 +1,275 @@
...
+    public function get($name, array $options = [])
+    {
+        // We start by checking if the service is cached (this is the fastest method). If options is not empty, we
+        // never "share" because this could create unexpected behaviour
+        if (empty($options) && isset($this->services[$name])) {
+            return $this->services[$name];
+        }
+
+        // Let's create the service by fetching the factory
+        $factory = $this->getFactory($name);
```
weierophinney: Considering `$factory` is not used in the second condition below, shouldn't it be lazy-fetched at the point it's needed?

src/ServiceManager.php
```diff
@@ -0,0 +1,275 @@
...
+            $name
+        ));
+    }
+
+    /**
+     * @param  string $name
+     * @param  array  $options
+     * @return object
+     */
+    protected function createDelegatorFromFactory($name, array $options = [])
```
weierophinney: I'm wondering if this method is mis-named; it's not being passed a factory, it's being passed a name. Perhaps `createDelegatorForName()`?

test/AbstractAliasedPluginManagerTest.php
```diff
@@ -0,0 +1,63 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license.
```
weierophinney: I think @samsonasik pointed this out before, but ZF is licensed as BSD-3-Clause…

test/bootstrap.php
```diff
@@ -4,7 +4,6 @@
  * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
```
weierophinney: Get rid of this file; it's not necessary any longer.

weierophinney:
@bakura10 Overall, a very nice vision, and I'm impressed by how slim the code is! My main concern is migration, and I've outlined the various places where I have specific concerns. The migration scripts can be part of a later pull request, but the specific BC breaks MUST be outlined, either as part of a README or part of a "migration" document (we'll be adding component-specific docs to all component repos soon, and this could be part of that) before acceptance for merge. (If you go the route of a migration document, push it into a `doc/book/` tree.)

Finally, I'd really like to see the PR synopsis updated to show the final status of the PR before we merge. Since we're now using the [keep a changelog](http://keepachangelog.com/) format, for many people, the PR will be the first, most comprehensive place to understand the changes, so having the information up front will help aid comprehension of the changes.

Thanks a ton; again, this is stellar work!

src/Factory/LazyServiceFactoryFactory.php
bakura10: Make sense, I'm okay for keeping Config here.

src/Initializer/InitializerInterface.php
bakura10: I cna change that to match ZF2 implementation, but I just wanted something homogeneous, so that anything factory related had the SL as first parameter.

bakura10:
Hi,

Sorry for the late reply, I won't be able to work on that until next week, but here are a few answers.

> Do you mean literally __invoke(), or would the code be using the callable typehint?

Yes, the idea is that factory (or delegator factory, or abstract factory) are just callable. This allow to remove all the check (no need any more to check if instance of FactoryInterface then call createService...).

> Also, what will the defined behavior be if you call something that has different numbers and/or types of arguments? As an example, DelegatorFactory accepts:

That's not a problem, because you still need to register factories, abstract factories... separately into the SM. So first the SM will iterate through factories, so it knows the expected order (it's part of the interface), then traverses all the delegators, then the abstract factories.

> Additionally, the AbstractFactoryInterface defines two methods, one for creating an instance, but the other for determining if the factory is capable of creating an instance for the named service. How would naming be affected in this case?

The idea is that the creation itself is made through the __invoke. The method that check is still called canCreateServiceWithName. Not really a problem.

> Last but not least: what's the migration plan for people who have implemented these interfaces previously? A script can take care of scanning classes implementing the interfaces and then rename the methods, but I'd like to see that script as part of the proposal. As noted in some other comment threads, one of the things I want to avoid is the migration debacle that was ZF1 -> ZF2. Incremental improvements like this proposal are fantastic, but will be all the better if they help define and automate that migration for users.

For now I didn't write one, but to be honest it's less dramatic than the EVM.

> I like this change. However, again: how do we manage migration? Likely we can do a script that looks for calls to getServiceLocator() within various factory implementations, but we definitely need to address how that will happen.

Same, I didn't write one, and while I can undertsand the need for this I'm really not interested in taking time for writing this. So it will depend on my motivation.

src/ServiceManager.php
bakura10: Yes you're right, it's not needed if instance is created from a delegator, I'll change that! ;)

src/ServiceManager.php
bakura10: Sounds sane! I was asking myself the same thing when I just read my code again.

bakura10:
@weierophinney , I've made some changes, specifically:

* Fix the feedback
* I've added back native support for aliases right into the SM. I could therefore removed the `AbstractAliasedPluginManager` class that I introduced. Honestly, the new alias mechanism is really fast and very lightweight, so it should not be a problem in terms of performance.
* I've added a changelog. It does not contain all the changes but should be a good first step.

I still need to update the tests. The only thing that I'm still highly unsure is the options thing. What is your opinion on this @weierophinney ?

sandrokeil:
> Same, I didn't write one, and while I can undertsand the need for this I'm really not interested in taking time for writing this. So it will depend on my motivation.

@bakura10 @weierophinney You are not alone. If this will be accepted, I think somebody can write a migration tool. That's not the problem. Maybe it could be usefull to write PHPCS sniffs which detects BC breaks?

CHANGELOG.md
```diff
@@ -0,0 +1,195 @@
+# CHANGELOG
```
weierophinney: :+1: Thanks for adding this; I think it will be a fantastic reference.

CHANGELOG.md
```diff
@@ -0,0 +1,195 @@
...
+## v3.O.0
+
+### Changed
+
+v3 of the ServiceManager component is a completely rewritten, more efficient implementation of the service locator
+pattern. It includes a number of breaking changes, that are outlined in this section.
+
+- Peering has been removed. It was a complex and rarely used feature that was misunderstood most of the time.
+
+- Integration with `Zend\Di` has been removed. It may be re-integrated later as part of another component.
```
weierophinney: It can be re-added via an abstract factory fairly easily, no?

bakura10: I think so but as I've never really used Di I prefer not to pronounce :)

bakura10:
This PR now adds support for Interop container. As it will be v3 and allow BC, I was unsure if I shouldn't just remove ServiceLocatorInterface, and make the SM implement ContainerInterface only?

samsonasik:
should related with #4 that merged to develop

weierophinney:
> This PR now adds support for Interop container. As it will be v3 and allow BC, I was unsure if I shouldn't just remove ServiceLocatorInterface, and make the SM implement ContainerInterface only?

You should rebase; I just merged #4 a short bit ago to the develop branch, and it already does this. :)

In terms of `ServiceLocatorInterface`, I think we should keep it; it provides a package-specific context.

bakura10:
Okay, I'll change that tomorrow. It seems that the interop also has exception interface. I've updated this PR but I'm a bit unsure of the hierarchy

bakura10:
But actually i find it a bit strange to keep both interface. If serviceLocatorInterface extends containerInterface it will just have the same methods.  If ServiceManager implement both, it will implement two interfaces that have the exact same content. Isn't it a bit redundant actually?

samsonasik:
rebasing will little bit hard as it should be pointed to develop, the reason it was pointed to master because develop didn't exists yet. Probably the only way is create new PR...

bakura10:
Replaced by #6

PR #3, opened 2015-06-03
use PHP 5.5, remove $self = $this

samsonasik:


samsonasik:
rebased

src/AbstractPluginManager.php
```diff
@@ -63,9 +63,9 @@ public function __construct(ConfigInterface $configuration = null)
     {
         parent::__construct($configuration);
         $self = $this;
```
sandrokeil: Why you don't remove the line `$self = $this`?

samsonasik: done ;)

weierophinney:
Moving to 2.6.0, as I want to do that release now.

PR #4, opened 2015-06-22
Implement container-interop

gianarb:
This proposal is a simple implementation of ContainerInterface.
Now we can use it in [zend-stratigility-skeleton](https://github.com/container-interop/container-interop) for example.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         }
     },
     "require": {
-        "php": ">=5.5"
+        "php": ">=5.5",
+        "container-interop/container-interop": "*"
     },
     "require-dev": {
         "zendframework/zend-di": "~2.5",
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -9,7 +9,9 @@
 
 namespace Zend\ServiceManager;
 
-class ServiceManager implements ServiceLocatorInterface
+use Interop\Container\ContainerInterface;
+
+class ServiceManager implements ServiceLocatorInterface, ContainerInterface
 {
     /**@#+
      * Constants
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -34,6 +34,12 @@ class ServiceManagerTest extends TestCase
         $this->serviceManager = new ServiceManager;
     }
 
+    public function testInteroperabilityInterface()
+    {
+        $serviceManager = new ServiceManager();
+        $this->assertInstanceOf("Interop\Container\ContainerInterface", $serviceManager);
+    }
+
     /**
      * @covers Zend\ServiceManager\ServiceManager::__construct
      */


weierophinney:
My only concern is that the container-interop interfaces were just admitted as a draft PSR... which means there's still a very high liklihood of changes possible. We may want to do this as an "InteropManager" or similar, with documentation in the README indicating it's experimental and unstable until the PSR for container interop is accepted.

gianarb:
I can try to write a `InteropManager` that extends `Zend\ServiceManager\ServiceManager` but extends `Interop\Container\ContainerInterface`

gianarb:
@weierophinney this is a project's draft :) 
https://github.com/gianarb/zend-servicemanager-container-interop

easy easy

weierophinney:
@gianarb That's exactly what I was suggesting!

Pinging @mvriel : is there a phpdoc annotation for indicating a class or method is "unstable"?

gianarb:
In your opinion is a good idea builds a new project? 

:+1: It's clear in my opinion and we can not add unstable class in zend-servicemanager

mvriel:
@weierophinney no, the best you could do is use to mark it using `@access protected` or `@access private`; though that tag is deprecated and removed from PSR-5

sandrokeil:
Looks good. Maybe one problem could be the [optional parameters](https://github.com/zendframework/zend-servicemanager/blob/master/src/ServiceManager.php#L492 "ZF2 ServicerManager"). I know it's not a problem in PHP but it's not so clean. Isn't it?

gianarb:
@sandrokeil the intereface is already implemented without modified.. Why could it be a problem?

sandrokeil:
@gianarb I think it's confusing because the ContainerInterface function definition is `get($id)` and in ServiceManager we have `get($name, $usePeeringServiceManagers = true)`. Also in ZF3 it's planned to have [public function get($name, array $options = [])](https://github.com/bakura10/zend-servicemanager/blob/refactor/src/ServiceManager.php#L94).

The ServiceLocatorInterface is ok, that's right. But it could be changed in [ZF3](https://github.com/bakura10/zend-servicemanager/blob/refactor/src/ServiceLocatorInterface.php#L24). So it's magic to call `$sl->get('service', $options)` or `$sl->get('service', false)`.

So I think it's not real support of interoperability. But we could fix this for ZF3.

Ocramius:
> My only concern is that the container-interop interfaces were just admitted as a draft PSR... 

`container-interop` as-is is a project on its own. It will maybe become a PSR, yes /cc @moufmouf

 > @gianarb I think it's confusing because the `ContainerInterface` function definition is `get($id)` and in ServiceManager we have `get($name, $usePeeringServiceManagers = true)`.

Note that this is perfectly legit PHP

weierophinney:
Based on the feedback from @Ocramius, I'm :+1: - however, it looks like we need to add a little more test coverage, @gianarb - this small change has decreased coverage. :)

weierophinney:
Merged to develop for release with 2.6.0.

composer.json
```diff
@@ -13,7 +13,8 @@
         }
     },
     "require": {
-        "php": ">=5.5"
+        "php": ">=5.5",
+        "container-interop/container-interop": "*"
```
Ocramius: Note: this should be a constrained version (1.0.*)

gianarb: Sorry #5

PR #6, opened 2015-07-09
Refactor v2

bakura10:
Hi,

This replace the other PR as I suck at rebasing.

What do we need next for the PR @weierophinney ? you didn't give us your opinion regarding the options stuff. This is the only important thing that I'm unsure yet.


weierophinney:
@bakura10 I'm finishing up the Apigility 1.2 release in the next day or so, and then will give this another thorough review. I'm torn on the `$options`, to be honest. I introduced them to the plugin managers, because most of our plugin types already accepted an array of options to their constructors, and many needed to vary the instance based on the options. I hesitate to recommend the pattern for application-level services. However, if it's predictable, has reasonable semantics, and is performant, it may be an interesting feature; I'll be sure to pay specific attention to your implementation when I review.

Thanks for all your hard work on driving this forward!

bakura10:
Thanks!

Just to know, I'm trying to run the tests locally (and they actually fail in  Travis for the same reason) because you told me to remove the bootstrap file. How are we supposed to do it now?

weierophinney:
Modify the phpunit.xml file to point to the vendor/autoload.php file for
bootstrapping.

bakura10:
Hi,

I've updated and fixed the tests. I'm a bit lost with all the travis tasks that are run, it seems there is a bit if inconsistency between various ZF components. Couldn't we have a simple template on which tasks should run?

Saeven:
Nice work @bakura10.  I'd started down this same road because I needed options on main SM; looks like you hit the nail on the head, our code wasn't too different - I'd serialized and hashed options in the cache to return same-configuration from cache (e.g., SMTP mailers).

I hope this gets pulled in quickly!  Glad aliases are still in.

bakura10:
Note for myself: in the configure method we must do an array_merge to allow a local configuration that is merged with the configuration given in constructor (mostly needed for plugin managers)

bakura10:
Hi @weierophinney,

I've mode some final changes, for me the code is nearly finished and could be reviewed :). Some parts are still untested (relative to delegators). If someone want to give it a try, do not hesitate to do a PR to my branch.

src/Factory/LazyServiceFactoryFactory.php
```diff
@@ -7,27 +7,24 @@
...
 {
     /**
      * {@inheritDoc}
-     *
-     * @return \Zend\ServiceManager\Proxy\LazyServiceFactory
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $requestedName, array $options = [])
     {
         $config = $serviceLocator->get('Config');
```
bakura10: @weierophinney , I'm still unsure how this is supposed to work now that Service Manager is a standalone component... If someone is using this component without the context of the ZF2 app, this will simply fail.

This gives yet another point for @Maks3w approach to really have one standalone component that can be used for non ZF user (zend-servicemanager) AND an additional module to be consumed by ZF user (zend-servicemanager-module) whose only task will be to integrate with the module manager and provide its config.

This will really going to be a huge challenge to make everything work nicely now that we cannot make any assumption on what is installed or not.

Maks3w:
I started to hate this kind of PR where the diff is fully of dirty changes.

Please @bakura10 Review your own PR file by file, line by line and make a PR clean and focused only in the changes related to the PR.

Also add a resume in the PR description of the changes you made, why are needed, etc.

bakura10:
This PR is cleaned and nearly finished. Once again it's a complete rewrite of the component, I'm not sure how I can do that better. I could do like for validator by doing a first PR that removes everything, then another PR on my branch with only the changes? But I'd prefer if review could be donehere.

Maks3w:
@bakura10 Your PR basically only should have changes in src and test directory. Rest of changes are not related and even are broken.

Again, *REVIEW your own PR*

bakura10:
What do you mean by review my own PR ? I have reveiwed it and I have nothing else to say about it. Why other changes are broken? Changes made to composer.json are part of the PR.

Maks3w:
@bakura10 Do you find normal to change the LICENSE?

bakura10:
Oops, definitely not. I'm not sure why it has been converted to .txt file actually :(.

Maks3w:
@bakura10 That is not the only file. Please last call open https://github.com/zendframework/zend-servicemanager/pull/6/files and clean the changes

froschdesign:
> What do you mean by review my own PR ?

It's hard to review your PR. Too much noise in your code changes. (Separation of concerns / separation of changes is missing in your PR. :wink:)

mnapoli:
I think the diff problem comes from the fact that this PR contains unrelated commits (see the first commits). I guess it's because of a wrong rebase. And I'm afraid that by fixing each of those changes manually in new commits you are only adding more changes.

If you rebase against master properly (and squash/rename some of your commits) it should solve the problems.

mnapoli:
Ah or maybe it's because this PR is against master and not develop?

Maks3w:
see the first commit of the author and your questions will be a answered

jaapio:
Since this is a development step to version 3 this should be merged to develop and not to master. I think that was the last comment of @weierophinney in #2

bakura10:
Yeah, I've been trying agian but the issue is that deleted files still appear in the PR and I'm unable to hide them :/. @Maks3w is currently offline but if anyone knows how to do that I4m on IRC right now (#zftalk.dev) :D

mnapoli:
First thing might be to recreate the pull request again develop (since this branch seems to be based on develop).

bakura10:
I've been able to do it against my branch: https://github.com/bakura10/zend-servicemanager/pull/6

bakura10:
Aren't you on IRC ? :)

bakura10:
Close it in favour of #8

mtymek:
Not sure if this helps, but sometimes when I run into such issues, I create fresh branch, import all changes from "corrupted" one:

`git merge --squash broken-branch`

and then I use `git rm --cached` to get rid of unwanted files.
Unfortunately, this will obviously break your commit history.

Maks3w:
you can do `git reset upstream/develop` this almost the same as rebase and squash but without deal with lots of conflicts

src/Initializer/InitializerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+interface InitializerInterface
+{
+    /**
+     * Initialize the given instance
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  object                  $instance
+     * @return void
+     */
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $instance);
```
snapshotpl: What is a state of this iterface? This thread looks interesting https://github.com/zendframework/zf2/issues/6068

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,55 @@
...
- * Automatically registers an initializer which should be used to verify that
- * a plugin instance is of a valid type. Additionally, allows plugins to accept
- * an array of options for the constructor, which can be used to configure
- * the plugin when retrieved. Finally, enables the allowOverride property by
- * default to allow registering factories, aliases, and invokables to take
- * the place of those provided by the implementing class.
+ * Abstract plugin manager
  */
-abstract class AbstractPluginManager extends ServiceManager implements ServiceLocatorAwareInterface
+abstract class AbstractPluginManager extends ServiceManager implements PluginManagerInterface
```
snapshotpl: What are you think to rename this class to AbstractInstanceOfPluginManager or similar? This is abstract implementation of PluginManager but focused on checking "instance of".

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,55 @@
...
-        $serviceLocator = $this->getServiceLocator();
-
-        if ($serviceLocator && $serviceLocator->has($serviceName)) {
-            throw Exception\ServiceLocatorUsageException::fromInvalidPluginManagerRequestedServiceName(
-                $this,
-                $serviceLocator,
-                $serviceName,
-                $exception
-            );
+        if (empty($this->instanceOf) || $instance instanceof $this->instanceOf) {
```
snapshotpl: I see it like that: in this abstract class exists abstract method `getInstanceOf()` and implementation of `validate()` in this abstract class is `final`

src/AbstractPluginManager.php
weierophinney: Too verbose. Keep it simple, and targetted: it's a specialized SM implementation for managing plugins, *which are typically expected to be of a specific type*. Making the name more verbose does not gain us anything, and makes discovery harder.

src/Initializer/InitializerInterface.php
weierophinney: For BC purposes, we will likely continue to need this, though we may want to mark it as deprecated.

While delegator factories are indeed an awesome feature, they also are opt-in, and in many cases — specifically with the EM — developers are currently trained to just expect the instance to be injected. I cannot think of a migration plan that would ease that transition currently.

src/AbstractPluginManager.php
snapshotpl: What is responsibility of PluginManagerInterface? Validate created objects. Checking `instance of` is a one of implementation. I think that we should keep `AbstractPluginManager` class, but without implementing `validate()` method and add new abstract class like `AbstractInstanceOfPluginManager`.

src/Initializer/InitializerInterface.php
snapshotpl: I like idea of initializers but not as default in zf. Initializers are great feature to use in plugin manager where we can use methods from validated interfaces

bakura10: I'm not sure to understand what you mean by "initializers are great feature to use in plugin manager where we can use methods from validated interfaces". Can you please clarify?

src/AbstractPluginManager.php
bakura10: I'm not sure to understand either. The AbstractPluginManager is made for enforce a given or interface, so there is no need to segregate this one into two different classes.

In the case you want to skip validation, you can always override the validate method in your plugin manager.

src/AbstractPluginManager.php
bakura10: No I don't want validate to be final. There are cases where you'd like to check against two interfaces (in this class the small shortcut introduced in this class is not enough so oyu definitely need to override validate).

bakura10:
@snapshotpl , please review #8 instead, this is the last one.

PR #8, opened 2015-07-21
[ZF3] Refactoring Service Manager

bakura10:
Hi,

This is another PR against develop. But those deleted files are still present in the PR :/.


Maks3w:
This is better but is not enough:

* changelog overwrite 2.6
* composer. json is a mesh
* test/bootstrap must be restored

bakura10:
All those changes (change log, removing test/bootstrap) were asked by Matthew actually. I feel lost u_u

Envoyé de mon iPhone

Maks3w:
@bakura10 tell me the reason because you  have removed each line. Put a note on every one. If you justify the 90 percent I invite you to 1 beer

Maks3w:
@bakura10 Matthew never told you to remove test/bootstrap. He explain you why phpunit is failling and you can workaround the issue changing the phpunit.xml file.

composer.json
```diff
@@ -3,8 +3,9 @@
     "description": " ",
     "license": "BSD-3-Clause",
     "keywords": [
-        "zf2",
-        "servicemanager"
+        "zf",
```
Maks3w: This can be move outside of the PR and merge now if approved

composer.json
```diff
@@ -17,21 +18,19 @@
...
-        "zendframework/zend-mvc": "~2.5",
+        "phpunit/phpunit": "~4.6",
         "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/PHPUnit": "~4.0"
+        "ocramius/proxy-manager": "~1.0"
     },
     "suggest": {
-        "ocramius/proxy-manager": "ProxyManager 0.5.* to handle lazy initialization of services",
-        "zendframework/zend-di": "Zend\\Di component"
+        "ocramius/proxy-manager": "ProxyManager 1.* to handle lazy initialization of services"
```
Maks3w: The change in the description is only applicable to this refactor or can be applied to current master?

bakura10: @Ocramius , were there BC in ProxyManager 1.0 ? Or can we still recommend 0.5.* ?

src/Exception/InvalidArgumentException.php
```diff
@@ -9,6 +9,11 @@
 
 namespace Zend\ServiceManager\Exception;
 
-class InvalidArgumentException extends \InvalidArgumentException implements ExceptionInterface
+use InvalidArgumentException as SplInvalidArgumentException;
+
+/**
+ * InvalidFactoryException
```
Maks3w: Wrong description. Basically this files does not introduce improvements.

CHANGELOG.md
```diff
@@ -2,6 +2,206 @@
...
+* You can now map multiple key names to the same factory. It was previously possible in ZF2 but it was not enforced
+by the `FactoryInterface` interface. Now, this interface receives the `$requestedName` as second parameter.
+
+Example:
+
+```php
+$sm = new \Zend\ServiceManager\ServiceManager([
+  'factories'  => [
+    MyClassA::class => MyFactory::class,
+    MyClassB::class => MyFactory::class
```
Maks3w: Let the second example to be strings so we can to see real values

CHANGELOG.md
```diff
@@ -2,6 +2,206 @@
...
+
+### Added
+
+* You can now map multiple key names to the same factory. It was previously possible in ZF2 but it was not enforced
+by the `FactoryInterface` interface. Now, this interface receives the `$requestedName` as second parameter.
+
+Example:
+
+```php
+$sm = new \Zend\ServiceManager\ServiceManager([
```
Maks3w: I preffer to remove the global namespace (leading backslash) in examples. Improves readibility

CHANGELOG.md
```diff
@@ -2,6 +2,206 @@
...
+  'factories'  => [
```
Maks3w: Respect CS rules (4 spaces indentation)

CHANGELOG.md
Maks3w: trailing comma too

CHANGELOG.md
```diff
@@ -2,6 +2,206 @@
...
+
+* `invokables` key no longer exists. It has been replaced by a built-in factory.
+
+In ZF 2.x:
+
+```php
+return [
+  'service_manager' => [
+    'invokables' => [
+      MyClass::class => MyClass:class
```
Maks3w: wrong syntax, expected `::`

CHANGELOG.md
```diff
@@ -2,6 +2,206 @@
...
+];
+```
+
+In ZF 3.x:
+
+```php
+return [
+  'service_manager' => [
+    'factories' => [
+      MyClass::class => \Zend\ServiceManager\Factory\InvokableFactory:class
```
Maks3w: In this case is better to wrap with string quotes the FQDN. Also the operator is wrong, expected `::`

src/Factory/DelegatorFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * A factory that creates delegates of a given service
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  string                  $name
+     * @param  callable                $callback
+     * @param  array                   $options
+     * @return object
+     */
```
Maks3w: What exceptions are expected to thrown? Is this method 100% safe?

src/Factory/FactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+interface FactoryInterface
+{
+    /**
+     * Create an object
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  string                  $requestedName
+     * @param  array                   $options
+     * @return object
+     */
```
Maks3w: What exceptions are expected to thrown? Is this method 100% safe?

src/Initializer/InitializerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+interface InitializerInterface
+{
+    /**
+     * Initialize the given instance
+     *
+     * @param  ServiceLocatorInterface $serviceLocator
+     * @param  object                  $instance
+     * @return void
+     */
```
Maks3w: What exceptions are expected to thrown? Is this method 100% safe?

src/PluginManagerInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * A plugin manager is a specialized service locator used to create homogeneous objects
+ */
+interface PluginManagerInterface extends ServiceLocatorInterface
+{
+    /**
+     * Validate an instance
+     *
+     * @param  object $instance
+     * @return void
+     * @throws InvalidServiceException
```
Maks3w: Please detail under what circunstances the exception is expected to throw

src/ServiceLocatorInterface.php
```diff
@@ -10,23 +10,23 @@
...
     /**
-     * Retrieve a registered instance
+     * Retrieve a service by its name, with optional options
      *
-     * @param  string  $name
-     * @throws Exception\ServiceNotFoundException
-     * @return object|array
+     * @param  string $name
+     * @param  array  $options
+     * @return object
```
Maks3w: What exceptions are expected to thrown? Is this method 100% safe?

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,55 @@
...
-     * @var mixed
-     */
-    protected $creationOptions = null;
-
-    /**
-     * The main service locator
+     * An object type that the created instance must be instanced of
      *
-     * @var ServiceLocatorInterface
+     * @var string
```
Maks3w: `null|string`

Maks3w: I'm not a fan of protected properties with null values.

Xerkus:
Several comments regarding this implementation:
I don't like options as part of service locator interface. It implies that service exists, even if it is transparently created behind the scene.
This addition makes certain edge cases easier to work with but open way for multiple contradicting bug-prone behaviors.
Service locator implementation does not use existing service acting strictly as a factory when options are supplied. It is not something implied by contract and goes directly against service locator pattern.
Injecting configured service will have no effect on fetching service with options provided, resulting in unexpected behavior.
Factory implementation gets volatile due to need to support all possible options. Swapping factories have high chance of introducing subtle integration bugs. I believe options must be linked to concrete factory and defined along with it in service manager configuration. Assuming options are factory options, not service constructor options.
Maintenance load associated with using `zend-servicemanager` is increased due to multiple (possible) conflicting usage scenarios. Swapping factory as black box is no longer enough.

Maks3w:
Why do you want all classes to implement `__invoke`

src/Factory/FactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+    public function __invoke(ServiceLocatorInterface $serviceLocator, $requestedName, array $options = []);
```
Xerkus: Non-abstract factory does not need to know or care for requested name.

CHANGELOG.md
```diff
@@ -2,6 +2,206 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
+## v3.0.0 - TBD
+
+### Added
+
+* You can now map multiple key names to the same factory. It was previously possible in ZF2 but it was not enforced
+by the `FactoryInterface` interface. Now, this interface receives the `$requestedName` as second parameter.
```
Xerkus: Isn't that definition of abstract factory?

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,55 @@
...
-     *
-     * Add a default initializer to ensure the plugin is valid after instance
-     * creation.
-     *
-     * @param null|ConfigInterface $configuration
+     * @param ServiceLocatorInterface $parentLocator
+     * @param array                   $config
      */
-    public function __construct(ConfigInterface $configuration = null)
+    public function __construct(ServiceLocatorInterface $parentLocator, array $config)
```
Xerkus: This change makes parent service locator required for plugin manager. It is not always desired behavior.

Xerkus: I don't agree here with changing config DTO to plain array. Config object is responsible for providing defaults, basic configuration validation and preventing typo induced bugs.

bakura10:
@Maks3w the __invoke allows to do interesting optimizations as we do not need to have multiple conditions (check if implement interface, then check if callable...). Now the creation is unified around the callable. You can see in the code how it has simplified a lot of things.

src/AbstractPluginManager.php
bakura10: Actually the old Config object did not do any of them iirc :). It was just an additional layer with very few benefits. The current approahc also does merging and basic configuration validation but is easier to use (the number of times where I simply wanted to just pass an array to configure a plugin manager...).

Maks3w:
callable can still having callable typehint, just use the format `[$object, 'method']`

Maks3w:
because callbable can be anything and not just FooInterface then does not have sense to enforce the implementation of the method.

bakura10:
You're right, the whole implementation could simply remove the interface. But I find it really useful to have it (implements the interface, and IDE automatically fill the missing methods with the correct signature, and parameter...).

CHANGELOG.md
bakura10: No, the idea is a bit different. Actually this is not really new, as ZF2 factories allowed since 2.1 or 2.2 to receive the name of the service, but it was not enforced in interfaces. Now it is enforced to be clearer.

The difference is that you can use the same factory for multiple services when you know in advance that factory can be used to create multiple services. This way there is no performance cost introduced by this: the lookup between the service name and its factory is immediate. But you still need to have access to the $requestedName to create the appropraite instance.

Abstract factory is different: you do not know in advance all the services that can be created with this abstract factory.

src/Factory/FactoryInterface.php
bakura10: See my other comment (that's not true and as I said ZF2 actually pass you the $requestedName as the last parameter because it's useful, it is just not enforced in interface currently, so most people are not aware it's available).

Xerkus: I see where it can be used, eg logger factory to lookup logger configuration by requested service name. It sounds useful but at the same somewhat dirty, convention over configuration.

bakura10: There are actually a lot of use cases. Sometimes I have several jobs that have the exact same dependencies (a strong relationship between them), all of them having the same dependencies. In this case, when I know in advance those classes, I can use this feature and map all those services to the same factory. Of course an abstract factory could do the work too but I like the idea of keeping abstract factories for things "I don't know in advance".

I'm not sure to mean what you mean by dirty? You still need to map each of your services in the config:

```php
return [
    'factories' => [
      Service1::class => Factory::class,
      Service2::class => Factory::class
   ]
]
```

Still has to be de configured explicitely.

Xerkus:
Regarding get() with options, i suppose it would be cleaner to introduce new interface, that will define createWithOptions($serviceName, array $options) instead?
That way it can be implemented only where it is needed for edge cases, such as in validator plugin manager.

bakura10:
This is a bit what we have currently. As of today, there is an obscure interface called MutableOptionsInterface (defined in ZF 2.2 iirc), that you must implement in your factories. Then you can use a method called "getOptions" that will allow to retrieve you the options. But this only work for factories and is super confusing to use (will my proposal make it unified and work for factories, delegators...)

CHANGELOG.md
bakura10: Not sure to understand what you mean by "see real values by using strings". The real value is to show that we can map two different services to the same service, which is shown by this example.

Maks3w: Put an example of string values i.e. 'MyClass', 'MyFactory'  so any one knows what is the equivalence of ::class

src/Factory/DelegatorFactoryInterface.php
bakura10: We cannot know as the delegator is done in user land. Should I write @return \Exception to assume that "any exception may be thrown"?

src/Factory/FactoryInterface.php
bakura10: Same.

src/Initializer/InitializerInterface.php
bakura10: Same.

src/AbstractPluginManager.php
bakura10: By default, the plugin manager does not do any check (if it stays to null). This is means to be overriden by concrete implementation of plugin manager. This is really just a shortcut to avoid write the complete `validate` method as in ZF2, but if this really bothers you I could simply remove this shortcut (one nice side is that side wya it kinda enforces that the `InvalidServiceException` is thrown.

In ZF2, because each people had to write their own validate method, there was no consensus on which exception to thrown (sometimes RuntimeException, sometimes InvalidArgumentException...).

src/Factory/DelegatorFactoryInterface.php
Maks3w: Are you not throwing ServiceNotFoundException in the implementation?

bakura10: No, it's ServiceManager that may throw this exception, not the delegator itself :).

roelvanduijnhoven:
I'm with @Xerkus on options being a bad idea. If you want this then simply create a _real_ factory class and register it as a service. So maybe we should drop `MutableOptionsInterface` instead. That would mean that ZF2 would have to migrate away from using it: but I guess that would be for the better.

bakura10:
Hi @roelvanduijnhoven , I'm not sure to understand the alternative you're suggesting. The options is meant to be used with get. Let me show you a simple example: you want a validator and give some options:

```php
$validator = $validatorPluginManager->get(StringLength::class, ['min' => 5]);
```

I don't see how creating "real factory" would solve this issue. Are you suggesting adding one service per variant? (so having a StringLengthWithMin50, StringLengthWithMin50AndMax100, String...). This is clearly not manageable (and what about when you need to create them at runtime?).

You could argue that I could simply set the parameters using setters:

```php
$validator = $validatorPluginManager->get(StringLength::class);
$validator->setMin(50);
```

But the issue is that if Validators become stateless, we'll likely want to inject parameters at construction and not being able to change them (and actually, that's the case for a lot of immutable services).

Honestly, I can understand that options is not the most beautiful way, but until someone suggests a simple solution that does not involve creating 50 factories for the same service, that definitely sounds the simplest and sanest idea.

roelvanduijnhoven:
By adding the `$options` argument you turn the service manager into a _factory_ manager. You now decide at runtime what the requested services looks like. Defining what a service is should be the task of the service manager, right?

Secondly, there is a problem with semantics. The get call should be idempotent (or be renamed). Thus calling `get('class', ['min' => 3])` twice would yield the same instance. However calling `get('class', ['min' => 3])` and `get('class', ['min' => 4])` would yield two different instances. This behaviour is strange and not understandable to the user.

If you want a factory for validators why not simply do it like so: 
```php
$v = $sl->get('ValidatorFactory')->build(StringLength::class, ['min' => 5])
```

But let us indeed turn to the problem. I am in no way an expert with the validator component as is, thus not sure if this is the current ZF behaviour. Storing multiple instances of `StringLength` with some other min and max values does not seem a good idea to me! But isn't this a problem of the validator component?

bakura10:
I definitely agree with all your points! Once again, this was really a solution I’m not 100% proud but that solve some use cases I already had.

However I definitely like the idea of the ValidatorFactory. If we can introduce that sort of factory to every component that need that, it should definitely solve the issue in a much more elegant way. Your point @weierophinney?

Maks3w:
I think if you need to pass options when build then you are coupled to specific implementation.

Basically it makes a proxy to the constructor method.

In resume I don't see the difference between ask to the service manager to call a constructor with X arguments or instante the class dire using "new".


`$serviceManager->get('Foo')->build(arguments) === new Foo(arguments)`

bakura10:
After some thoughts, I'm convinced to remove the options. This will simplify the code even more and remove any confusion of services not cached because of options.

@Maks3w : what if Foo has required dependencies coming from the service locator AND additional options that need to be passed at creation? This is where the "new" approach does not really work.

weierophinney:
> But let us indeed turn to the problem. I am in no way an expert with the validator component as is, thus not sure if this is the current ZF behaviour. Storing multiple instances of StringLength with some other min and max values does not seem a good idea to me! But isn't this a problem of the validator component?

`<snark>`I just love how people chime in on things they admit to not understanding.`</snark>`

Please take the time, then, to understand the use cases and current behavior, before suggesting changes that invalidate these use cases.

Now, for some explanations and background.

The `AbstractPluginManager` was created to handle several specific use cases:

- Creation of instances that are all of the same basic type (usually, instances of a specific interface).
- In the case of filters and validators, due to the fact that their behavior differs based on the options they encompass, we wanted the ability to return *discrete* instances of the *same* type. Per the examples above, you may use the `Between` validator in multiple inputs of the same form, once with a configuration of `min = 5, max = 10`, another time with `min = 100, max = 1000`, etc; these will validate completely different sets of input, and, as such, need to be discrete instances.

As such, the `AbstractPluginManager` has a toggle to determine whether it returns the same instance or a new instance by default. Additionally, it allows passing `$options` as a second parameter to `get()` to configure the instance being retrieved, *if the factory that will return the instance understands it.* By default, all filters and validators accept an array of options to the constructor, so "invokables" will be passed those at instantiation; several factories provided *also* understand them. Others do not. The fact is: the `$options` passed are not a guarantee, but official plugins we provide support the concept in order to be internally consistent. Is it a good pattern? Maybe, maybe not, but it's practical, and can already support the idea of stateless validators, which is a goal for the future.

Related, the ServiceManager allows determining on a service-by-service basis whether the instance is shared or not; if not, a *new instance* is returned on each retrieval of that service. One specific example of where this facility is used is with events; `EventManager` instances are not shared, while the `SharedEventManager` *is*. This allows per-instance EM instances, each with the shared SEM instance. (Please, let's not debate the practice here; I'm using it only to illustrate how the service manager works.)

What I'm trying to say is that it's not as simple as this statement might appear:

> By adding the $options argument you turn the service manager into a factory manager. You now decide at runtime what the requested services looks like. Defining what a service is should be the task of the service manager, right?

Or this one:

> I think if you need to pass options when build then you are coupled to specific implementation.
> 
> Basically it makes a proxy to the constructor method.
> 
> In resume I don't see the difference between ask to the service manager to call a constructor with X arguments or instante the class using "new".
> 
> $serviceManager->get('Foo')->build(arguments) == new Foo(arguments)

(BTW, this latter is completely ignoring the fact that we may *not* just be passing the options directly to the constructor! In some cases, the options are processed by a factory in order to pull other services or determine which setters to call on the instance before returning it!)

The *context* in which a service manager exists matters. For plugins, being able to pass options makes a lot of sense, as multiple instances of a plugin, each with different state, are expected. Defining a service per instance type gets very complicated very quickly, if nothing else because of names!

Next, recall that *not every plugin is an invokable*! One big reason for using the `AbstractPluginManager` was to allow the same suite of options for plugin creation as we have for general service creation: factories, abstract factories, delegators, etc. This provides a ton of consistency to the framework: if it looks, acts, or smells like a plugin of some sort, then a plugin manager exists for it, and, since you know how to configure the service manager, you now also know how to configure new plugins for the component in question. It also simplifies *consumers*; they get the plugin manager and pull from it in a consistent way, without needing to know how it works.

---

Moving from plugins to the ServiceManager…

*Unshared* services are a desirable feature. Most IoC containers *do* support this idea, because there are services that you want to be able to inject into other services, but as multiple discrete instances, even if the initial configuration is the same. Is it necessarily a factory? Does it matter? IoC does not *require* that every retrieval of a given service returns the same instance.

As a new point: one of the goals @bakura10 has made is to reduce the number of factories developers need to build. Right now, because of the 1:1 relationship between a service and a factory (because factories do not know what service name was requested), we end up with a lot of factories that are essentially cut-and-paste code. Those factories are redundant, but also further slow the application due to I/O lookups and memory overhead (the single biggest factor in performance of ZF2 applications I've seen is number of objects; objects take up memory, but also mean filestat calls). My impression is that @bakura10 was using `$options` as one way to solve this problem.

---

So, with this background in mind, the question is: is there another way to accomplish the same goals, *particularly with regards to having a clean, easy to accomplish, upgrade path for end-users*?

This looks promising:

> If you want a factory for validators why not simply do it like so:
> 
> ```php
> $v = $sl->get('ValidatorFactory')->build(StringLength::class, ['min' => 5])
> ```

Considering that the majority of places that plugins are retrieved are inside other factories or in proxy/facade methods governed by the framework (e.g., `Zend\View\Placeholder::plugin()`, `Zend\Mvc\Controller\AbstractController::plugin()`), this would be a relatively easy migration. I'd actually suggest having such a factory use the `get()` method as well, as that will allow seamless migration in most cases:

```php
$serverUrl = $services->get('HelperPluginManager')->get(ServerUrl::class, $options);
```

Note, however, that in order to facilitate the same spectrum of use cases that we currently have, this means that the `AbstractPluginManager` implementation needs to support the same features as the ServiceManager, which leaves us in exactly the position we are now!

I'll sum up:

- A key feature of plugin managers is that they support the same creation mechanisms as service managers. This provides consistency within the framework, and simplifies learning (learn the system once, apply everywhere).
- Plugin managers *often* need to support *discrete instances* based on the options provided *at runtime*. As such, they often act more like factories.
- Adding `$options` to the application-level service manager may not be necessary or desired, but my understanding is that it's in the current incarnation in order to facilitate re-use for plugin managers and/or the ability to use the same factory for multiple services.

I appreciate the desire for theoretical purity, but we also have to be pragmatic. Having something re-usable in multiple contexts and easily understood are important goals as well. Additionally, as I've been pointing out everywhere lately, we **must** consider the migration impact of changes. Yes, we're talking about a new major version, but "break all the things!" is bad for adoption. If the migration will not be easily accomplished, people will not migrate.

Please read the above and understand the use cases. @bakura10 — you're doing great work.

Maks3w:
@weierophinney Thanks for your comment, I think you have hit in one important point. PRs like this are a bunch of lines of code where is not clear what are the use cases. It's needed document the PRs with clear list of use cases.

My comment is related to the discussion thread along with the recent (BC Break) change in StaticValidator where are cases of constructor coupling.


I

bakura10:
Thanks @weierophinney for the lengthy explanation.

> As such, the AbstractPluginManager has a toggle to determine whether it returns the same instance or a new instance by default. Additionally, it allows passing $options as a second parameter to get() to configure the instance being retrieved, if the factory that will return the instance understands it.

This has to be noted that as pointed before this can has the side-effect of NOT caching a service, even if a factory does not understand options (or, more accurately, does not use them, as factory interface now contains options in the signature to avoid the confusing MutableCreationOptions interface, as theorically now all factories understand options).

The basic rule is that once you give a non-empty option, caching will always be bypassed, no matter how the manager is configured.

> Related, the ServiceManager allows determining on a service-by-service basis whether the instance is shared or not; if not, a new instance is returned on each retrieval of that service. One specific example of where this facility is used is with events;

This is not really true in my implementation: once options is given, it won't be cache, no matter how the service manager is configured.

So... are we good to go ? :) I'll try to find some time for writing documentation in addition of the changelog, but may that can be done after?

bakura10:
Any news on that @weierophinney ?

weierophinney:
I'd hoped to start formal review and documentation yesterday, but to many
other tasks; should be able to start early next week, and get it merged to
develop.

src/Factory/FactoryInterface.php
weierophinney: @Xerkus I hadn't realized that the `FactoryInterface` already defines a `$requestedName` parameter until first reading through this PR last month. Since then, I've already found several examples of where I can use it.

In a way, I like it better than `AbstractFactory`, as it alllows you to know *exactly* which factory is handling creation of a requested service; as it is right now, I could potentially slipstream in a different `AbstractFactory` implementation that indicates it can create an instance, but then creates something completely different than expected.

As such, I totally support this feature; I think it's quite useful, and promotes better, more predictable practices.

src/AbstractPluginManager.php
weierophinney: While I get the idea of config objects, I admit to thinking they're largely unnecessary. I tend to prefer either being able to pass configuration directly to the instance, or using a factory to create an instance from configuration.

Config objects tend to be simply "middlemen"; I like to remove the middleman when possible.

As an example, this is a current pattern:

```php
use Zend\ServiceManager\Config;

$container = new ServiceManager(new Config(include 'config.php'));
```

I have to import two classes, and remember to pass my configuration to a config object.

Simpler, from a DX perspective, is one of the following:

```php

$container = new ServiceManager(include 'config.php');

// or
use Zend\ServiceManager\ContainerFactory;

$container = ContainerFactory::fromConfigFile('config.php');
```

These are far simpler to understand and consume.

The only real benefits of DTO-style Config objects are:

- Simplification of the configurable object. It does not need logic for validating incoming configuration. This can often be accomplished with an internal class, however. (I.e., you might keep something like the Config object, but it's something that would never be publicly consumed, only consumed internally).
- Programmatic configuration; i.e., using the interface the DTO provides to setup configuration before configuring the main object. However, I have yet to see somebody do this in actual production code. The standard use case at this time is the first example I posted above.

So, my recommendation: ditch the DTO *except potentially as an internal detail*.

weierophinney:
> This is not really true in my implementation: once options is given, it won't be cache, no matter how the service manager is configured.

This makes absolute sense to me. If you provide `$options`, the implicit agreement is you want a new instance, regardless of whether or not they are actually used. I consider that consistent behavior, as it sets up a consistent expectation.

weierophinney:
I've gone and re-reviewed old comments and new ones, and have also been researching other implementations recently as we worked on zend-expressive. As such, I have a little bit of feedback to add at this point about options.

Options are typically useful only for plugin managers, and particularly when you allow or require new instances each time an instance is retrieved from it.

In many ways, it's analogous to what Aura.Di terms [factories](https://github.com/auraphp/Aura.Di/blob/3.x/docs/factories.md). This feature in Aura.Di has a separate method for instance retrieval, `$container->newFactory('NameOfService')`. When using that feature, you can pass items to the factory to pass on to the underlying constructor; they're passed as if you were calling `ReflectionClass::newInstanceArgs()`; in other words, sequentially. The pattern would work very well with our plugin managers.

To recap:

- Have a separate "factory" feature for factories that create new instances and which can accept arguments/options to use during instance creation. *This could be isolated to a plugin manager interface/implementation.*
  - This could be as simple as passing the class name, and using its constructor.
  - Or allow passing any callable, and invoking the callable with the arguments provided at runtime.
- Plugin managers would utilize this almost exclusively (if not completely exclusively). Instead of using the current `get()` method, we would use a `factory()` or similar method. This would pose a small challenge for people upgrading, but in most cases would be masked by the framework or established framework components (as the bulk of plugin retrieval/invocation is done via proxy methods).

@bakura10 Does the above sound good to you? It would resolve some of the issues raised on the thread (specifically reservations about having `$options` part of the `get()` signature), and create a clear separation between a standard container and a plugin manager.

CHANGELOG.md
```diff
@@ -2,6 +2,212 @@
...
+In ZF 3.x:
+
+```php
+$sm = new Zend\ServiceManager\ServiceManager([
+    'factories'  => [...]
+]);
+```
+
+* Service manager is now immutable. Once configured, it cannot be altered. You need to create a new service manager
+if you need to change the configuration. This allow to ensure safer and more aggressive caching.
```
weierophinney: I like this a lot. It's more intuitive than the current `$allowOverride` system, and makes it explicit when you are changing its state.

One question: does the implementation contain mechanisms for one or more of the following:

- pulling configured factories and whatnot from the current instance, in a format that can be passed to another instance; **and/or**
- a method for cloning an existing instance *with* provided configuration.

The latter would be similar to what we did in PSR-7, and might look like this:

```php
$newContainer = $container->withConfig([ 'factories' => /* ... */]);
```

People *will* want to be able to do this on occasion; it's our job to make sure it's possible and documented. :smile:

CHANGELOG.md
```diff
@@ -2,6 +2,212 @@
...
+
+* Interfaces for `FactoryInterface`, `DelegatorFactoryInterface` and `AbstractFactoryInterface` have changed. Now,
+they are all callable. This allow to optimize performance. Most of the time, rewriting a factory to match new interface
+implies replacing the method name by `__invoke`.
```
weierophinney: This is how I've typically taught creation of everything but abstract factories, and a lot of examples I come across do this already as well. Good choice! The main problem will be signatures for those methods, now. :smile:

CHANGELOG.md
```diff
@@ -2,6 +2,212 @@
...
+    }
+}
+```
+
+The equivalent ZF 3.x factory:
+
+```php
+class MyFactory implements FactoryInterface
+{
+    function __invoke(ServiceLocatorInterface $sl, $requestedName)
```
weierophinney: I'd point out that the "canonical name" argument is dropped from the interfaces. That's the one that will invalidate most existing factories (and also the one argument nobody ever, ever used).

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,55 @@
...
-
-    /**
-     * The main service locator
+     * An object type that the created instance must be instanced of
      *
-     * @var ServiceLocatorInterface
+     * @var null|string
      */
-    protected $serviceLocator;
+    In $instanceOf = null;
```
weierophinney: `s/In/private/`

src/Exception/ExceptionInterface.php
```diff
@@ -9,6 +9,8 @@
 
 namespace Zend\ServiceManager\Exception;
 
-interface ExceptionInterface
+use Interop\Container\Exception\ContainerException;
+
+interface ExceptionInterface extends ContainerException
```
weierophinney: Nice - ran into this myself recently when playing with container-interop on zend-expressive. If all exceptions already implement the `ContainerException`, it means we'll be able to remove code in there eventually!

weierophinney:
@bakura10 Now that I've gone and reviewed the code directly, I'm in a quandary about the `$options` and plugin managers aspect again. If we take the route I suggested in my [above comment](#issuecomment-136864964), you would need to likely:

- remove `array $options = []` from the `ServiceLocatorInterface`.
- remove the `array $options = []` argument from all factories
- potentially re-add a `MutableOptionsInterface` that `AbstractPluginManager` can implement.
- alter `AbstractPluginManager` to remove implementation of `ServiceLocatorInterface` and/or implement any new interfaces; it could no longer extend `ServiceManager`.
- create additional factory types that provide the `array $options = []` argument

In other words, a lot of work, adding a lot more complexity (as you essentially have two systems to test now, one with options and one without).

My only real concern at this point with having `$options` in the interface and the implementation is conflicts with the container-interop interface. That said, I don't see that interface changing the `get()` signature, and adding optional arguments to an implementation and/or interface extension does not currently cause conflicts in any known PHP versions.

As such, I'm in favor of keeping `$options` in the main implementation. It keeps the code leaner, reduces complexity and relationships between the implementations, and will be more consistent in usage.

The only questions I have remaining are those I made in line comments today, specifically around container immutability. If you can address those, I can do a final review for merge.

Thanks, @bakura10 !!!!

mnapoli:
> and adding optional arguments to an implementation and/or interface extension does not currently cause conflicts in any known PHP versions.

It's something other containers do (https://gist.github.com/mnapoli/6159681), we saw that when discussing the ContainerInterface. Since, as you said, it's perfectly valid code to have more arguments in the implementation (it satisfies the interface), it was considered OK (the interface wouldn't have to contain it, but implementations were free to do it). So I'm just chiming in to +1 what you said.

Also want to point out another approach you mentioned (if I understood you correctly). Just like Aura, PHP-DI has a separate method (`make()`) for creating new instances with provided parameters (missing parameters are resolved by the container). It's defined in a separate interface ([`FactoryInterface`](https://github.com/PHP-DI/PHP-DI/blob/master/src/DI/FactoryInterface.php#L33)) so that it can be type-hinted without being too coupled to the container itself. I'd love to see such an interface be added to container-interop at some point (and PSR eventually but we are still far from it): containers would be free to implement it or not, but containers that do (many) would at least have an interface for it. And since it's a different use case than `get()` IMO it's fine to have a separate interface/method.

src/Factory/FactoryInterface.php
Xerkus: @weierophinney Yes, i discussed that part with @bakura10 and agreed it indeed makes a lot of sense.

bakura10:
I'd say that we should keep the same services and not clone them. That would make the withConfig extremely expensive is there are a lot of services.

weierophinney:
Per a conversation with @bakura10 in IRC, `get()` *does* use `resolveAlias($name)`, just at an earlier stage in the method not shown by the diff.

src/Factory/AbstractFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager\Factory;
+
+/**
+ * Interface for an abstract factory
+ *
+ * An abstract factory extends the factory interface, but also have an additional "canCreateService" method,
```
samsonasik: s/have/has

PR #9, opened 2015-09-02
[ZF3] Refactoring Service Manager

weierophinney:
This is a rebase of #8 against develop, intended for v3.0. Please see that PR for relevant discussion; we can now start testing and refining against this PR, however.

To test, fetch my branch:

```bash
$ git remote add weierophinney git://github.com/weierophinney/zend-servicemanager
$ git fetch weierophinney
$ git checkout -b refactor-sm weierophinney/refactor-sm
```

## Remaining TODO

- [x] Migration guide
  - [x] How does configuration differ? (mostly: invokables, lazy services)
  - [x] How does the consumable API differ? (no ability to `set` or `add` factories)
  - [x] What interfaces have changed?
    - [x] FactoryInterface
    - [x] AbstractFactoryInterface
    - [x] InitializerInterface
    - [x] DelegatorFactoryInterface
    - [x] ServiceLocatorInterface (extends ContainerInterface, adds `build()`)
  - [x] How are plugin managers different?
    - [x] New interface
    - [x] Abstract class has reasonable validation by default, based on a property you define
    - [x] get() proxies to build() (discrete instances by default)


weierophinney:
Pinging @bakura10 - this is the rabased version. :)

bakura10:
Awesome! Can I close my branch then?

weierophinney:
Yes, and we can close #8 to move discussion here.

xtreamwayz:
I have been playing with this a few weeks now and I really like it. However with the latest immutable changes it is a bit too limited for my needs or I am missing something? In version 2.x I can add configuration like this and pass it around easily:

    $serviceManager->setService('Config', $config);

I don't see a way to do this with 3.x without extending the class, since access to the services property is completely blocked. A simple solution could be to add a method like $serviceManager->withServices($array).

Second issue:

If I call get(Service/Name/X) with $options, at the end it is stored in the services array. So the next time when I call get(Service/Name/X) without options, it returns the stored service with options from the first call. I think there should be a check so an object is not stored in the services array if there are options set. I can imagine that a specific object is always called with options, but since it is always regenerated it's still pointless to store it.

weierophinney:
>If I call get(Service/Name/X) with $options, at the end it is stored in the services array. 

I think you're working with an outdated version. In the rebased version as shown in #9, get() no longer exposes the $options argument; it is only present in the new build() method. Further, only get() caches the created service.

weierophinney:
>A simple solution could be to add a method like $serviceManager->withServices($array).

This can be done with the withConfig() method now, can it not? If not, do a PR against my branch so we can enable it.

xtreamwayz:
Disregard my comments for now. I've found bakura10's version that does have the changes. For some reasons when I use the git commands in the first comment I don't get the changed ServiceManager.php. I blame my git skills. I'll wait with more testing until a 3.x-beta version or my git skills improve :)

bakura10:
Hi!

Actually yes, it seems your branch does not contain the lastest changes (does not contain build method).

bakura10:
Note for myself: when calling `configure`, we must only accept abstract factories that implement AbstractFactoryInterface, as just being callable is not enough for an abstract factory.

bakura10:
I've started writing docs on mwop's branch, for those who'd be keen to read them: https://github.com/weierophinney/zend-servicemanager/pull/1

weierophinney:
@bakura10 - I've added you as a collaborator on my own repo - go ahead and
commit directly to the branch, and the changes will show up on this PR.

bakura10:
Can you please handle the issue with hot branch though? I'm not sure what has happened :)

src/Factory/AbstractFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager\Factory;
+
+/**
+ * Interface for an abstract factory
+ *
+ * An abstract factory extends the factory interface, but also have an additional "canCreateService" method,
```
samsonasik: s/have/has

sandrokeil:
@bakura10 I really like the new docs and best practices. Awesome work. :+1:

sandrokeil:
@bakura10 @weierophinney I think the `ServiceLocatorInterface` should extends `ContainerInterface`, so I can create interopable factory classes. Also the `ServiceLocatorInterface` typehints should be replaced with `ContainerInterface`. Otherwise the component (zf2 module) factories are useless and not interopable.

bakura10:
That's the case. Mwop branch is not up to date with my latest work when he did the rebase :)

Envoyé de mon iPhone

bakura10:
Any news @weierophinney ? For now I'm stuck until you can update your branch :(.

weierophinney:
@bakura10 what commits do I need to cherry-pick in? I'll be in #zftalk.dev starting in a few minutes if you want to discuss.

weierophinney:
@bakura10 Updated now! Only 2 failing tests, too; I'll take a look at those shortly.

weierophinney:
@bakura10 tests and CS checks are now passing. :)

samsonasik:
@weierophinney I created PR https://github.com/weierophinney/zend-servicemanager/pull/2 to your branch for the typo fix ;)

weierophinney:
Test coverage is not horrible at this point. Overall is at 61%, with Exception, Initializer, plugin managers, and the ServiceLocatorInterface being at 100%. Factory is at 36% and ServiceManager is at 70%. I'll try and improve these today.

bakura10:
Thanks a lot Matthew ;)

Envoyé de mon iPhone

weierophinney:
Testing is now at 100% coverage, @bakura10 . Ball is in your court; are we done? Should I merge to develop? or are you still writing documentation?

weierophinney:
Don't know why coveralls isn't finishing, but here's the report:

- https://coveralls.io/builds/3569548

weierophinney:
@bakura10 Looks like delegators and lazy services are still to be documented; are you writing those?

Next: I made one change to configuration: I re-instated the "services" key, which allows passing key/(object|array) pairs via configuration. This allows setting things such as configuration, which we'll want to be able to do.

I have two related questions for this:

- Was services intentionally removed previously? If so, what was the rationale?
- Do you think there's another way to handle things like configuration? If so, how?

Thanks — this is looking excellent! I was able to read the code and understand the internals easily, and the functionality is definitely robust. Kudos!

bakura10:
Hi,

Awesome work! Thanks.

For the services key, I never used it, so I've just forgotten to bring it back. Regarding configuration, what do you mean? I love the way we enforce it through constructor, and can alter it using the withConfig.

My only concern with the "withConfig" thing, is that it will introduce the need for people using it to make sure that they change the main service locator (how?) so that the app service locator is the new they have created. Having said that, I've never had to alter the config at runtime, so...

Finally, for the doc, I'm so uncomfortable with delegators and lazy factories that I would actually copy-paste the existing documentation, so I preferred someone using them, with real use cases in mind, to write this part of the doc :).

Good for merging to me!!

bakura10:
By the way, here is our final results (with, I think, an even better coverage than before):

+2,195 lines added −4,817 lines removed.

Not bad!!

src/ServiceManager.php
```diff
@@ -48,1224 +36,337 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-            $stack[$cName] = $cName;
-            $cName = $this->aliases[$cName];
-        }
-
-        return $cName;
-    }
-
-    /**
-     * Retrieve a registered instance
+     *      - factories: a list of key value that map a service name with a factory
```
bakura10: services key is missing from phpDoc

src/ServiceManager.php
```diff
@@ -48,1224 +36,337 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
+            ? $config['delegators']
+            : []);
+        $this->shared          = (isset($config['shared']) ? $config['shared'] : []) + $this->shared;
+        $this->aliases         = (isset($config['aliases']) ? $config['aliases'] : []) + $this->aliases;
+        $this->sharedByDefault = isset($config['shared_by_default'])
+            ? $config['shared_by_default']
+            : $this->sharedByDefault;
+
+        // For abstract factories and initializers, we always directly
+        // instantiate them to avoid checks during construction
```
bakura10: Actually, it may be better to be "[...] to avoid checks during object service construction"?

Maks3w:
Please limit coverage to class under test. Add `@covers` annotations into class docblocks

weierophinney:
@Maks3w — Why? All cases where I added `@covers` were done to indicate the
specific methods being covered, most of which were non-public and needed
the annotation in order for coverage to be picked up.

Maks3w:
Because limit the scope of the class reveal tests are not doing all the expected unit testing. 

If after add the annotation you keep the same grade of coverage then congratulations. If not there're a leak in the unit test

weierophinney:
In each case, I can verify lines were covered using a step debugger; until
the annotations were added, however, the report did not reflect those
lines.

Maks3w:
Note I'm not saying `@cover` annotations in other part of the test suite are wrong. I just say at least classblock level annotations should be added and should be endorsed to our CS guidelines.

We can appreciate how ZF2.4 had a coverage >80% and after the split lots of components even the most important ones have a coverage lower than 70% or even lower than 50%. For this reason I suggest to add `@cover` annotations at class level at least.

weierophinney:
> For the services key, I never used it, so I've just forgotten to bring it back.

Okay, cool. Consider it re-added. :)

> Regarding configuration, what do you mean? I love the way we enforce it through constructor, and can alter it using the withConfig.

I'm talking a "config" service, such as we often have in ZF2, containing the full application configuration for factories to consume when configuring service instances. Totally separate from `withConfig()`, which is for configuring the service manager and how it resolves services.

weierophinney:
> Note I'm not saying @cover annotations in other part of the test suite are wrong. I just say at least classblock level annotations should be added and should be endorsed to our CS guidelines.

@Maks3w That was entirely unclear from your original message; the original message had the implication that the *current* `@covers` annotations were incorrect. I'll add the class-level annotations now.

weierophinney:
> Finally, for the doc, I'm so uncomfortable with delegators and lazy factories that I would actually copy-paste the existing documentation, so I preferred someone using them, with real use cases in mind, to write this part of the doc :).

I'll get that done today.

bakura10:
> I'm talking a "config" service, such as we often have in ZF2, containing the full application configuration for factories to consume when configuring service instances. Totally separate from withConfig(), which is for configuring the service manager and how it resolves services.

Can't we simply add it as a service in the constructor?

```php
public function __construct(array $config)
{
  $this->configure($config);
  $this->services['config'] = $config;
}
```

This is a bit opiniated but this is definitely the simplest way to solve this.

weierophinney:
> Can't we simply add it as a service in the constructor?

It's a different kind of configuration entirely! The `$config` that the constructor takes needs to be in a specific form, and contain specific keys; it'd be more akin to the configuration found under `vendor.package` in interop-config (as we're discussing in zend-expressive). The `config` service I want is the one that has all the various `vendor.package` details, not *just* the service manager configuration.

xtreamwayz:
@bakura10 Not sure if this makes it easier, but this is what needs to be achieved:

```php
$config = [
    'service_manager' => [
        'services' => [
            'config' => 'array, config class or service'
        ],
        'factories' => []
    ],
];
```

This can be done in zf2 which is really useful by adding it like this: 

```php
$container->setService('config', 'array, config class or service');
```

bakura10:
I'm not sure to understand @weierophinney. Basically you are saying that the configuration that Zend Expressive receives is a different configuration? Sorry, I'm a bit lost.

weierophinney:
I've completed writing documentation for the lazy services and delegators, and edited all other chapters.

@bakura10 — I think we may have an issue with lazy services at this time. In the existing v2 code, you do the following:

```php
```php
use MyApp\Buzzer;
use Zend\ServiceManager\Factory\InvokableFactory;
use Zend\ServiceManager\Factory\LazyServiceFactoryFactory;
use Zend\ServiceManager\Proxy\LazyServiceFactory;

$config = [
    'lazy_services' => [
         // mapping services to their class names is required
         // since the ServiceManager is not a declarative DIC
         'class_map' => [
             Buzzer::class => Buzzer::class,
         ],
    ],
];

$serviceManager = new \Zend\ServiceManager\ServiceManager([
    'services' => [
        'config' => $config,
    ],
    'factories' => [
        Buzzer::class             => InvokableFactory::class,
        LazyServiceFactory::class => LazyServiceFactoryFactory::class,
    ],
    'delegators' => [
        Buzzer::class => [
            LazyServiceFactory::class,
        ],
    ],
]);
```

This is problematic because currently we do not pull delegator factories as services; they *must* be callables or instantiable classes that are callable. I'll see if I can come up with something that works around this, but wanted to give you a heads-up.

weierophinney:
> I'm not sure to understand @weierophinney. Basically you are saying that the configuration that Zend Expressive receives is a different configuration? Sorry, I'm a bit lost.

Don't worry about it; it's solved already by allowing passing the `services` key.

bakura10:
Once you merge to develop, I'll be happy to backport to Zend-Expressive and performs some benchmarks :).

akrabat:
Is this completely BC with Zend\ServiceManager 2.x? If not, can we have a section in the docs about what to do to upgrade?

weierophinney:
@akrabat It's not completely BC (doesn't have any of the methods for attaching services; factories have different signatures). I'll add a migration doc to my TODO for this.

akrabat:
Right.

bakura10:
This is the first time I'm using Blackfire so I'm not sure I'm doing it right, but on a very simple Expressive app, using a container driven with just one route, here is the result I have, in comparaison of using ZF2 SM:

image

weierophinney:
@bakura10 All I'm seeing is an image icon for a PNG...

bakura10:
REfresh the page, I messed up.

weierophinney:
Interpretation: 5x speed improvement, large memory gains. :)

bakura10:
Yep! I'd love if you could do the test on a larger app that use SM a bit more extensively (your blog?) to see how it scales on larger app.

bakura10:
(And my benchmark is on PHP 5.5... it will really fly with PHP 7)

bakura10:
Just a thing I've realized by reading Expressive doc. This refactor makes this impossible due to the fact that invokables are gone:

```php
$container->setInvokableClass(
    'Zend\Expressive\Router\RouterInterface',
    'Zend\Expressive\Router\FastRouteRouter'
);
```

With this refactor, it would need two paths: set an alias from RouterInterface to FastRouteRouter, and set invokable of FastRouteRouter.

weierophinney:
> With this refactor, it would need two paths: set an alias from RouterInterface to FastRouteRouter, and set invokable of FastRouteRouter.

Yep; not terribly concerned, as we can handle that detail from the installer. :)

bakura10:
I'm just amazed how easy it is to switch implementation and still having everything work just by changing a single line.

weierophinney:
> I'm just amazed how easy it is to switch implementation and still having everything work just by changing a single line.

Now you see why I was insisting on keeping as much BC as possible. :) It's a good thing sometimes!

bakura10:
No, I was more referring to the fact of being able to change router. I'm happy to break BC for those components (I love how much streamline the SM component has become).

I'm playing with FastRoute and AuraRouter, I feel that FastRoute should be the default. I love how easy it is to introduce constraints without having to deal with the options.

weierophinney:
> I'm playing with FastRoute and AuraRouter, I feel that FastRoute should be the default. I love how easy it is to introduce constraints without having to deal with the options.

Bring that up on the expressive issue tracker. I think we actually already have an issue setup to do that; we originally didn't as we didn't have URI generation capabilities, but once I added those, there's no longer a blocker.

weierophinney:
I've revamped the lazy services at this time. Usage is now:

```php
$config = [
    'lazy_services' => [
        'class_map' => [
            InvokableObject::class => InvokableObject::class,
        ],
        'proxies_namespace'  => 'TestAssetProxy',
        'proxies_target_dir' => $this->proxyDir,
        'write_proxy_files'  => true,
    ],
    'factories' => [
        InvokableObject::class => InvokableFactory::class,
    ],
    'delegators' => [
        InvokableObject::class => [LazyServiceFactory::class],
    ],
];

$container = new ServiceManager($config);

// Discrete instance:
$instance  = $container->build(InvokableObject::class, ['foo' => 'bar']);

// Shared instance:
$instance = $container->get(InvokableObject::class)
```

Essentially, I've moved the lazy service configuration directly into the ServiceManager, and moved creation of the `LazyServiceFactory` into the ServiceManager as well. This simplifies usage, and gets rid of a conventions-based approach (i.e., that you have a `config` service with a top-level `lazy_services` key).

I'm going to update the PR description to indicate remaining tasks now, and hopefully we can merge this to develop in the next day or two.

bakura10:
Looks awesome. Thanks!

Envoyé de mon iPhone

weierophinney:
@bakura10 I'm currently creating the migration guide, and I see one other change I feel needs to be made: `AbstractFactoryInterface::canCreateServiceWithName()` needs to accept a container instance in addition to the requested name. The reason: there may be dependencies on other services that determine if service creation is possible.

We use this in several of the shipped abstract factories already. Many of them check to see if we have the `config` service, and then if required keys are present; if they are not, it can then return `false` immediately.

I'll make the change, but wanted to give you a heads-up first. I'll be on IRC shortly.

weierophinney:
Okay, PR is complete! @bakura10 and @akrabat — please review!

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,61 @@
...
 /**
- * ServiceManager implementation for managing plugins
- *
- * Automatically registers an initializer which should be used to verify that
- * a plugin instance is of a valid type. Additionally, allows plugins to accept
- * an array of options for the constructor, which can be used to configure
- * the plugin when retrieved. Finally, enables the allowOverride property by
- * default to allow registering factories, aliases, and invokables to take
- * the place of those provided by the implementing class.
+ * Abstract plugin manager
```
Maks3w: Could be the description improved?

src/Factory/DelegatorFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager\Factory;
+
+use Interop\Container\ContainerInterface;
+
+/**
+ * Delegator factory interface
```
Maks3w: Could be the description improved?

src/PluginManagerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace Zend\ServiceManager;
+
+use Zend\ServiceManager\Exception\InvalidServiceException;
+
+/**
+ * Interface for a plugin manager
+ *
+ * A plugin manager is a specialized service locator used to create homogeneous objects
```
Maks3w: I'm not sure but I think the phrase should be "..used **for** to create ..."

src/ServiceManager.php
```diff
@@ -9,36 +9,27 @@
...
+use Zend\ServiceManager\Exception\InvalidArgumentException;
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
+use Zend\ServiceManager\Factory\AbstractFactoryInterface;
+use Zend\ServiceManager\Factory\DelegatorFactoryInterface;
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
+use Zend\ServiceManager\Initializer\InitializerInterface;
 
-class ServiceManager implements ServiceLocatorInterface, ContainerInterface
+/**
+ * Service Manager
```
Maks3w: Could be the description improved?

src/PluginManagerInterface.php
weierophinney: Nope; for is an unnecessary preposition in this case.

Maks3w: thx

test/AbstractPluginManagerTest.php
```diff
@@ -9,278 +9,109 @@
...
+use Zend\ServiceManager\Factory\FactoryInterface;
+use Zend\ServiceManager\Factory\InvokableFactory;
+use ZendTest\ServiceManager\TestAsset\InvokableObject;
+use ZendTest\ServiceManager\TestAsset\SimplePluginManager;
 
-class AbstractPluginManagerTest extends \PHPUnit_Framework_TestCase
+/**
+ * @covers \Zend\ServiceManager\AbstractPluginManager
+ */
+class AbstractPluginManagerTest extends TestCase
```
Maks3w: This should extend from ServiceManagerTest for to assert complaints with liskov substitution principle

Maks3w:
I disagree with the use of `__invoke` magic method on interfaces. To have more than one interface with a definition for the same method name it's bad (IMO)

Maks3w:
I would like to see more compliance tests for reuse outside of the repo (e.x. assert component Foo plugin manager does not break LSP)

Maks3w:
Could be the `array` type anotations to be replaced with typed arrays? (Foo[])

src/ServiceLocatorInterface.php
```diff
@@ -9,25 +9,24 @@
...
-    /**
-     * Check for a registered instance
+     * Build a service by its name, using optional options (such services are NEVER cached).
      *
-     * @param  string|array  $name
-     * @return bool
+     * @param  string $name
+     * @param  null|array  $options
+     * @return mixed
+     * @throws Exception\ServiceNotFoundException If no factory/abstract
```
Maks3w: May exception type should be `NotFoundException`

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-     *
-     * @param DelegatorFactoryInterface|callable $delegatorFactory the delegator factory
-     * @param string                             $rName            requested service name
-     * @param string                             $cName            canonical service name
-     * @param callable                           $creationCallback callback for instantiating the real service
+     * Get a factory for the given service name
      *
+     * @param  string $name
      * @return callable
+     * @throws ServiceNotFoundException
```
Maks3w: May exception type should be `NotFoundException` for to be more close with container interop expectations.

src/Proxy/LazyServiceFactory.php
```diff
@@ -21,17 +21,17 @@
  *
  * @link https://github.com/Ocramius/ProxyManager/blob/master/docs/lazy-loading-value-holder.md
  */
-class LazyServiceFactory implements DelegatorFactoryInterface
+final class LazyServiceFactory implements DelegatorFactoryInterface
```
Maks3w: I've doubts if the name of this class should include the external library, something like `ProxyManagerLazyServiceFactory`

Maks3w: Also should this class to have the suffix `Proxy`?

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
+            // Code is inlined for performance reason, instead of abstracting the creation
+            $factory = $this->getFactory($name);
+            return $factory($this->creationContext, $name, $options);
+        };
 
-                if ($peeringServiceManager->has($name)) {
-                    $peeringServiceManager->serviceManagerCaller = null;
-                    return true;
-                }
+        for ($i = 0; $i < $delegatorsCount; $i += 1) {
```
Maks3w: Why the for loop instead foreach?

test/AbstractPluginManagerTest.php
weierophinney: It can't do it directly, as the constructor arguments differ (`AbstractPluginManager` requires a parent locator instance in addition to the optional configuration). I'll see if I can move the relevant tests into a trait, however.

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-        $result = false;
-        $this->nestedContext[$cName] = false;
-        foreach ($this->abstractFactories as $abstractFactory) {
-            $pendingKey = get_class($abstractFactory).$cName;
-            if (isset($this->pendingAbstractFactoryRequests[$pendingKey])) {
-                $result = false;
-                break;
+            if (is_string($delegatorFactory) && ! class_exists($delegatorFactory)) {
+                throw new ServiceNotCreatedException(sprintf(
+                    'An invalid delegator was provided. A callable or an instance of "%s" as expected, '
```
Maks3w: s/as expected/was expected/

weierophinney:
> I disagree with the use of __invoke magic method on interfaces. To have more than one interface with a definition for the same method name it's bad (IMO)

We'll have to agree to disagree on that. One valuable aspect to using `__invoke()` is that we can assume any `callable` is valid in the given context, which is part of the huge performance and memory boost this implementation provides.

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-            if (isset($stack[$alias])) {
-                throw new Exception\CircularReferenceException(sprintf(
-                    'The alias definition "%s" : "%s" results in a circular reference: "%s" -> "%s"',
-                    $alias,
-                    $nameOrAlias,
-                    implode('" -> "', $stack),
-                    $alias
+            if (! is_callable($delegatorFactory)) {
+                throw new ServiceNotCreatedException(sprintf(
+                    'An invalid delegator was provided. A callable or an instance of "%s" as expected, '
```
Maks3w: s/as expected/was expected/

src/ServiceManager.php
weierophinney: I'll have to ask @bakura10. It shouldn't make a difference either way.

src/ServiceManager.php
Maks3w: This exception message is the same as above. To have different exceptions messages will help to understand the problem better.

Such like, "delegator is not callable, only callables and instance %s are allowed"

src/Proxy/LazyServiceFactory.php
weierophinney: > I've doubts if the name of this class should include the external library, something like ProxyManagerLazyServiceFactory

No; it's providing a specific feature *for this component*: lazy services. It does so via `ProxyManager`, but that's an implementation detail.

> Also should this class name to have the suffix Proxy?

No; it's a factory, not a proxy class. It's specifically a delegator factory, so the only potential change I could agree with is to add the verbiage "DelegatorFactory" to it. That said: this class comes from v2, and I don't recall a naming objection in that version.

weierophinney:
@Maks3w —

I don't understand the following comment:

> I would like to see more compliance tests for reuse outside of the repo (e.x. assert component Foo plugin manager does not break LSP)

Can you elaborate, please?

weierophinney:
> Could be the array type anotations to be replaced with typed arrays? (Foo[])

Where, @Maks3w ?

src/ServiceManager.php
Maks3w: I suggest replace "An invalid delegator was provided" with something like "An invalid delegator was provided because the class %s not exists"

src/ServiceManager.php
weierophinney: Fixed now.

Maks3w:
What I expect about compliant tests are two things.

1) My implementation of interface X at least follow the same contract as the interface.

2) Compliance test stress my implementations about scenarios I didn't think yet (wrong inputs, etc). Usually a developer creates an implementation for resolve a particular need and things like float instead integer or case sensitive is sometimes forgotten.

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,72 @@
...
-     *
-     * @param  string                               $canonicalName
-     * @param  string                               $requestedName
-     * @return null|\stdClass
-     * @throws Exception\ServiceNotCreatedException If resolved class does not exist
+     * @param ContainerInterface $parentLocator
+     * @param array              $config
      */
-    protected function createFromInvokable($canonicalName, $requestedName)
+    public function __construct(ContainerInterface $parentLocator, array $config)
```
Maks3w: Parent $config its optional, defaults to `[]` Why is here mandatory?

src/AbstractPluginManager.php
```diff
@@ -9,308 +9,72 @@
...
-     *
-     * Overrides parent implementation by passing $creationOptions to the
-     * constructor, if non-null.
-     *
-     * @param  string                               $canonicalName
-     * @param  string                               $requestedName
-     * @return null|\stdClass
-     * @throws Exception\ServiceNotCreatedException If resolved class does not exist
+     * @param ContainerInterface $parentLocator
+     * @param array              $config
```
Maks3w: Please add `@see parent constructor $config details` or similar

Maks3w: From ServiceManager 

> * See {@see \Zend\ServiceManager\ServiceManager::configure()} for details
> * on what $config accepts.

weierophinney:
@Maks3w —

> What I expect about compliant tests are two things.
> 
> 1) My implementation of interface X at least follow the same contract as the interface.

We're already doing this; we have several test assets that implement the various interfaces in order to demonstrate that the service manager consumes them. If this is not sufficient, I ask that you please provide examples that demonstrate what you're asking.

> 2) Compliance test stress my implementations about scenarios I didn't think yet (wrong inputs, etc). Usually a developer creates an implementation for resolve a particular need and things like float instead integer or case sensitive is sometimes forgotten.

Within the tests, we have a number of data providers that we handle input to methods correctly, raising exceptions where necessary and otherwise verifying behavior. Does that not meet your criteria?

If you're stressing the idea of testing something like how an updated ValidatorPluginManager might work with this, I believe that's beyond the scope of this component; those tests would need to live in the component that implements.

Alternately: are you suggesting having a suite of tests that an implementation could consume and test against, similar to the PSR-7 suite you were developing? If so, we can do that later; it doesn't need to be part of this pull request. (As it is, I'm not 100% convinced it's required.)

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-     * Set flag indicating whether services are shared by default
+     * 'shared' => [
+     *     MyService::class => true, // will be shared, even if "sharedByDefault" is false
+     *     MyOtherService::class => false // won't be shared, even if "sharedByDefault" is true
+     * ]
      *
-     * @param  bool $shareByDefault
-     * @return ServiceManager
-     * @throws Exception\RuntimeException if allowOverride is false
+     * @var array
```
Maks3w: Seems this could be `boolean[]`

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-
-    /**
      * @var array
      */
     protected $aliases = [];
 
     /**
+     * A list of already loaded services (this act as a local cache)
+     *
      * @var array
```
Maks3w: object[]

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
+     * A list of aliases
+     *
+     * Should map one alias to a service name, or another alias (aliases are recursively resolved)
      *
-     * @var array
-     */
-    protected $instances = [];
-
-    /**
      * @var array
```
Maks3w: string[]

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
 
     /**
      * @var array
      */
-    protected $pendingAbstractFactoryRequests = [];
+    protected $lazyServices = [];
 
     /**
-     * @var integer
+     * @var Proxy\LazyServiceFactory
```
Maks3w: this can be null too https://github.com/weierophinney/zend-servicemanager/blob/refactor-sm/src/ServiceManager.php#L254

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
      */
-    protected $nestedContextCounter = -1;
+    protected $lazyServicesDelegator;
```
Maks3w: this property can be change to private. It's only used by constructor and createLazyServiceDelegatorFactory

src/ServiceManager.php
weierophinney: Not true; they can also be arrays (e.g., `config` service). `array` is correct here.

src/ServiceManager.php
```diff
@@ -48,1224 +51,427 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
...
-     * @var array[]
+     * @var string[]|DelegatorFactoryInterface[]
      */
     protected $delegators = [];
 
     /**
      * @var array
      */
-    protected $pendingAbstractFactoryRequests = [];
+    protected $lazyServices = [];
```
Maks3w: I expected to contain a collection of services but seems contains the configuration of ProxyConfiguration. 

Could be this rename to `lazyServicesConfiguration`?

src/ServiceManager.php
samsonasik: maybe `bool[]` for consistencies ?

src/AbstractPluginManager.php
weierophinney: Fixed.

Maks3w:
Yes, I'm suggesting a suite of test for to consume outside of this component (like the PSR-7)

Let's say I don't like to test the `validate()` method in the tests. I prefer to test only public API (see https://github.com/zendframework/zend-inputfilter/blob/master/test/InputFilterPluginManagerTest.php)

Maks3w:
Missing feedback for this notes.

https://github.com/zendframework/zend-servicemanager/pull/9#discussion-diff-39661614
https://github.com/zendframework/zend-servicemanager/pull/9#discussion-diff-39661727
https://github.com/zendframework/zend-servicemanager/pull/9#discussion-diff-39662009

src/ServiceManager.php
```diff
@@ -9,1263 +9,469 @@
...
+     * @param  null|array $options
+     * @return object
      */
-    public function canCreate($name, $checkAbstractFactories = true)
+    private function createDelegatorFromName($name, array $options = null)
     {
-        trigger_error(sprintf('%s is deprecated; please use %s::has', __METHOD__, __CLASS__), E_USER_DEPRECATED);
-        return $this->has($name, $checkAbstractFactories, false);
-    }
+        $delegatorsCount  = count($this->delegators[$name]);
```
Maks3w: unused variables (previously used in for loop)

Maks3w:
https://github.com/zendframework/zend-servicemanager/pull/9#discussion-diff-39658784 too

src/ServiceManager.php
```diff
@@ -9,1263 +9,469 @@
...
+                }
 
-        while ($this->hasAlias($cName)) {
-            if (isset($stack[$cName])) {
-                throw new Exception\CircularReferenceException(sprintf(
-                    'Circular alias reference: %s -> %s',
-                    implode(' -> ', $stack),
-                    $cName
+                throw new InvalidArgumentException(sprintf(
+                    'An invalid abstract factory was registered. A callable or an instance of "%s" was expected, '
```
Maks3w: Exception message claims factory could be an instance of X but there is no check testing that. Just must be callable.

Note: If `new` fail the exception won't be raised

src/ServiceManager.php
```diff
@@ -9,1263 +9,469 @@
...
 
-        // Still no instance? raise an exception
-        if ($instance === null) {
-            $this->checkNestedContextStop(true);
-            if ($isAlias) {
-                throw new Exception\ServiceNotFoundException(sprintf(
-                    'An alias "%s" was requested but no service could be found.',
-                    $name
+                throw new InvalidArgumentException(sprintf(
+                    'An invalid initializer was registered. A callable or an instance of "%s" was expected, but
```
Maks3w: Exception message claims initializer could be an instance of X but there is no check testing that. Just must be callable.

Note: If new fail the exception won't be raised

src/ServiceManager.php
```diff
@@ -9,1263 +9,469 @@
...
-    protected function doCreate($rName, $cName)
+    private function getFactory($name)
     {
-        $instance = null;
+        $factory = isset($this->factories[$name]) ? $this->factories[$name] : null;
 
-        if (isset($this->factories[$cName])) {
-            $instance = $this->createFromFactory($cName, $rName);
+        if (is_string($factory)) {
+            $this->factories[$name] = $factory = new $factory();
```
Maks3w: Oops! this change the object state (>factories) before to check if the value it's valid.

Exception may raised (end of loop) but the object state keep the same.

Maks3w:
I see a point of performance improvement with a subset of AbstractFactories.

If the AbstractFactory can provide a full set of names all of that names can be registered as normal factories instead call the canCreateServiceWithName loop

weierophinney:
> I see a point of performance improvement with a subset of AbstractFactories.
> 
> If the AbstractFactory can provide a full set of names all of that names can be registered as normal factories instead call the canCreateServiceWithName loop

Which is one of the benefits of `AbstractFactoryInterface` extending `FactoryInterface`; you can do *exactly* that with this proposed implementation (assign specific services to the `AbstractFactoryInterface` implementation in the `factories` configuration).

src/ServiceManager.php
weierophinney: Fixed here: https://github.com/weierophinney/zend-servicemanager/commit/13deb3bcbff26e2670958eb36e3bec8332eff52f#diff-14ccad9c10b086d30c53ca4b694cae5dR83

src/ServiceManager.php
weierophinney: Fixed locally; will push with other changes shortly.

src/ServiceManager.php
weierophinney: I'd prefer not to; the configuration key is `lazy_services`, which means the normalized key I'd look for in the class is `lazyServices`. Changing the name breaks internal consistency with the other property names (which all map to configuration keys).

Maks3w: I suggest then to add the config key literal to each mapped property. Property names are a transformation of the config key (don't match when you perform a string search across the repo)

src/ServiceManager.php
```diff
@@ -9,1263 +9,469 @@
...
+     * configuration present.
      *
-     * @param  string $canonicalName
-     * @param  string $requestedName
-     * @return mixed
-     * @throws Exception\ServiceNotCreatedException If factory is not callable
+     * @return Proxy\LazyServiceFactory
      */
-    protected function createFromFactory($canonicalName, $requestedName)
+    private function createLazyServiceDelegatorFactory()
```
Maks3w: I feel all this method could be moved away from here. May the factory its a better place for it.

src/ServiceManager.php
weierophinney: > Exception message claims initializer could be an instance of X but there is no check testing that. Just must be callable.

Referencing the interface in the exception message is reasonable here, and a good idea as they can then check the signature as well.

weierophinney: > Note: If new fail the exception won't be raised

Right; PHP will raise an error! :smile:

That said, I'm adding a `class_exists()` check to be more safe.

src/ServiceManager.php
```diff
@@ -9,1263 +9,468 @@
...
-                return $instance;
+        try {
+            if (!isset($this->delegators[$resolvedName])) {
+                // Let's create the service by fetching the factory
+                $factory = $this->getFactory($resolvedName);
+                $object  = $factory($this->creationContext, $resolvedName, $options);
+            } else {
+                $object = $this->createDelegatorFromName($resolvedName, $options);
             }
+        } catch (Exception $exception) {
```
Maks3w: I think we can go without wrap the exception in another one. I guess will improve the performance a bit

src/ServiceManager.php
```diff
@@ -9,1263 +9,468 @@
...
-        return $this;
-    }
-
-    /**
-     * Attempt to retrieve an instance via a peering manager
-     *
-     * @param  string $name
+     * @param  string     $resolvedName
+     * @param  null|array $options
      * @return mixed
```
Maks3w: Missing `@throws` declaration

Maks3w:
Another performance suggestion. Resolve alias when create instead perform a loop each time a service it's get/build.

src/ServiceManager.php
weierophinney: I chose the current solution for the following reasons:

- We should only create the `LazyServiceFactory` once; this method enforces that, returning early if an instance already exists. (Yes, we could also do it in a ternary.)
- The logic for creating the `LazyServiceFactory` needs to use configuration passed to the service manager. In v2, we use the `config` service to store lazy service configuration, and require a specific key under that; this practice binds the user to a *convention* rather than a *contract*; by having the lazy service configuration understood directly by the service manager, we create a contract.
- Related, if the configuration is internal to the `ServiceManager`, the only viable mechanism for creating the `LazyServiceFactory` outside the class would be a static factory (as it would need to be stateless). This imposes additional overhead (class autoloading, memory for storing the class), contradicting a primary goal of the rewrite (performance).

For those reasons, I inlined the `LazyServiceFactory` creation as a method in the `ServiceManager` itself

src/ServiceManager.php
weierophinney: Except that one TODO I have is to make our exceptions implement container-interop exception interfaces, to ensure they enforce the same interface. The try/catch needs to be here as a result.

It also ensures consistency when handling exceptions raised by the container.

Maks3w: Then the correct place should be the method which must comply the contract (get/build) any exception should be catched no matter if it's only on create

weierophinney:
> Another performance suggestion. Resolve alias when create instead perform a loop each time a service it's get/build.

That *creates* a performance hit; if the aliased service is never retrieved during execution, it never needs to be resolved, meaning no performance hit. Doing it ahead of time means resolving all aliases on each request.

bakura10:
Interesting idea @Maks3w . I think the overhead of doing it will be negligible toward the fact we save one for loop and one function call.Do you want to try it @weierophinney ? Maybe you should try it in your codebase first.

bakura10:
Typically, you create far more services than you have aliases. If you have 5 aliases and resolve them at construction time, you will indeed do the 5 loops all time. But you will save one function call and the loop for potentially hundreds of objects created using the SM.

bakura10:
`get` would become like this:

```php
public function get($name)
    {
        $name = isset($this->resolvedAlias[$name]) ? $this->resolvedAlias[$name] : $name;
    }
```

bakura10:
Every little bit can be important in this performance-sensitive component, it's an easy fix actually. Except this I'm not sure there are many more way to improve performance ;).

Maks3w:
Basically the optimizations I see are about to fill the factories property (and related ones) as soon as the value is known.

Maks3w:
The other optimization I suggest before is add a method to `AbstractFactoryInterface` for to retrieve all the possible names inmeditly if the factory allows that. That save from call `canCreateServiceWithName` in the loop always and just perform a search in the factories table.

Maks3w:
As I said, everything it's hide behind some kind of abstraction then retrieve it if possible and avoid query the same abstraction again.

weierophinney:
> The other optimization I suggest before is add a method to AbstractFactoryInterface for to retrieve all the possible names inmeditly if the factory allows that. That save from call canCreateServiceWithName in the loop always and just perform a search in the factories table.

One of the big refactors that @bakura10 did was two-fold:

- Ability to re-use the same factory for multiple service names on an N:1 basis.
- Define `AbstractFactoryInterface` to *extend* the `FactoryInterface` specifically for the instance creation aspect.

This gives the ability to optimize *as needed*, but still retains the ability to do *dynamic resolution* (which is the chief facility an abstract factory provides). I don't think we need to or should change the abstract factory interface at this time to support returning a list of services; there's already a mechanism for it: assign the abstract factory as a *regular factory* for known service names:

```php
[
    'abstract_factories' => [
        'SomeAbstractFactoryThatWillResolveAnything',
    ],
    'factories' => [
        'SpecificServiceName' => 'SomeAbstractFactoryThatWillResolveAnything',
    ],
]
```

weierophinney:
Regarding the comments on alias resolution: I'm giving it a try now. If I see the tests taking longer, I'll revert. ;-)

Maks3w:
@weierophinney I see your point but I don't have clear how that will be done of ZF2 components like Zend\Validator.

Actually we rely on PluginManagers as a bag for the collection. Are you proposing ship an static array on Zend\Validator?

src/ServiceManager.php
```diff
@@ -9,1263 +9,476 @@
...
+                }
 
-        while ($this->hasAlias($cName)) {
-            if (isset($stack[$cName])) {
-                throw new Exception\CircularReferenceException(sprintf(
-                    'Circular alias reference: %s -> %s',
-                    implode(' -> ', $stack),
-                    $cName
+                throw new InvalidArgumentException(sprintf(
+                    'An invalid abstract factory was registered. An instance of "%s" was expected, '
```
Maks3w: With this exception message seems strings are not allowed but they are allowed if resolve to a class.

Maks3w: Also the value of the string (when the string its invalid) its a valuable part of the message but `string` is display

src/ServiceManager.php
```diff
@@ -9,1263 +9,476 @@
...
+        if (isset($config['lazy_services']) && ! empty($config['lazy_services'])) {
+            $this->lazyServicesDelegator = null;
+        }
+
+        // For abstract factories and initializers, we always directly
+        // instantiate them to avoid checks during service construction.
+        if (isset($config['abstract_factories'])) {
+            foreach ($config['abstract_factories'] as $abstractFactory) {
+                if (is_string($abstractFactory) && class_exists($abstractFactory)) {
+                    $abstractFactory = new $abstractFactory();
```
Maks3w: For to solve note below I suggest to throw exception as `else` of `class_exists`

src/ServiceManager.php
```diff
@@ -9,1263 +9,476 @@
...
-        $result = false;
-        $this->nestedContext[$cName] = false;
-        foreach ($this->abstractFactories as $abstractFactory) {
-            $pendingKey = get_class($abstractFactory).$cName;
-            if (isset($this->pendingAbstractFactoryRequests[$pendingKey])) {
-                $result = false;
-                break;
+            if (is_string($delegatorFactory) && ! class_exists($delegatorFactory)) {
+                throw new ServiceNotCreatedException(sprintf(
+                    'An invalid delegator was provided. A callable or an instance of "%s" was expected, '
```
Maks3w: The exception message could detail the problem is the class not exists

Maks3w:
Coverage drop /cc @weierophinney

Maks3w:
As performance tip we could add a note on the documentation about to put the more common factories on top of the array.

src/ServiceManager.php
weierophinney: We have documented `snake_case` -> camelCase for configuration from v2.0.0;
we use the former in configuration for readability and to enforce
consistency (all lowercase prevents accidental typo problems), and the
latter in code, to conform with our choosing standards.

There's no compelling reason to change, and many compelling ones to retain
the status quo. I'm keeping this as-is.

src/ServiceManager.php
weierophinney: This method is used by both build and get, and is this the correct location
for the try/catch block.

weierophinney:
No; see the current implementations: default, shipped plugins are mapped in
the appropriate properties (factories, etc).

weierophinney:
I'll address coverage *when I complete incorporating feedback*.

test/AbstractPluginManagerTest.php
```diff
@@ -9,278 +9,154 @@
...
+        $pluginManager = new SimplePluginManager($container, $config);
 
-        $this->assertSame($barDelegator, $pluginManager->get('foo-service'));
-        $this->assertCount(1, $barDelegator->instances);
-        $this->assertCount(1, $fooDelegator->instances);
-        $this->assertInstanceOf('stdClass', array_shift($fooDelegator->instances));
-        $this->assertSame($fooDelegator, array_shift($barDelegator->instances));
+        $first  = $pluginManager->get(InvokableObject::class);
+        $second = $pluginManager->get(InvokableObject::class);
+        $this->assertInstanceOf(InvokableObject::class, $first);
```
Maks3w: Add detailed comments to each assert. There are more than one testing different objects and it's not possible to know what assert exactly fails without browse the source code.

test/AbstractPluginManagerTest.php
```diff
@@ -9,278 +9,154 @@
...
+            'share_by_default' => $shareByDefault,
+        ];
+        $options = ['foo' => 'bar'];
+
+        $container     = $this->getMock(ContainerInterface::class);
+        $pluginManager = new SimplePluginManager($container, $config);
+
+        $first  = $pluginManager->get(InvokableObject::class, $options);
+        $second = $pluginManager->get(InvokableObject::class, $options);
+        $this->assertInstanceOf(InvokableObject::class, $first);
```
Maks3w: Add detailed comments to each assert. There are more than one testing different objects and it's not possible to know what assert exactly fails without browse the source code.

test/AbstractPluginManagerTest.php
```diff
@@ -9,278 +9,154 @@
...
+                ],
+            ],
+        ]);
 
-            $this->fail('Expected the plugin manager to throw a RuntimeException, none thrown');
-        } catch (RuntimeException $exception) {
-            $this->assertTrue($pluginManager->has(__CLASS__));
-        }
+        $instance = $pluginManager->get(stdClass::class);
+        $this->assertEquals($config['option'], $instance->option);
```
Maks3w: Add detailed comments to each assert. There are more than one testing different objects and it's not possible to know what assert exactly fails without browse the source code.

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+        $serviceManager = $this->createContainer([
+            'factories' => [
+                InvokableObject::class => InvokableFactory::class
+            ]
+        ]);
+
+        $object = $serviceManager->build(InvokableObject::class, ['foo' => 'bar']);
+
+        $this->assertInstanceOf(InvokableObject::class, $object);
+        $this->assertEquals(['foo' => 'bar'], $object->options);
```
Maks3w: Need a comment this assert

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+            ],
+            'aliases' => [
+                'foo' => InvokableObject::class,
+                'bar' => 'foo'
+            ]
+        ]);
+
+        $object = $serviceManager->get('bar');
+
+        $this->assertInstanceOf(InvokableObject::class, $object);
```
Maks3w: Need a comment these assertions

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+
+    public function testCanCheckServiceExistenceWithoutCheckingAbstractFactories()
+    {
+        $serviceManager = $this->createContainer([
+            'factories' => [
+                stdClass::class => InvokableFactory::class
+            ]
+        ]);
+
+        $this->assertTrue($serviceManager->has(stdClass::class));
```
Maks3w: Need a comment these assertions

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+        $serviceManager = $this->createContainer([
+            'factories' => [
+                stdClass::class => InvokableFactory::class
+            ],
+            'abstract_factories' => [
+                new SimpleAbstractFactory() // This one always return true
+            ]
+        ]);
+
+        $this->assertTrue($serviceManager->has(stdClass::class, true));
```
Maks3w: Need a comment these assertions

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+     */
+    public function testHasCanCheckAbstractFactoriesWhenRequested($abstractFactory, $assertion)
+    {
+        $serviceManager = $this->createContainer([
+            'abstract_factories' => [
+                $abstractFactory,
+            ],
+        ]);
+
+        $this->{$assertion}($serviceManager->has(DateTime::class, true));
```
Maks3w: It's preferable do AssertEquals, improves provider reusability

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+                    if (! $instance instanceof stdClass) {
+                        return;
+                    }
+                    $instance->bar = 'baz';
+                },
+            ],
+        ]);
+
+        $dateTime = $serviceManager->get(DateTime::class);
+        $this->assertInstanceOf(DateTime::class, $dateTime);
```
Maks3w: Need a comment these assertions

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -0,0 +1,544 @@
...
+                stdClass::class => InvokableFactory::class,
+            ],
+            'initializers' => [
+                TestAsset\SimpleInitializer::class,
+            ],
+        ]);
+
+        $instance = $serviceManager->get(stdClass::class);
+        $this->assertInstanceOf(stdClass::class, $instance);
+        $this->assertEquals('bar', $instance->foo);
```
Maks3w: Need a comment this assert

test/Factory/InvokableFactoryTest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public function testCanCreateObject()
+    {
+        $container = $this->getMock(ContainerInterface::class);
+        $factory   = new InvokableFactory();
+
+        $object = $factory($container, InvokableObject::class, ['foo' => 'bar']);
+
+        $this->assertInstanceOf(InvokableObject::class, $object);
+        $this->assertEquals(['foo' => 'bar'], $object->options);
```
Maks3w: Need a comment this assert

test/LazyServiceIntegrationTest.php
```diff
@@ -0,0 +1,132 @@
...
+        return new RegexIterator($rii, '/^.+\.php$/i', RecursiveRegexIterator::GET_MATCH);
+    }
+
+    public function assertProxyDirEmpty()
+    {
+        $count = 0;
+        foreach ($this->listProxyFiles() as $file) {
+            $this->assertFail('Expected empty proxy directory; found files');
+        }
+        $this->assertEquals(0, $count);
```
Maks3w: Need a comment because the message is thrown from a different method

test/LazyServiceIntegrationTest.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    public function assertProxyFileWritten()
+    {
+        $count = 0;
+        foreach ($this->listProxyFiles() as $file) {
+            $count += 1;
+            break;
+        }
+        $this->assertNotEquals(0, $count);
```
Maks3w: Need a comment because the message is thrown from a different method

test/LazyServiceIntegrationTest.php
```diff
@@ -0,0 +1,132 @@
...
+
+        $this->assertProxyFileWritten();
+
+        // Test we got a usable proxy
+        $this->assertInstanceOf(InvokableObject::class, $instance);
+        $this->assertContains('TestAssetProxy', get_class($instance));
+
+        // Test proxying works as expected
+        $options = $instance->getOptions();
+        $this->assertInternalType('array', $options);
```
Maks3w: Need a comment these tests

test/ServiceManagerTest.php
```diff
@@ -9,1221 +9,92 @@
...
-    public function testPeeringService()
-    {
-        $di = new Di();
-        $di->instanceManager()->setParameters('ZendTest\ServiceManager\TestAsset\Bar', ['foo' => ['a']]);
-        $this->serviceManager->addAbstractFactory(new DiAbstractServiceFactory($di));
-        $sm = $this->serviceManager->createScopedServiceManager(ServiceManager::SCOPE_PARENT);
-        $sm->setFactory('di', new DiFactory());
-        $bar = $sm->get('ZendTest\ServiceManager\TestAsset\Bar', true);
-        $this->assertInstanceOf('ZendTest\ServiceManager\TestAsset\Bar', $bar);
+        $this->assertTrue($serviceManager->has(DateTime::class));
```
Maks3w: Need a comment these tests

test/ServiceManagerTest.php
```diff
@@ -9,1221 +9,92 @@
...
 
-        // check if we can receive service from peered service managers
-        $this->assertTrue($this->serviceManager->has('peered_service'));
-        $this->assertTrue($peeredServiceManager->has('peered_service'));
-        $this->assertTrue($secondParentServiceManager->has('peered_service'));
-        $this->assertSame($expectedService, $this->serviceManager->get('peered_service'));
-        $this->assertSame($expectedService, $peeredServiceManager->get('peered_service'));
-        $this->assertSame($expectedService, $secondParentServiceManager->get('peered_service'));
+        $instance = $serviceManager->get(stdClass::class);
+        $this->assertEquals($config['option'], $instance->option);
```
Maks3w: Need a comment these tests

src/ServiceManager.php
```diff
@@ -9,1263 +9,476 @@
...
-            } while ($this->hasAlias($name));
+        try {
+            if (!isset($this->delegators[$resolvedName])) {
+                // Let's create the service by fetching the factory
+                $factory = $this->getFactory($resolvedName);
+                $object  = $factory($this->creationContext, $resolvedName, $options);
+            } else {
+                $object = $this->createDelegatorFromName($resolvedName, $options);
+            }
+        } catch (Exception $exception) {
```
Maks3w: I can't find the original thread of discussion about to move this to the parent methods.

This can be keep here if we are sure all exceptions thrown outside of this try/catch complian with container interop. If not then should be moved to the parent class or locate and wrap the exception type leaks.

weierophinney: This is catching *all* exceptions. I've got a change in my working branch
to do:

```php
try {
   // ...
} catch (ContainerException $exception) {
    throw $exception;
} catch (Exception) {
    // wrap as a ServiceNotCreatedException
}
```

This approach allows exceptions raised by `getFactory()` to be rethrown
properly, as well as for factories to throw `ContainerException` instances
for proper reporting; any other exception we then wrap. It's an appropriate
solution, and guarantees interop with other containers (particularly if
somebody uses ServiceManager as a peering or delegator container within
another solution).

On Wed, Sep 16, 2015 at 3:10 PM, Maks3w <notifications@github.com> wrote:

> In src/ServiceManager.php
> <https://github.com/zendframework/zend-servicemanager/pull/9#discussion_r39678530>
> :
>
> > -
> > -        $name = $this->canonicalizeName($name);
> > -
> > -        if ($this->hasAlias($name)) {
> > -            do {
> > -                $name = $this->aliases[$name];
> > -            } while ($this->hasAlias($name));
> > +        try {
> > +            if (!isset($this->delegators[$resolvedName])) {
> > +                // Let's create the service by fetching the factory
> > +                $factory = $this->getFactory($resolvedName);
> > +                $object  = $factory($this->creationContext, $resolvedName, $options);
> > +            } else {
> > +                $object = $this->createDelegatorFromName($resolvedName, $options);
> > +            }
> > +        } catch (Exception $exception) {
>
> I can't find the original thread of discussion about to move this to the
> parent methods.
>
> This can be keep here if we are sure all exceptions thrown outside of this
> try/catch complian with container interop. If not then should be moved to
> the parent class or locate and wrap the exception type leaks.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/zend-servicemanager/pull/9/files#r39678530>
> .
>



-- 
Matthew Weier O'Phinney
matthew@weierophinney.net
https://mwop.net/

test/AbstractPluginManagerTest.php
weierophinney: PHPUnit details the test class and method, as well as the filename and line number where the assertion failed; additionally, most test methods indicate the specific behavior being tested in natural language. In most of the areas you've commented here and below, there are minimal numbers of assertions; adding commentary will add little if any clarity over what is already present.

In cases where we have multiple assertions in the same method, I *will* do some clarifications, but otherwise I feel this is busy work for very little maintenance gain.

weierophinney:
I've incorporated feedback at this point. As noted above, I'm not adding comments to all assertions, as it's not necessary in order to understand the majority of tests; I *did* add comments and custom error messages when specific test methods had multiple assertions, and where the intent could be clarified more than it already was.

I've updated exception throwing to be consistent between each of abstract factories, initializers, and delegator factories; the logic for detecting valid instances is now roughly the same, and I've varied the messages thrown based on whether or not a string was provided (vs a callable or instance of a given type).

I've also audited for coverage and written a number of new tests to ensure that all paths are covered properly.

Finally: @bakura10 and @Maks3w : I was able to add pre-resolution for aliasing, and it had no measurable impact on the test timing, which tells me that it's likely a micro-optimization, but it'll be interesting to see if we can benchmark it further. Considering it removes a function call from each of `get()`, `build()`, and `has()`, it should have a measurable impact when invocations reach > 1000.

At this time, I plan to merge to develop.

Maks3w:
The point of add a message when more than one assert exists is PHPUnit when executed on Travis-CI does not display enough information for to know what assertion is failing.

If you are able to reproduce the issue locally you are lucky. If make coffee while perform a pseudo remote debugging session

https://travis-ci.org/zendframework/zend-servicemanager/jobs/80681560#L244

Maks3w:
Thx @weierophinney I see the exception messages looks very good.

As a side note current v2 version is very painless to debug when something fails on apigility. Just a typo in the filename or any other thing it's covered by a pile of exceptions wrapping exceptions and the error message was not helpful.

:+1:

weierophinney:
@Maks3w — You still have the information necessary to look up the code in question and understand it, and start troubleshooting. In the particular case you linked to, there's a *single* assertion, a *meta* assertion even (as it uses `setExpectedException()`) — which you can easily determine by finding the associated class file and test method, printed right there in the failure report. The information is present to help you debug.

When you have *actual* assertions (not meta assertions such as `setExpectedException()` or those created by the mocking library as expectations of mock behavior), you *do* get the specific lines leading to the failure. When only a handful or less of assertions are present, and they're ordered correctly it's relatively easy to determine root causes. I stand by what I commented earlier.

weierophinney:
> As a side note current v2 version is very painless to debug when something fails on apigility. Just a typo in the filename or any other thing it's covered by a pile of exceptions wrapping exceptions and the error message was not helpful.

I'm confused: is this sarcasm? I'm not sure if v2 is "painless" or "painful", and/or if v3 is an improvement on the situation…

Maks3w:
*painful sorry

akrabat:
👍
This is excellent work!

PR #16, opened 2015-09-17
Microoptimization

Maks3w:


src/Proxy/LazyServiceFactory.php
```diff
@@ -47,7 +47,7 @@ public function __construct(LazyLoadingValueHolderFactory $proxyFactory, array $
     /**
      * {@inheritDoc}
      *
-     * @return object|\ProxyManager\Proxy\LazyLoadingInterface|\ProxyManager\Proxy\ValueHolderInterface
+     * @return object|\ProxyManager\Proxy\VirtualProxyInterface
```
bakura10: Maybe we could typehint against that https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/Proxy/ProxyInterface.php ?

ping @Ocramius

src/ServiceManager.php
```diff
@@ -241,23 +241,34 @@ public function has($name, $checkAbstractFactories = false)
...
-        $this->shared          = (isset($config['shared']) ? $config['shared'] : []) + $this->shared;
-        $this->aliases         = (isset($config['aliases']) ? $config['aliases'] : []) + $this->aliases;
-        $this->lazyServices    = array_merge_recursive($this->lazyServices, isset($config['lazy_services'])
-            ? $config['lazy_services']
-            : []);
-        $this->sharedByDefault = isset($config['shared_by_default'])
-            ? $config['shared_by_default']
-            : $this->sharedByDefault;
+        if (isset($config['services'])) {
+            $this->services = $config['services'] + $this->services;
```
bakura10: WAAAAAAW MY BELOVED ONE LINERS :'(. Okay, I ahave to admit it is easier to read ;).

Can you please add a @TODO when we will increase to PHP7, there is a good use of coalesce opeartor here.

Maks3w: There are three reasons because I did this.

1. Always the value of the property was written even if the config key does not exists.

2. Most cases a heavy operation like array concat and array merge recursive is done even when there is no new data to set.

3. Other parts of the method already did isset operations

Maks3w: As per the optimization described on point 1 (write properties with the same value it has) there is no place for coalesce operator. Just I want to skip the "else" side of the ternary operator.

src/ServiceManager.php
```diff
@@ -241,23 +241,34 @@ public function has($name, $checkAbstractFactories = false)
...
+
+        if (isset($config['shared'])) {
+            $this->shared = $config['shared'] + $this->shared;
+        }
+
+        if (isset($config['aliases'])) {
+            $this->aliases = $config['aliases'] + $this->aliases;
+        }
+
+        if (isset($config['shared_by_default'])) {
```
bakura10: There is point for coalesce HERE :D

bakura10:
I think we've squizzed every little bit of performance. It will be very hard to make this component faster ;).

src/Proxy/LazyServiceFactory.php
```diff
@@ -47,7 +47,7 @@ public function __construct(LazyLoadingValueHolderFactory $proxyFactory, array $
     /**
      * {@inheritDoc}
      *
-     * @return object|\ProxyManager\Proxy\LazyLoadingInterface|\ProxyManager\Proxy\ValueHolderInterface
+     * @return \ProxyManager\Proxy\VirtualProxyInterface
```
Maks3w: Bakura said:

> Maybe we could typehint against that https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/Proxy/ProxyInterface.php ?
>
> ping @Ocramius

Maks3w:
I was thinking in call this "picooptimizations" however IMO the code is more readable because the ternary operator was involved with other operators and its not enough clear what do without to read it three times.

Also the "lazy_services" key its better understood.

Maks3w:
On my commit bag I have a commit for move out the loops of abstract factory and delegators so the config method become very clean.

bakura10:
Speaking about pico-optimizations I wondered if we could optimize this: https://github.com/zendframework/zend-servicemanager/blob/develop/src/ServiceManager.php#L169-L170

test/LazyServiceIntegrationTest.php
```diff
@@ -215,12 +212,49 @@ public function testWillNotGenerateProxyClassFilesByDefault()
...
+        $container = new ServiceManager($config);
+        $instance  = $container->build(InvokableObject::class, ['foo' => 'bar']);
+        $this->assertInstanceOf(
+            InvokableObject::class,
+            $instance,
+            'Service returned does not extend ' . InvokableObject::class
+        );
+        $instance  = $container->build(\stdClass::class, ['foo' => 'bar']);
+        $this->assertInstanceOf(
+            \stdClass::class,
```
weierophinney: Import stdClass, if it isn't already.

test/LazyServiceIntegrationTest.php
```diff
@@ -215,12 +212,49 @@ public function testWillNotGenerateProxyClassFilesByDefault()
...
+            $instance,
+            'Service returned does not extend ' . \stdClass::class
+        );
+
+        $proxyAutoloadFunctions = $this->getRegisteredProxyAutoloadFunctions();
+        $this->assertCount(1, $proxyAutoloadFunctions, 'Only 1 proxy autoloader must be registered');
```
weierophinney: `s/must/should`/

test/Proxy/LazyServiceFactoryTest.php
```diff
@@ -0,0 +1,116 @@
...
+use ProxyManager\Proxy\LazyLoadingInterface;
+use ProxyManager\Proxy\VirtualProxyInterface;
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
+use Zend\ServiceManager\Factory\DelegatorFactoryInterface;
+use Zend\ServiceManager\Proxy\LazyServiceFactory;
+
+/**
+ * @covers \Zend\ServiceManager\Proxy\LazyServiceFactory
+ */
+final class LazyServiceFactoryTest extends TestCase
```
weierophinney: Why is the test case final?

Maks3w: Just for to replicate the same constraints as the original.

Maks3w: On components V3 I try to make the tests to follow the same hierarchy as the class under the test.

test/LazyServiceIntegrationTest.php
Maks3w: It's not and I thought was intentional https://github.com/Maks3w/zend-servicemanager/blob/feature/optimizations/test/LazyServiceIntegrationTest.php#L265

weierophinney: Nope, mistake on my part. Surprised that's working, tbh!

test/Proxy/LazyServiceFactoryTest.php
weierophinney: I guess my point is: I'm not sure why tests should *ever* be final. I understand it for library code, but for tests, I'm not sure it really has any meaning.

Maks3w:
@weierophinney done

PR #20, opened 2015-09-17
Optimization

bakura10:
Squeezing every little bit of performance...

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -166,7 +166,8 @@ class ServiceManager implements ServiceLocatorInterface
 
         $object = $this->doCreate($name);
 
-        if ($this->sharedByDefault ^ isset($this->shared[$name])) {
+        if (($this->sharedByDefault && !isset($this->shared[$name]))
+            || (isset($this->shared[$name]) && $this->shared[$name])) {
             $this->services[$name] = $object;
         }
 
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -166,8 +166,7 @@ class ServiceManager implements ServiceLocatorInterface
 
         $object = $this->doCreate($name);
 
-        if (($this->sharedByDefault && !isset($this->shared[$name]))
-            || (isset($this->shared[$name]) && $this->shared[$name])) {
+        if ($this->sharedByDefault ^ isset($this->shared[$name])) {
             $this->services[$name] = $object;
         }
 
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -104,20 +104,22 @@ class ServiceManagerTest extends TestCase
         $this->assertEquals('bar', $instance->foo);
     }
 
-    public function cacheProvider()
+    public function shareProvider()
     {
+        // @codingStandardsIgnoreStart
         return [
-            [true, true, true],
-            [true, false, false],
-            [false, false, false],
-            [false, true, true]
+            [true,  true,  true],  // [shared by default, service is explicitly shared => should be shared]
+            [true,  false, false], // [shared by default, service is explicitly NOT shared => should NOT be shared]
+            [false, false, false], // [NOT shared by default, service is explicitly NOT shared => should NOT be shared]
+            [false, true,  true]   // [NOT shared by default, service is explicitly shared => should be shared]
         ];
+        // @codingStandardsIgnoreEnd
     }
 
     /**
-     * @dataProvider cacheProvider
+     * @dataProvider shareProvider
      */
-    public function testCacheability($sharedByDefault, $serviceShared, $shouldBeSameInstance)
+    public function testShareability($sharedByDefault, $serviceShared, $shouldBeSameInstance)
     {
         $serviceManager = new ServiceManager([
             'shared_by_default' => $sharedByDefault,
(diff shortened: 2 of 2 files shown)


bakura10:
Tests are passing but I'm not sure of the logic.

src/ServiceManager.php
```diff
@@ -166,8 +166,7 @@ public function get($name)
 
         $object = $this->doCreate($name);
 
-        if (($this->sharedByDefault && !isset($this->shared[$name]))
-            || (isset($this->shared[$name]) && $this->shared[$name])) {
+        if ($this->sharedByDefault ^ isset($this->shared[$name])) {
```
Maks3w: Before `false` was a possible value of `isset($this->shared[$name]))` and was evaluated all as `false` now I think its evaluted as `true`

bakura10:
Yeah, this can be ignored. This does not work :D.

Maks3w:
Almost works

```
$a = true;
$b = null;

var_dump(($a && !isset($b)) || (isset($b) && $b));
var_dump((bool) ($a ^ isset($b)));

$a = false;
$b = null;

var_dump(($a && !isset($b)) || (isset($b) && $b));
var_dump((bool) ($a ^ isset($b)));

$a = true;
$b = false;

var_dump(($a && !isset($b)) || (isset($b) && $b));
var_dump((bool) ($a ^ isset($b)));

$a = false;
$b = false;

var_dump(($a && !isset($b)) || (isset($b) && $b));
var_dump((bool) ($a ^ isset($b)));

bool(true)
bool(true)
bool(false)
bool(false)
bool(false)
bool(false)
bool(false)
bool(true) // oops
```

bakura10:
Damn, that was a nearly good try :(.

weierophinney:
@bakura10 Just to be sure: Write some tests to verify each of the behaviors, and run them against the original and the changes.

bakura10:
It does not work, I just let it open if someone want to try it. I'm pretty sure it's possible to optimize it with some guru bitwise magic.

However it proves that the test coverage is not good enough, because my wrong fix was passing :).

Maks3w:
Current expression can be simplified a bit even could fix a bug if we understand `shared: null` should be managed as false.

```
$toShare = true;
$isShared = null;

var_dump(($toShare && !isset($isShared)) || (isset($isShared) && $isShared));
var_dump($toShare && (isset($isShared) && $isShared));

bool(true);
bool(false); 
```

bakura10:
I'm going to write additional tests so we'll be able to explore that a bit more.

bakura10:
I've added tests @weierophinney , they now make my wrong fix fail the test ;).

bakura10:
Done!

Maks3w:
Looks better

1. I see 3 variables with 2 possible values each one so the number of permutations should be 2**3 = 8. You wrote 6

2. Sort the data sets like this

000
001
010
011
100
101
110
111

Maks3w:
Pd. Add permutations for null values

bakura10:
It's not possible for shareByDzfzult to be null :)

Envoyé de mon iPhone

Maks3w:
Neither of the values its expected to be null but because the code use `isset` instead `array_key_exists` its better to cover that case safety

test/ServiceManagerTest.php
```diff
@@ -103,4 +103,49 @@ function ($container, $name, $callback) {
...
+
+        // @codingStandardsIgnoreStart
+        return [
+            '[shared by default, service is explicitly shared => should be shared]'             => [$sharedByDefault,  $serviceShared,  $serviceDefined, $shouldReturnSameInstance],
+            '[shared by default, service is explicitly NOT shared => should NOT be shared]'     => [$sharedByDefault,  !$serviceShared, $serviceDefined, !$shouldReturnSameInstance],
+            '[NOT shared by default, service is explicitly NOT shared => should NOT be shared]' => [!$sharedByDefault, !$serviceShared, $serviceDefined, !$shouldReturnSameInstance],
+            '[NOT shared by default, service is explicitly shared => should be shared]'         => [!$sharedByDefault, $serviceShared,  $serviceDefined, $shouldReturnSameInstance],
+            '[shared by default, service is not defined => should be shared]'                   => [$sharedByDefault,  $serviceShared,  !$serviceDefined, $shouldReturnSameInstance],
+            '[NOT shared by default, service not defined => should NOT be shared]'              => [!$sharedByDefault,  !$serviceShared, !$serviceDefined, !$shouldReturnSameInstance],
```
Maks3w: The main reason for put a description is when PHPUnit fails the only reference is the test case index which is not very useful, with the description you can take an idea about what was the context.

The expected result its implicit by the assertion so its redundant.

weierophinney: In this particular case, the assertion is not explaining the expectation well; it's testing equality of the passed expectation against what was received, but looking at the context, without the `=> ...` verbiage, it would be difficult to know what the expectation is. 

The current form was instantly readable by me, and I'd definitely understand what *failed* for any given item in the set. I'm going to merge as-is.

Maks3w: @weierophinney There are missing permutations. Please don't fall under the same asumptions as was do on InputFilter

weierophinney:
@Maks3w —

> None of the values its expected to be null

Because, internally, they *cannot* be `null`. They're booleans, and the only way that would change is to muck with the internals, which a *consumer* using the public API cannot do unless they use reflection. In that case, they deserve what they get.

test/ServiceManagerTest.php
Maks3w: 8 permuations are mandatory and 12 are desirable. I was waiting to @bakura10 complete the task before merge.

PR #21, opened 2015-09-17
Add test

bakura10:


bakura10:
My mistake, this was defined here: https://github.com/zendframework/zend-servicemanager/blob/develop/test/CommonServiceLocatorBehaviorsTrait.php

This trait is confusing, I didn't spot it at first sight.

PR #22, opened 2015-09-17
Improve code legibility

Maks3w:
This is made on top of #16 and the final result its a very readable config method.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -275,74 +275,106 @@ class ServiceManager implements ServiceLocatorInterface
         // For abstract factories and initializers, we always directly
         // instantiate them to avoid checks during service construction.
         if (isset($config['abstract_factories'])) {
-            foreach ($config['abstract_factories'] as $abstractFactory) {
-                if (is_string($abstractFactory) && class_exists($abstractFactory)) {
-                    $abstractFactory = new $abstractFactory();
-                }
-
-                if ($abstractFactory instanceof AbstractFactoryInterface) {
-                    $this->abstractFactories[] = $abstractFactory;
-                    continue;
-                }
-
-                // Error condition; let's find out why.
-
-                // If we still have a string, we have a class name that does not resolve
-                if (is_string($abstractFactory)) {
-                    throw new InvalidArgumentException(sprintf(
-                        'An invalid abstract factory was registered; resolved to class "%s" '
-                        . 'which does not exist; please provide a valid class name resolving '
-                        . 'to an implementation of %s',
-                        $abstractFactory,
-                        AbstractFactoryInterface::class
-                    ));
-                }
-
-                // Otherwise, we have an invalid type.
-                throw new InvalidArgumentException(sprintf(
-                    'An invalid abstract factory was registered. Expected an instance of "%s", '
-                    . 'but "%s" was received',
-                    AbstractFactoryInterface::class,
-                    (is_object($abstractFactory) ? get_class($abstractFactory) : gettype($abstractFactory))
-                ));
-            }
+            $this->resolveAbstractFactories($config['abstract_factories']);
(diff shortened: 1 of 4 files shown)


src/ServiceManager.php
```diff
@@ -241,97 +241,140 @@ public function has($name, $checkAbstractFactories = false)
...
-        $this->shared          = (isset($config['shared']) ? $config['shared'] : []) + $this->shared;
-        $this->aliases         = (isset($config['aliases']) ? $config['aliases'] : []) + $this->aliases;
-        $this->lazyServices    = array_merge_recursive($this->lazyServices, isset($config['lazy_services'])
-            ? $config['lazy_services']
-            : []);
-        $this->sharedByDefault = isset($config['shared_by_default'])
-            ? $config['shared_by_default']
-            : $this->sharedByDefault;
+        if (isset($config['services'])) {
+            $this->services = $config['services'] + $this->services;
```
danizord: Maybe `$this->services += $config['services'];` does the trick?

Maks3w:
That is one of the next steps.

src/ServiceManager.php
Maks3w: For to be honest I don't like to use the addition operator outside of arithmetic operations.

src/ServiceManager.php
```diff
@@ -241,97 +241,140 @@ public function has($name, $checkAbstractFactories = false)
...
-                        . 'which does not exist; please provide a valid class name resolving '
-                        . 'to an implementation of %s',
+    /**
+     * Instantiate abstract factories for to avoid checks during service construction.
+     *
+     * @param string[]|AbstractFactoryInterface[] $abstractFactories
+     *
+     * @return void
+     */
+    protected function resolveAbstractFactories(array $abstractFactories)
```
bakura10: Could be private.

src/ServiceManager.php
```diff
@@ -241,97 +241,140 @@ public function has($name, $checkAbstractFactories = false)
...
-                    $initializer = new $initializer();
-                }
+    /**
+     * Instantiate initializers for to avoid checks during service construction.
+     *
+     * @param string[]|callable[]|InitializerInterface[] $initializers
+     *
+     * @return void
+     */
+    private function resolveInitializers(array $initializers)
```
bakura10: This one is private ;). I'd say this is impl details and we should keep them private so that we can potentially modify them in the future without risking BC.

src/ServiceManager.php
Maks3w: true, I forgot to change that

src/ServiceManager.php
weierophinney: I agree with @bakura10 on this one.

test/LazyServiceIntegrationTest.php
```diff
@@ -22,7 +24,7 @@
 use ZendTest\ServiceManager\TestAsset\InvokableObject;
 
 /**
- * @covers \Zend\ServiceManager\Proxy\LazyServiceFactory
+ * @covers \Zend\ServiceManager\ServiceManager
```
weierophinney: Previous was correct; I wrote this to test usage of the LazyServiceFactory when used inside ServiceManager. I'll revert on merge.

Maks3w: NO

Maks3w: This test covers how ServiceManager process the settings (the missing 2.5% in the coverage)

Both `@covers` could be added but I already added a unit test for LazyServiceFactory so there is no need of to have both

Maks3w: Also this comment is out of place. This should has been discussed on #16. This PR only adds a commit to the stack

PR #24, opened 2015-09-18
Improve IDE completion

Maks3w:


bakura10:
Hhhmm... I'm not sure the extra dependency is worthwhile considering the few benefits it brings. I really prefer to use $this->assert. I find it clearer and it is more convenient for people familiar with PhpUnit.

weierophinney:
As I've mentioned in IRC previously, the trait is to be considered in the context of the class composing it, which means calling assertion methods directly on `$this` is perfectly valid and readable.

If your IDE is complaining about the file, add an exclusion for it.

Maks3w:
My IDE offers autocompletetions and suggestion where before not. Scrutinizer-CI and other type checkers tools are fine with this.


This was already done on zend-inputfilter and there was not problems. I don't understand why reject this.


Both ways are valid but this offers a noticeable gain.

bakura10:
What is the gain? I don't see any expect it makes it confusing because I don't know why this assert is here while it's as easy to type $this-> :)

PR #25, opened 2015-09-18
Add classblock to exceptions and missing throws tags

Maks3w:


src/Exception/InvalidArgumentException.php
```diff
@@ -11,6 +11,9 @@
 
 use InvalidArgumentException as SplInvalidArgumentException;
 
+/**
+ * {@inheritDoc}
```
danizord: I think it should be `@inheritDoc` (without braces), because you're not extending the docblock

Maks3w: There is no mention in the actual PHPDoc site neither PSR-5 this tag can be used standalone, only inline, so braces are required.

weierophinney: @Maks3w I've spoken to @mvriel about this, and you're wrong in this case.

`@inheritdoc` (without braces) should be used to inherit the full docblock. `{@inheritDoc}` is the inline form, and should be used to inherit specific portions of the docblock, based on contextual position:

```php
/**
 * {@inheritDoc} <-- inherits the short description.
 *
 * {@inheritDoc} <-- inherits the long description.
 *
 * {@inheritDoc} <-- inherits annotations.
 */
```

Maks3w: As you said in another thread. I'm not wrong. If the feature its not documented then "not exists"

Maks3w: Also if someone wants to change the CS rule that is a topic for a different PR because the form used here its the used across the whole codebase in zendframework repos

PR #26, opened 2015-09-18
Revert "Revert `@covers` annotation"

Maks3w:
This reverts commit e282b3d1f8be70edee430ae2b9516fa2582ff3b0.

### DIFF
--- a/test/LazyServiceIntegrationTest.php
+++ b/test/LazyServiceIntegrationTest.php
@@ -22,7 +22,7 @@ use Zend\ServiceManager\ServiceManager;
 use ZendTest\ServiceManager\TestAsset\InvokableObject;
 
 /**
- * @covers \Zend\ServiceManager\Proxy\LazyServiceFactory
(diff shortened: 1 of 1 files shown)


Maks3w:
It's magic "Coverage increased (+2.7%) to 100.0%"

https://coveralls.io/builds/3609502

weierophinney:
I hate that you essentially have to lie to PHPUnit about what your intentions are in order for it to recognize what is actually covered in the tests. *grrrr*

Maks3w:
As I described in the previous thread. Another approach could be 

```
/**
 * @covers \Zend\ServiceManager\Proxy\LazyServiceFactory
 * @covers \Zend\ServiceManager\ServiceManager
 */
```

So both classes are covered but what I did was create a whole test specific for the factory https://github.com/zendframework/zend-servicemanager/blob/develop/test/Proxy/LazyServiceFactoryTest.php

PR #27, opened 2015-09-21
Fix key name in aliased invokable migration doc

akrabat:


mwillbanks:
Since the migration document has been changed; this looks to be no longer applicable.

PR #29, opened 2015-09-29
Re-instate invokables

weierophinney:
This patch re-instates invokables, but not by re-adding the original functionality, but instead modifing the `configure()` method to create `factories` and `aliases` configuration from the `invokables` configuration, using the following rules:

- If the service name and class name match, it creates a `factories` entry, mapping the service name to an `InvokableFactory`.
- If the service name and class name *do not* match, it creates an `aliases` entry mapping the service name to the class name, and a `factories` entry, mapping the class name to an `InvokableFactory`.

This is exactly what was previously recommended in the migration guide; the patch simply automates it from the configuration.

Using this would remove the largest BC break for end-users.

### DIFF
--- a/doc/book/migration.md
+++ b/doc/book/migration.md
@@ -57,31 +57,10 @@ Configuration for v2 consisted of the following:
 ]
 ```
 
-In v3, the configuration remains roughly the same, with the following changes:
+In v3, the configuration remains the same, with the following additions:
 
 ```php
 [
-    'services' => [
-        // service name => instance pairs
-    ],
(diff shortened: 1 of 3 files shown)


weierophinney:
Ping @mwillbanks and @akrabat to review.

mwillbanks:
This looks good to me, I don't see anything visibly wrong.  The one part of the migration guide that might need to be pointed out is that if the user has extended the plugin manager and subsequently utilized the protected $invokables that has changed and is not BC.

PR #30, opened 2015-09-29
Re-instate config features

weierophinney:
This patch re-instates the `ConfigInterface` and `Config` class, with updates to follow the other changes in the component:

- `ConfigInterface::configureServiceManager()` is now specified as returning a *new* `ServiceManager` instance.
- `Config` now aggregates configuration and passes it to `ServiceManager::withConfig()`, returning the resulting cloned instance.

This was done due to the copius usage of both within components, to facilitate migration.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager;
+
+class Config implements ConfigInterface
+{
+    /**
+     * @var array
+     */
+    protected $config = [];
+
+    /**
+     * Constructor
+     *
+     * @param array $config
+     */
+    public function __construct($config = [])
+    {
+        $this->config = $config;
+    }
+
+    /**
+     * Configure service manager
+     *
+     * @param ServiceManager $serviceManager
+     * @return ServiceManager Returns a new instance with the merged configuration.
+     */
+    public function configureServiceManager(ServiceManager $serviceManager)
+    {
+        $config            = [];
+        $abstractFactories = isset($this->config['abstract_factories']) ? $this->config['abstract_factories'] : [];
+        $aliases           = isset($this->config['aliases']) ? $this->config['aliases'] : [];
+        $delegators        = isset($this->config['delegators']) ? $this->config['delegators'] : [];
+        $factories         = isset($this->config['factories']) ? $this->config['factories'] : [];
+        $initializers      = isset($this->config['initializers']) ? $this->config['initializers'] : [];
+        $invokables        = isset($this->config['invokables']) ? $this->config['invokables'] : [];
+        $lazyServices      = isset($this->config['lazy_services']) ? $this->config['lazy_services'] : [];
+        $services          = isset($this->config['services']) ? $this->config['services'] : [];
+        $shared            = isset($this->config['shared']) ? $this->config['shared'] : [];
+
+        if (! empty($abstractFactories)) {
+            $config['abstract_factories'] = $abstractFactories;
+        }
+
+        if (! empty($aliases)) {
+            $config['aliases'] = $aliases;
+        }
+
+        if (! empty($delegators)) {
+            $config['delegators'] = $delegators;
+        }
+
+        if (! empty($factories)) {
+            $config['factories'] = $factories;
+        }
+
+        if (! empty($initializers)) {
+            $config['initializers'] = $initializers;
+        }
+
+        if (! empty($invokables)) {
+            $config['invokables'] = $invokables;
+        }
+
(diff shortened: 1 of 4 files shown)


weierophinney:
Ping @akrabat and @mwillbanks — I found I needed this to facilitate migration within zend-mvc. Unfortunately, it also means I'll need to update a few more components in order to complete that task…

mwillbanks:
@weierophinney Ooof.  This one is a bit more BC heavy (from the standpoint of several applications that I've run in the past).  I do think that the changes make sense here (immutability) my only major concern is that the BC break here will likely affect many applications that have in the past dynamically changed the service manager but this seems like it would easily be migrated against.  @akrabat do you have any additional thoughts that would prevent us from merging otherwise I do believe this change is ready to go :+1:

src/Config.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Configure service manager
+     *
+     * @param ServiceManager $serviceManager
+     * @return ServiceManager Returns a new instance with the merged configuration.
+     */
+    public function configureServiceManager(ServiceManager $serviceManager)
+    {
+        $config            = [];
+        $abstractFactories = isset($this->config['abstract_factories']) ? $this->config['abstract_factories'] : [];
```
Ocramius: Giant method of CCLOC doom :|

src/Config.php
```diff
@@ -0,0 +1,86 @@
...
+        $delegators        = isset($this->config['delegators']) ? $this->config['delegators'] : [];
+        $factories         = isset($this->config['factories']) ? $this->config['factories'] : [];
+        $initializers      = isset($this->config['initializers']) ? $this->config['initializers'] : [];
+        $invokables        = isset($this->config['invokables']) ? $this->config['invokables'] : [];
+        $lazyServices      = isset($this->config['lazy_services']) ? $this->config['lazy_services'] : [];
+        $services          = isset($this->config['services']) ? $this->config['services'] : [];
+        $shared            = isset($this->config['shared']) ? $this->config['shared'] : [];
+
+        if (! empty($abstractFactories)) {
+            $config['abstract_factories'] = $abstractFactories;
```
Ocramius: Why not simply always define these keys?

src/Config.php
```diff
@@ -0,0 +1,86 @@
...
+
+        if (! empty($services)) {
+            $config['services'] = $services;
+        }
+
+        if (! empty($shared)) {
+            $config['shared'] = $shared;
+        }
+
+        return $serviceManager->withConfig($config);
```
Ocramius: Shouldn't this be some sort of static constructor instead?

src/Config.php
mwillbanks: I think the point is that the user land developer might not define all of those keys when configuring the service manager and as such could lead to some more BC breaks if we forced all of the keys to be defined.  A lot of the documentation and usage of the service manager has the aspect that you only pass in what you need.

src/Config.php
mwillbanks: Are you lobbying for something like the following:

```
public static function configureServiceManager(ServiceManager $sm, array $config = [])
...
```

src/Config.php
Ocramius: Nah, I mean just always define them in the `$config` variable in here, rather than adding keys to the array only when non-empty

src/Config.php
Ocramius: Nope: `class ServiceManager { ... public static function fromArrayConfig(array $config) {...} }`

weierophinney: @Ocramius No. The idea here is to configure an *existing* SM instance; you want to merge the new configuration with the *existing* configuration. As such, `withConfig()` is the appropriate method — but it means that you now have to capture the return value, as `withConfig()` returns a *new* instance.

src/Config.php
weierophinney: @Ocramius I'll make that change.

weierophinney:
@mwillbanks —

> This one is a bit more BC heavy (from the standpoint of several applications that I've run in the past).

There's two things going on here.

First, this is re-instating the `ConfigInterface`; the original BC break was its removal.

Second, the *new* BC break at this time is that we cannot re-introduce it with *the same behavior*, because one of the choices made in the refactor was to make the SM immutable. This was done for two reasons:

- To reduce complexity. We had a *lot* of code that was essentially checking against existing state to ensure changes wouldn't put the SM in an invalid state. Those can go away if the SM is immutable.
- To improve performance. By making it immutable, we can do some front-loading optimizations when we configure the original instance that ensure it's incredibly fast. For example: we resolve all aliases ahead of time (which is a *very* fast operation; it's essentially a map-reduce), vs. at runtime (which was slow, as it was a recursive lookup).

That means implementations *must* essentially call the SM's `withConfig()` and return the result, which is the *real* BC break in the interface definition. And believe me, I'm already feeling it as I attempt to upgrade zend-mvc; I've identified at least 4 components I have to update already. But it's in service to the above goals.

weierophinney:
@Ocramius Updated per your suggestion. Far simpler code now.

akrabat:
Is the new BC break obvious?

weierophinney:
I tried to call it out in the updated migration docs... Also, existing
configuration implementations will be breaking anyways as they formerly
used the various injection methods (setService, setFactory, etc).

PR #31, opened 2015-10-01
Add method for retrieving configuration to ConfigInterface

weierophinney:
Added `toArray()` method to `ConfigInterface` to provide a mechanism for fetching configuration without needing to configure the service manager. This will be useful in particular for `Zend\ModuleManager\Listener\ServiceListener`, which primarily needs to aggregate and merge configuration *prior* to creating a new instance of a plugin manager.

An implementation was added to the `Config` class, along with tests.

### DIFF
--- a/doc/book/migration.md
+++ b/doc/book/migration.md
@@ -159,6 +159,10 @@ class HelperConfig implements ConfigInterface
 }
 ```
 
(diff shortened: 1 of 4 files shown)


mwillbanks:
:+1:  This looks good to me and is also something that I've wanted in the past anyway.

@zendframework/community-review-team any additional thoughts?

PR #33, opened 2015-10-05
Cache by requested service name

weierophinney:
Previous behavior of the service manager allowed specifying aliases when creating the `shared` configuration. Prior to this patch, that did not work with the refactored code.

This patch updates the code to check for a cached service under the _requested_ service name, not the _resolved_ service name, which allows specifying sharing configuration based on aliases.

Internally, it caches up to twice:

- If the resolved service name is supposed to be shared, it will cache under the resolved name.
- If the requested service name is supposed to be shared, it will cache under the resolved name.

This allows having separate instances based on service name.

(Issue created due to observed issue when refactoring zend-view to use the dev-develop branch of zend-servicemanager.)

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -157,18 +157,32 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function get($name)
     {
+        $requestedName = $name;
         $name = isset($this->resolvedAliases[$name]) ? $this->resolvedAliases[$name] : $name;
 
-        // We start by checking if the service is cached (this is the fastest method).
-        if (isset($this->services[$name])) {
-            return $this->services[$name];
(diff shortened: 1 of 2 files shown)


src/ServiceManager.php
```diff
@@ -157,20 +157,34 @@ public function withConfig(array $config)
      */
     public function get($name)
     {
+        $requestedName = $name;
         $name = isset($this->resolvedAliases[$name]) ? $this->resolvedAliases[$name] : $name;
```
bakura10: Alignment.

weierophinney: I don't feel alignment is useful here, as the names and statements differ so much in length. Additionally, as it's only two such statements, it's less visibly useful.

PR #34, opened 2015-10-05
Added the first benchmarks

ezimuel:
This PR adds the first benchmarks for ServiceManager. I created a first round of benchmarking fetching the following services:
- factories;
- invokables;
- alias;
- services;
- abstract services;

These the results of the execution on a CPU Intel i5-2500 at 3.30GHz, 8 GB RAM, Ubuntu 14.04, PHP 5.5.9:
```
ZendBench\ServiceManager\FetchServices
    Method Name                   Iterations   Average Time      Ops/second   
    ---------------------------- ------------ ----------------- ---------------
    fetchFactoryService        : [     1,000] [0.0000220508575] [45,349.71023 ]
    fetchInvokableService      : [     1,000] [0.0000179929733] [55,577.25129 ]
    fetchService               : [     1,000] [0.0000026111603] [382,971.51205]
    fetchAliasService          : [     1,000] [0.0000124194622] [80,518.78443 ]
    fetchAbstractFactoryService: [     1,000] [0.0000263185501] [37,996.01406 ]
```

### DIFF
--- a/benchmarks/FetchServices.php
+++ b/benchmarks/FetchServices.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace ZendBench\ServiceManager;
+
+use Zend\ServiceManager\ServiceManager;
+use Athletic\AthleticEvent;
+
+class FetchServices extends AthleticEvent
+{
+    const NUM_SERVICES = 50;
+
+    /**
+     * @var ServiceManager
+     */
+    protected $sm;
+
+    public function setUp()
+    {
+        $this->sm = new ServiceManager();
+        // factories
+        for ($i = 0; $i < self::NUM_SERVICES; $i++) {
+            $this->sm->setFactory("factory_$i", 'ZendTest\ServiceManager\TestAsset\FooFactory');
+            $this->sm->setInvokableClass("invokable_$i", 'ZendTest\ServiceManager\TestAsset\Foo');
+            $this->sm->setService("service_$i", new \ZendTest\ServiceManager\TestAsset\Foo);
+            $this->sm->setAlias("alias_$i", "service_$i");
+        }
+        // abstract factory
+        $this->sm->addAbstractFactory('ZendTest\ServiceManager\TestAsset\FooAbstractFactory');
+        $this->sm->addAbstractFactory('ZendTest\ServiceManager\TestAsset\BarAbstractFactory');
+    }
+
+    /**
+     * Fetch the factory services
+     *
+     * @iterations 1000
+     */
+    public function fetchFactoryService()
+    {
+        $result = $this->sm->get(sprintf("factory_%d", rand(0, self::NUM_SERVICES - 1)));
+    }
+
+    /**
+     * Fetch the invokable services
+     *
+     * @iterations 1000
+     */
+    public function fetchInvokableService()
+    {
+        $result = $this->sm->get(sprintf("invokable_%d", rand(0, self::NUM_SERVICES - 1)));
+    }
+
+    /**
+     * Fetch the services
+     *
+     * @iterations 1000
+     */
+    public function fetchService()
+    {
+        $result = $this->sm->get(sprintf("service_%d", rand(0, self::NUM_SERVICES - 1)));
+    }
+
+    /**
+     * Fetch the alias services
+     *
+     * @iterations 1000
+     */
+    public function fetchAliasService()
+    {
+        $result = $this->sm->get(sprintf("alias_%d", rand(0, self::NUM_SERVICES - 1)));
+    }
+
+    /**
+     * Fetch the abstract factory services
+     *
+     * @iterations 1000
+     */
+    public function fetchAbstractFactoryService()
+    {
+       $result = $this->sm->get(rand(0,1) ? 'foo' : 'bar');
+    }
+}
--- a/benchmarks/FetchServices.php
+++ b/benchmarks/FetchServices.php
@@ -18,7 +18,7 @@ class FetchServices extends AthleticEvent
     {
         $this->sm = new ServiceManager();
         // factories
-        for ($i = 0; $i < self::NUM_SERVICES; $i++) {
+        for ($i = 0; $i <= self::NUM_SERVICES; $i++) {
             $this->sm->setFactory("factory_$i", 'ZendTest\ServiceManager\TestAsset\FooFactory');
             $this->sm->setInvokableClass("invokable_$i", 'ZendTest\ServiceManager\TestAsset\Foo');
             $this->sm->setService("service_$i", new \ZendTest\ServiceManager\TestAsset\Foo);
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -14,7 +14,8 @@
     },
     "require": {
         "php": ">=5.5",
-        "container-interop/container-interop": "~1.0"
+        "container-interop/container-interop": "~1.0",
+        "athletic/athletic": "dev-master"
```
Ocramius: Please move to `require-dev`

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Fetch the invokable services
+     *
+     * @iterations 1000
+     */
+    public function fetchInvokableService()
+    {
+        $result = $this->sm->get(sprintf("invokable_%d", rand(0, self::NUM_SERVICES - 1)));
```
Ocramius: Remove any unnecessary operations in here. `sprintf()` and the subtraction here, for example.

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Fetch the services
+     *
+     * @iterations 1000
+     */
+    public function fetchService()
+    {
+        $result = $this->sm->get(sprintf("service_%d", rand(0, self::NUM_SERVICES - 1)));
```
Ocramius: Remove any unnecessary operations in here. `sprintf()` and the subtraction here, for example.

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Fetch the alias services
+     *
+     * @iterations 1000
+     */
+    public function fetchAliasService()
+    {
+        $result = $this->sm->get(sprintf("alias_%d", rand(0, self::NUM_SERVICES - 1)));
```
Ocramius: Remove any unnecessary operations in here. `sprintf()` and the subtraction here, for example.

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Fetch the factory services
+     *
+     * @iterations 1000
+     */
+    public function fetchFactoryService()
+    {
+        $result = $this->sm->get(sprintf("factory_%d", rand(0, self::NUM_SERVICES - 1)));
```
Ocramius: Remove any unnecessary operations in here. `sprintf()` and the subtraction here, for example.

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * @var ServiceManager
+     */
+    protected $sm;
+
+    public function setUp()
+    {
+        $this->sm = new ServiceManager();
+        // factories
+        for ($i = 0; $i < self::NUM_SERVICES; $i++) {
```
Ocramius: Use `<=` to get rid of all the subtractions inside the test methods

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+            $this->sm->setFactory("factory_$i", 'ZendTest\ServiceManager\TestAsset\FooFactory');
```
Ocramius: `::class`

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+            $this->sm->setInvokableClass("invokable_$i", 'ZendTest\ServiceManager\TestAsset\Foo');
```
Ocramius: `::class`

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+            $this->sm->setService("service_$i", new \ZendTest\ServiceManager\TestAsset\Foo);
+            $this->sm->setAlias("alias_$i", "service_$i");
+        }
+        // abstract factory
+        $this->sm->addAbstractFactory('ZendTest\ServiceManager\TestAsset\FooAbstractFactory');
```
Ocramius: `::class`

benchmarks/FetchServices.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->sm->addAbstractFactory('ZendTest\ServiceManager\TestAsset\BarAbstractFactory');
```
Ocramius: `::class`

benchmarks/FetchServices.php
ezimuel: @Ocramius why the sprintf and the subtraction is unnecessary?

Ocramius: We're measuring the exact performance impact of *only* the `ServiceManager` operations: any function call that can be avoided and isn't part (*strictly*) of the tested API should be avoided, to reduce overhead in the measurement.

If the `ServiceManager` does 3 method calls to fetch a service, you already added 25% of the runtime of the benchmark body here ;-)

benchmarks/FetchServices.php
ezimuel: Why? The goal of a benchmark is to compare the performance of different commits/tags. We are not interested in absolute values, the operations inside each experiment are not relevant, if they are consistent between the comparisons.

benchmarks/FetchServices.php
ezimuel: See my comment below, this is completely irrelevant for the goal of a benchmark. We are not interested in absolute numbers, even if with some small overhead.

benchmarks/FetchServices.php
Ocramius: I'd still remove them for correctness, tbh.

ezimuel: @Ocramius I can change it because i'm diplomatic, not because it's wrong :)

Ocramius: @ezimuel I'm fine with having a test suite for tracking perf gains/losses over time, but we should still do it right, no? ;-)

ezimuel: @Ocramius you are not getting the point of benchmarking. There is not right or wrong here. Anyway, I'm changing the code according to your comments. As I said I'm very diplomatic. ;)

Ocramius: > There is not right or wrong here

There actually is :-P

Athletic is designed around test methods containing the minimum possible executable code to get the scenario working. It even goes far enough to run an empty benchmark for calibrating the results at the end, and removing that overhead from the calculations.

ezimuel: @Ocramius ok, I don't want to lose my time with this thread. You right!

ezimuel:
@Ocramius I removed the sprintf and the substraction in the fetch of the services. The overhead was between 1% and 4% before the previous benchmark.

Ocramius:
@ezimuel thanks!

ezimuel:
I'm closing this PR in favor of #38. I included all the feedbacks of @Ocramius. Thanks!

PR #35, opened 2015-10-05
Ensure plugin manager creation context travels with clone

weierophinney:
Prior to this patch, calling `withConfig()` on a plugin manager would reset the `creationContext` to the plugin manager itself — instead of retaining the original creation context. This patch overrides the `withConfig()` method to set the creation context to the original within the cloned instance.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -147,7 +147,7 @@ class ServiceManager implements ServiceLocatorInterface
     public function withConfig(array $config)
     {
         $container                  = clone $this;
-        $container->creationContext = $container;
+        $container->creationContext = ($this === $this->creationContext) ? $container : $this->creationContext;
         $container->configure($config);
         return $container;
     }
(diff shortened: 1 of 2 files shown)


weierophinney:
Ping @zendframework/community-review-team — yet another issue I encountered while updating the MVC.

mwillbanks:
Seems straight forward enough to me...  Although, are you not just moving the context down another layer?

bakura10:
Yes, can't you simply do it in the withConfig of the SM ?

Maks3w:
What about to overrride `configure()`? It's the common method for constructor and withConfig

weierophinney:
I was able to move the logic into the `ServiceManager::withConfig()` logic:

- If the creation context is the same as the original instance, we set the cloned creation context to the clone instance.
- Otherwise, we carry over the original creation context into the clone.

src/ServiceManager.php
```diff
@@ -147,7 +147,7 @@ public function __construct(array $config = [])
     public function withConfig(array $config)
     {
         $container                  = clone $this;
-        $container->creationContext = $container;
+        $container->creationContext = $this === $this->creationContext ? $container : $this->creationContext;
```
Maks3w: Please wrap the conditional with parentheses. Make more easy understand the operator priority

Maks3w:
:+1:

weierophinney:
@Maks3w Updated per your request, but I decided to squash the commits.

PR #43, opened 2015-10-25
Error in Quick-Start with build() method

RalfEggert:
If we request an instance with the `build()` method shouldn't we get a new instance for every call? If so, the comparison should print `false` instead of `true`, shouldn't it?


weierophinney:
Nice catch, @RalfEggert !

Issue #45, opened 2015-11-03
Thrown exceptions don't implement the container-interop interfaces


zerocrates:
The latest version of the ServiceManager advertises that it implements the container-interop interface, but it doesn't follow the spec rules on exceptions:

>Exceptions directly thrown by the container MUST implement the `Interop\Container\Exception\ContainerException`.
>
> A call to the get method with a non-existing id SHOULD throw a `Interop\Container\Exception\NotFoundException`.

Maks3w:
Can you provide a proof of violation? Did you verify the code on `develop` branch?

weierophinney:
This is definitely fixed on the develop branch; I'm not 100% certain on the master branch.

@zerocrates — would you be willing to create a PR for the master branch?

zerocrates:
Yes, to clarify, I'm talking about the master branch.

I can make a pull request if you'd like.

PR #46, opened 2015-11-04
Use container-interop exception interfaces

zerocrates:
Fix for #45

### DIFF
--- a/src/Exception/ExceptionInterface.php
+++ b/src/Exception/ExceptionInterface.php
@@ -9,6 +9,8 @@
 
 namespace Zend\ServiceManager\Exception;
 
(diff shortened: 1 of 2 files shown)


Maks3w:
The only thing I see `ExceptionInterface` is used too for `LogicalException` like `InvalidArgumentException` and these exception shouldn't be caught.

zerocrates:
Well, I was following the pattern from develop, which does the same thing (`InvalidArgumentException` implements `ExceptionInterface` which extends the interop `ContainerException`). I think that's the most faithful interpretation of the spec as well: it says all exceptions thrown by the container "MUST" implement `ContainerException`.

PR #48, opened 2015-11-04
Remove the need for the $checkAbstractFactories flag

weierophinney:
In updating zend-mvc to v3 of zend-servicemanager, I ran into a subtle issue: `has()` does not look in abstract factories by default, and requires that you pass an additional flag, `$checkAbstractFactories`, with a boolean true value, if you want to test if an abstract factory can resolve the given `$name`.

This breaks the Liskov Substitution Principal, as it means that the behavior of `has()` depends on how the `ServiceManager` is configured. Users then must be aware that an abstract factory might supply a dependency, and cannot rely on `has()` per the container-interop contract to work for the specific use case of zend-servicemanager.

As such, I'd like to remove the flag, and check abstract factories whenever a service or factory matching the provided `$name` is not found.


weierophinney:
Closing; I opened against the wrong branch.

bakura10:
I can't fight against a breakage of the Liskov law =).

More seriously, no problem with that. Best practices encourage not to use many abstract factories so the introduced overweight should be negligeable. Go on! ;)

PR #49, opened 2015-11-04
Removed $checkAbstractFactories flag from has()

weierophinney:
In updating zend-mvc to v3 of zend-servicemanager, I ran into a subtle issue: `has()` does not look in abstract factories by default, and requires that you pass an additional flag, `$checkAbstractFactories`, with a boolean true value, if you want to test if an abstract factory can resolve the given `$name`.

This breaks the Liskov Substitution Principal, as it means that the behavior of `has()` depends on how the ServiceManager is configured. Users then must be aware that an abstract factory might supply a dependency, and cannot rely on `has()` per the container-interop contract to work for the specific use case of zend-servicemanager.

As such, this patch removes the flag, and checks abstract factories whenever a service or factory matching the provided `$name` is not found.

### DIFF
--- a/doc/book/migration.md
+++ b/doc/book/migration.md
@@ -428,42 +428,6 @@ $alsoBetween = $container->build(Between::class, [
 The above two validators would be different instances, with their own
 configuration.
 
-### has() no longer checks abstract factories by default
-
-In v2, `has()` would also check abstract factories to see if any would match the
-service. Depending on the number of abstract factories present, this can be an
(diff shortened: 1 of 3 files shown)


weierophinney:
Ping @bakura10 … (ignore #48)

bakura10:
+1 for this change. This makes the signature coherent with container-interop contract, and as said in the other, the performance penalty will be negligeable anyway.

PR #47, opened 2015-11-04
Remove immutability from v3 changes

weierophinney:
In refactoring zend-modulemanager, zend-mvc, zend-view, and components with view helper configuration, I've been having to do some rather ugly hacks:

- https://github.com/zendframework/zend-modulemanager/commit/8de7f22e761c90779a302f8dd14dc28821e7e498
- https://github.com/zendframework/zend-mvc/commit/a46be7d9e9512f3f624e36f840e7993b268639d1

These are due to the fact that the v3 refactor uses immutability as a core feature of the ServiceManager.

I'm starting to question the decision:

- It's forcing some BC breaks that would largely be unnecessary otherwise.
- It's forcing a situation where the only way to configure plugin managers managed by the SM is via application-level configuration. In other words, you cannot add plugins to plugin managers *unless* you do so via application configuration.

I've been doing a little testing, and I've found that the current `configure()` logic is actually quite lightweight from a resource perspective: it is not costly in terms of memory, CPU cycles, or time, and, because it takes care of things like alias resolution and abstract factory instantiation, removes the state-based issues we had in v2.

This patch does the following:

- Removes `withConfig()`.
- Makes `configure()` public.
- Re-instates each of the following methods, which now all proxy to `configure()`:
  - `setAlias()`
  - `setFactory()`
  - `addAbstractFactory()`
  - `addDelegator()`
  - `addInitializer()`
  - `setService()`
  - `setShared()`
- Re-instates the `setAllowOverride()` and `getAllowOverride()` methods. When disabled, `configure()` (and hence all methods that proxy to it) will raise a `ContainerModificationsNotAllowedException`.
- Adds a `mapLazyService($name, $class = null)` method, proxying to `configure()` to allow mapping lazy services after instantiation.

Tests were added to cover the new behavior, and documentation updated (including migration documentation).


weierophinney:
Patch is ready for review! Pinging @ezimuel !

doc/book/configuring-the-service-manager.md
```diff
@@ -484,14 +484,31 @@ This works because the `InvokableFactory` will automatically pass the options
...
-factories. To do that, you can use the `withConfig()` method. This method creates a
-*new* service manager instance, which merges its own configuration with the new
-configuration you provide to it:
+Assuming that you have not called `$container->setAllowOverride(false)`, you can,
+at any time, configure the service manager with new services using any of the
+following methods:
+
+- `configure()`, which accepts the same configuration array as the constructor.
+- `setAlias($alias, $target)`
+- `setInvokableClass($name, $class = null)`; if no `$class` is passed, the
```
bakura10: Do we really need all those methods ? Just having configure is not enough?

weierophinney:
One note: I re-ran the Athletic benchmarks, on both the current develop branch and the one on which this PR is based; there's zero difference!

doc/book/configuring-the-service-manager.md
weierophinney: The methods all proxy to `configure()`, and reduce BC breaks. I think that's a win.

src/ServiceManager.php
```diff
@@ -265,10 +273,18 @@ public function has($name, $checkAbstractFactories = false)
...
      * @param  array $config
-     * @return void
+     * @return self
+     * @throws ContainerModificationsNotAllowedException if the allow
+     *     override flag has been toggled off.
      */
-    protected function configure(array $config)
+    public function configure(array $config)
     {
+        if ($this->allowOverride === false) {
```
bakura10: oh no... please don't bring back that :(.

src/ServiceManager.php
```diff
@@ -326,6 +342,109 @@ protected function configure(array $config)
         }
 
         $this->resolveAliases();
+
+        return $this;
```
bakura10: @Ocramius to the rescue.

src/ServiceManager.php
```diff
@@ -326,6 +342,109 @@ protected function configure(array $config)
...
+    }
+
+    /**
+     * Add an invokable class mapping.
+     *
+     * @param string $name Service name
+     * @param null|string $class Class to which to map; if omitted, $name is
+     *     assumed.
+     */
+    public function setInvokableClass($name, $class = null)
```
bakura10: This naming was (and is still) inconsistent to me. setInvokableClass/setFactory/setFactoryClass ?

weierophinney: Again, the only reason I'm re-adding these is because it's an opportunity
to remove a BC break. Changing the names at this point would be a pointless
break.

src/ServiceManager.php
weierophinney: This is here to simplify updated in components where we were previously
calling withConfig after the first round of refactoring. Without this, we
are likely to make mistakes when we update again.

src/ServiceManager.php
weierophinney: I'm going back and forth on this one. It's available in v2. What's interesting is that it's *disabled* by default (i.e., you cannot override). In looking through the code, when allow override is disabled, we essentially do checks on service registration to see if the service exists and/or can be created, and then raise an exception. The check I'm doing is a naive check on `configure()`, and essentially makes it completely immutable at that point.

I guess the question at this point is: do we want the ability to lock the instance down, or not? If not, I'll remove it.

src/ServiceManager.php
weierophinney: @bakura10 Also, there *is* a semantic difference. With `setInvokableClass()`, the idea is that you're registering a *class* **as** a *service name* (this is more clear in the version I'm proposing here, as you can omit the `$class` argument entirely, which was not true in v2); the action the container needs to take is only to instantiate it on first request. With `setFactory()`, we're mapping a factory to a service name; we're not registering a factory class *as* the service, but to *create* the service. As such, I don't see an inconsistency here; they're different.

weierophinney:
I've modified the behavior of the "allow overrides" flag to more closely mimic that of v2 at this point; i.e., if an alias, invokable, factory, delegator, or sharing rule is added, allow overrides is false, and an *instance* of the named service is already present (the provided service name is mapped in the `$services` table), then an exception will be raised. Otherwise, the operation is permitted.

In discussion with @ezimuel, he suggested potentially more fine-grained settings, with a flag per service indicating ability to override. I think this could be done at a later date without BC ramifications (as the default would be still to have that disallowed by default).

While I know @bakura10 is not a fan of the flag, it had specific use cases in v2, particularly to prevent ambiguity about what happens when a change in the state is made. The patch presented in b253d70 keeps the performance improvements introduced for v3, while retaining those protections, and keeping the v2 bootstrap workflow usable.

Unfortunately, coveralls won't show me the changes in coverage, and I'm unsure how to improve it at this time (as I have introduced tests for each of the new execution paths created). I think the change is acceptable at this time.

bakura10:
Okay... You are allowed to change it then =).

weierophinney:
@bakura10 :fist: — glad that will work! It's a quite inexpensive check, in the end!

Issue #50, opened 2015-11-06
ServiceManager returns that service does not exist with has() but can create with get()


svycka:
I created service manager like this:
```php
use Zend\ServiceManager\AbstractPluginManager;
use Zend\ServiceManager\Exception\RuntimeException;

class TypesManager extends AbstractPluginManager
{
    protected $invokableClasses = [
        InArrayType::class => InArrayType::class,
        RegexType::class   => RegexType::class,
    ];
    protected $aliases = [
        'in_array' => InArrayType::class,
        'regex'    => RegexType::class,
    ];

    public function validatePlugin($type)
    {
        if ($type instanceof SettingTypeInterface) {
            return; // we're okay
        }
        throw new RuntimeException(sprintf(
            'SettingType of type %s is invalid; must implement %s',
            (is_object($type) ? get_class($type) : gettype($type)),
            SettingTypeInterface::class
        ));
    }
}
```

I am using like this:
```php
$manager = new TypesManager();
$manager->has(InArrayType::class); // returns false
$manager->has('in_array'); // returns false(canonicalization problem because of `_` ?)
$manager->has(RegexType::class); // returns false
$manager->has('regex'); // returns true
$manager->get(InArrayType::class); // returns instance off InArrayType::class
$manager->get('in_array'); // // returns instance off InArrayType::class
$manager->get(RegexType::class); // returns instance off RegexType::class
$manager->get('regex'); // // returns instance off RegexType::class
```
when I sow this I thought maybe I doing it wrong so checked other zend modules like zend-filter and wrote some tests:
```php

    public function testFirstGet()
    {
        $manager = new FilterPluginManager();
        $this->assertInstanceOf('Zend\Filter\Boolean', $manager->get('Zend\Filter\Boolean'));
        $this->assertTrue($manager->has('Zend\Filter\Boolean'));
    }
    public function testFirstHas()
    {
        $manager = new FilterPluginManager();
        $this->assertTrue($manager->has('Zend\Filter\Boolean')); // fails here
        $this->assertInstanceOf('Zend\Filter\Boolean', $manager->get('Zend\Filter\Boolean'));
    }
```
First pass second fails.
order of has() and get() is important?

demichl68:
Regarding your tests:

The first one passes because after calling "get", the service is added to your invokable services (autoAddInvokableClass).

Your second case fails, because the service has not been added to your invokables when you call "has"

svycka:
[...]

PR #51, opened 2015-11-06
get and has order is important

svycka:
if has() returns false and works as expected then how can I check if manager can give me and an instance

### DIFF
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -1226,4 +1226,18 @@ class ServiceManagerTest extends TestCase
         $this->assertSame($expectedService, $peeredServiceManager->get('peered_service'));
         $this->assertSame($expectedService, $secondParentServiceManager->get('peered_service'));
     }
+
+    public function testFirstGet()
+    {
+        $manager = new \Zend\Filter\FilterPluginManager();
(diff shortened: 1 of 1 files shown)


svycka:
more on this zendframework/zend-servicemanager#50

weierophinney:
I'm sorry, but this is by design.

`has()` only checks the following sources:

- services already registered or created
- known invokable mappings
- known factory mappings
- all abstract factories

The "auto-add invokable classes" flag is specific to the plugin manager, and the assumption is that *any* fully qualified class name can be passed to the plugin manager, and it will attempt to create an instance. However, this is not the same as *having* an instance already; it's an opportunistic approach that *could fail* (e.g., if the constructor has required arguments!).

As such, having `has()` report true for such instances can and would be misleading — and far moreso than having it report false.

If you want equivalent behavior, but also want `has()` to report `true`, you could do the following:

- create an abstract factory, WHERE
- `canCreateServiceWithName()` returns `true` when `class_exists($requestedName)`; THEN
- change the `$autoAddInvokableClass` property to `false`, AND
- register your abstract factory with your plugin manager.

This will have the effect of having `has()` return true if the class exists.

PR #54, opened 2015-11-12
 Added missing withConfig(array $config) method in ServiceManager class 

samsonasik:
It documented in doc migrations:

```php
        return $serviceManager->withConfig([
            'invokables' => $this->invokables,
        ]);
```
but there is no method `withConfig()` in ServiceManager class.

/cc @weierophinney

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -141,6 +141,19 @@ class ServiceManager implements ServiceLocatorInterface
     }
 
     /**
+     * Configure ServiceManager
+     *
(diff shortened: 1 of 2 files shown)


samsonasik:
the `withConfig()` is used in zend-mvc dev-develop branch: https://github.com/zendframework/zend-mvc/blob/develop/src/Application.php#L265-L267

Xerkus:
@samsonasik take a look at #47

samsonasik:
Ok , then zend-mvc part need to be updated. Thank you ;)

Issue #57, opened 2015-12-04
canonicalizeName(): use str_replace instead of strtr function


Colmea:
Hello,

After profiling my zf2 app, I saw that strtr() function in ServiceManager::canonicalizeName() was a little bit "too expensive": 3.41MB memory (48% of the memory used), almost 4ms CPU time (5% of the total time).
strtr function is actually called 500 times (in my app) from canonicalizeName method.

After some tests, it seems that str_replace function is far more efficient than strtr (in this particular case at least): 132KB for memory and 704µs for CPU time.

I'm not a suporter of "over-optimization", but I think this little enhancement could be interesting

I can create a PR if you want.

#### My app
Benchmark on a working zf2 app (499 services use the *strtr* function).

<img src="https://box.everhelper.me/attachment/343693/uCmUcdQtHYtrtMyLtACNK5r5XzgC2WmN/528539-CqrFXxEUTtCRqJQ1/screen.png" width="100" /> **VS** <img src="https://box.everhelper.me/attachment/343709/uCmUcdQtHYtrtMyLtACNK5r5XzgC2WmN/528539-pitzUfPlRbwe59Mx/screen.png" width="100" />

#### strtr() vs str_replace()
Isolated benchmark of str and str_replace functions. Almost 4 millions iterations. Use services string from existing zf2 app.

<img src="https://box.everhelper.me/attachment/343845/uCmUcdQtHYtrtMyLtACNK5r5XzgC2WmN/528539-ckphAUQb6OpmZQ8H/screen.png" width="100" /> **VS** <img src="https://box.everhelper.me/attachment/343846/uCmUcdQtHYtrtMyLtACNK5r5XzgC2WmN/528539-zEVUc3ugC08GpQTH/screen.png" width="100" />

----

Arthur Schwaiger

Maks3w:
The first example is not conclusive because one method was called half of times than other. So seems to be a comparison with different scenarios.

Please send a PR with your proposal and we'll review if its safe to make that replacement.

Colmea:
Yep, my bad. The screenshot is not good. Actually the 3.8ms are 95% used by strtr() from canonicalizeName() method.

I will send a PR :)

Ocramius:
Interesting... I remember removing str_replace and using strtr for similar
(opposite) reasons. Did PHP just get better at it?

Maks3w:
[...]

PR #58, opened 2015-12-07
canonicalizeName(): use str_replace  instead of strtr function (#57)

Colmea:
See #57 for more infos.

Use *str_replace* function instead of *strtr* in the canonicalizeName() method (for performance).

We need to check if it's *really* faster.

Tested platform:
- Zend framework v2.4 - PHP 5.5.9-1ubuntu4.14

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -112,9 +112,9 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
     protected $throwExceptionInCreate = true;
 
     /**
-     * @var array map of characters to be replaced through strtr
+     * @var array map of characters to be replaced through str_replace
      */
-    protected $canonicalNamesReplacements = ['-' => '', '_' => '', ' ' => '', '\\' => '', '/' => ''];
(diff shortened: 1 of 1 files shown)


weierophinney:
I've run locally, and what I observe is that:

- for fetching factories, invokables, and aliased services, it's around 15% faster.
- for fetching service instances and services via abstract factories, it's around 10% *slower*

The former combined with the latter means that the initial fetch of a service will be faster, but all subsequent requests will be slower. If it's never retrieved again, or only retrieved once, that's overall faster still, but could be an issue. In the case of abstract factories, it means a slowdown across the board, however.

I tested using the benchmarks already in the master branch (`./vendor/bin/athletic -p benchmarks`), and saw the same results with both PHP 5.5 and 5.6. For v7, I noticed that your patch actually is typically performing at the same speed or *slower* than the master branch across all service types!

As such, and considering that we're about to release v3, which has *substantial* improvements, and which eliminates canonicalization entirely, I'm going to close this without merging, as the results lead to overall negative performance changes.

PR #59, opened 2016-01-01
Forward compatibility features

weierophinney:
This patch provides a number of features to allow us to provide forwards compatibilty in v2, as well as backwards compatibility with v2 code.

## AbstractFactoryInterface

This patch renames the `canCreateServiceWithName()` method to `canCreate()`.  This ensures there are no overlapping methods between the v2 and v3 interfaces.

## Re-introduction of v2 interfaces

The v3 factory and initializer interfaces were all pushed under new namespaces, and, as such, have zero overlap with the existing v2 interfaces. However, the v2 interfaces cannot be used currently with v3, meaning any factories or initializers that implemented the interfaces (vs using callables) are de facto incompatible with v3.

This patch reintroduces the v2 interfaces, and modifies them to inherit from the v3 interfaces. This provides a forwards compatibility feature: developers can implement the new methods from the v3 interfaces in their existing v2 code, and thus ensure that their factories and initializers will continue to work on upgrading to v3.

As an example, consider the following factory implementation for v2:

```php
use Zend\ServiceManager\FactoryInterface;
use Zend\ServiceManager\ServiceLocatorInterface;

class SomethingFactory implements FactoryInterface
{
    public function createService(ServiceLocatorInterface $services)
    {
        return new Something($services->get('SomethingElse'));
    }
}
```

To prepare this for v3 compatibility, you could do the following:

```php
use Interop\Container\ContainerInterface;
use Zend\ServiceManager\FactoryInterface;
use Zend\ServiceManager\ServiceLocatorInterface;

class SomethingFactory implements FactoryInterface
{
    public function createService(ServiceLocatorInterface $services)
    {
        return $this($services, 'Something');
    }

    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
    {
        return new Something($container->get('SomethingElse'));
    }
}
```

In v2, `createService()` will be called, and proxy to `__invoke()`; on an upgrade to v3, only `__invoke()` will be called. Developers can then later update the factory to read as the following:

```php
use Interop\Container\ContainerInterface;
use Zend\ServiceManager\Factory\FactoryInterface;

class SomethingFactory implements FactoryInterface
{
    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
    {
        return new Something($container->get('SomethingElse'));
    }
}
```

The re-introduced interfaces include:

- `Zend\ServiceManager\AbstractFactoryInterface` (extends `Zend\ServiceManager\Factory\AbstractFactoryInterface`)
- `Zend\ServiceManager\DelegatorFactoryInterface` (extends `Zend\ServiceManager\Factory\DelegatorFactoryInterface`)
- `Zend\ServiceManager\FactoryInterface` (extends `Zend\ServiceManager\Factory\FactoryInterface`)
- `Zend\ServiceManager\InitializerInterface` (extends `Zend\ServiceManager\Initializer\InitializerInterface`)

## AbstractPluginManager

The `AbstractPluginManager` suffered a number of BC breaks, and this patch attempts to address each of them.

First, the `$autoAddInvokableClass` flag was re-introduced, along with its behavior. This allows fetching a class by FQCN from a plugin manager, without it first being registered. As plugin managers include instance validation, this is a safe operation.

Second, the `validate()` method was updated to perform a check for a `validatePlugin()` method, which was an `abstract` method defined in v2. If present, the `validate()` method will proxy to it, after first raising an `E_USER_DEPRECATED` notice prompting the user to rename the method.

Third, the constructor was updated to allow overloading of the first argument.  In v2, the constructor expects a null or `ConfigInterface` first argument, and many tests and factories already exist that operate on this assumption. As such, the constructor now supports the following combinations, with the associated behavior:

- A single `ContainerInterface` argument (v3 behavior); sets the `$creationContext` to the argument.
- A `ContainerInterface` argument and an array of configuration (v3 behavior); sets the `$creationContext` to the first argument, and uses the second to configure the plugin manager instance.
- No arguments (v2 behavior); sets the `$creationContext` to the plugin manager instance, but raises an `E_USER_DEPRECATED` notice.
- A `ConfigInterface` first argument (v2 behavior); uses the argument to configure the plugin manager, sets the `$creationContext` to the plugin manager, and raises deprecation notices for both conditions.

Additionally, similar changes were proposed for v2 in #60, ensuring a migration path for users.

Fourth, the `setServiceLocator()` and `getServiceLocator()` methods were reintroduced, though without the related `ServiceLocatorAwareInterface`. This was done as follows:

- `AbstractPluginManager` now defines `setServiceLocator()`, and uses it to set the `$creationContext`, but raises an `E_USER_DEPRECATED` notice when doing so. This is done to facilitate existing tests and plugin manager factories.
- `ServiceManager` (from which `AbstractPluginManager` inherits) defines `getServiceLocator()`, and returns the `$creationContext`, while raising an `E_USER_DEPRECATED` notice.

This latter change requires explanation. Essentially, many plugin factories look like the following:

```php
function ($plugins)
{
    $services = $plugins->getServiceLocator();
    return new Plugin($services->get('SomeDependency'));
}
```

In v3, the first argument passed to a factory is actually the parent locator/creation context, analagous to `$services` in the above example. In order to ensure that existing factories will continue to work, we had to add the `getServiceLocator()` method to the `ServiceManager`, as it will generally be an instance of that class being passed to the factory. This makes the call redundant, but ensures it continues to work. By raising the deprecation notice, we can signal developers to upgrade their factories.

## Documentation

The migration documentation has been updated to reflect these changes, highlighting how to prepare v2 code to work under v3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,10 @@ All notable changes to this project will be documented in this file, in reverse 
   Of course, you can still override the `validate` method if your logic is more
   complex.
 
+  To aid migration, `validate()` will check for a `validatePlugin()` method (which
+  was required in v2), and proxy to it if found, after emitting an
+  `E_USER_DEPRECATED` notice prompting you to rename the method.
+
 - A new method, `configure()`, was added, allowing full configuration of the
   `ServiceManager` instance at once. Each of the various configuration methods —
   `setAlias()`, `setInvokableClass()`, etc. — now proxy to this method.
@@ -168,8 +172,13 @@ changes, outlined in this section.
   enforced through the interface, that allows you to easily map multiple service
   names to the same factory.
 
+  To provide forwards compatibility, the original interfaces have been retained,
+  but extend the new interfaces (which are under new namespaces). You can implement
+  the new methods in your existing v2 factories in order to make them forwards
+  compatible with v3.
+
 - The for `AbstractFactoryInterface` interface renames the method `canCreateServiceWithName()`
-  to `has()`, and merges the `$name` and `$requestedName` arguments.
+  to `canCreate()`, and merges the `$name` and `$requestedName` arguments.
 
 - Plugin managers will now receive the parent service locator instead of itself
   in factories. In version 2.x, you needed to call the method
@@ -212,12 +221,20 @@ changes, outlined in this section.
   In practice, this should reduce code, as dependencies often come from the main
   service locator, and not the plugin manager itself.
 
+  To assist in migration, the method `getServiceLocator()` was added to `ServiceManager`
+  to ensure that existing factories continue to work; the method emits an `E_USER_DEPRECATED`
+  message to signal developers to update their factories.
+
 - `PluginManager` now enforces the need for the main service locator in its
   constructor. In v2.x, people often forgot to set the parent locator, which led
   to bugs in factories trying to fetch dependencies from the parent locator.
   Additionally, plugin managers now pull dependencies from the parent locator by
   default; if you need to pull a peer plugin, your factories will now need to
   pull the corresponding plugin manager first.
+
+  If you omit passing a service locator to the constructor, your plugin manager
+  will continue to work, but will emit a deprecation notice indicatin you
+  should update your initialization code.
 
 - It's so fast now that your app will fly!
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,6 +168,9 @@ changes, outlined in this section.
   enforced through the interface, that allows you to easily map multiple service
   names to the same factory.
 
+- The for `AbstractFactoryInterface` interface renames the method `canCreateServiceWithName()`
+  to `has()`, and merges the `$name` and `$requestedName` arguments.
+
 - Plugin managers will now receive the parent service locator instead of itself
   in factories. In version 2.x, you needed to call the method
   `getServiceLocator()` to retrieve the parent (application) service locator.
--- a/doc/book/configuring-the-service-manager.md
+++ b/doc/book/configuring-the-service-manager.md
@@ -170,7 +170,7 @@ implement `Zend\ServiceManager\Factory\AbstractFactoryInterface`:
 
 class MyAbstractFactory implements AbstractFactoryInterface
 {
-    public function canCreateServiceWithName(ContainerInterface $container, $requestedName)
+    public function has(ContainerInterface $container, $requestedName)
     {
         return in_array('Traversable', class_implements($requestedName), true);
     }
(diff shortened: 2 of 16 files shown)


weierophinney:
Ping @zendframework/community-review-team, and particularly @ezimuel and @akrabat. I think the changes proposed in here will significantly lessen the migration impact for v3. As noted, still a WIP, as I want to figure out how to make plugin managers forwards compatible as well.

bakura10:
I'm not sure to understand the rationale for the has method. It is quite confusing and does not represent at all what the abstract factory is doing. It is not checking if it has something, it just says if it can create something. The naming seems wrong.

What did you mean by the mirror container usage? Are you considering an abstract factory as a container ? ;o

For the rest of the PR, I don't like those new interfaces. It clutters the library with redundant interfaces for a new major version, but I suppose that BC is the most important here =).

weierophinney:
Would `exists()` make sense to you, @bakura10? Other options I can think of:

- canCreate
- instantiable
- mappable
- handles

I'm flexible on the naming. The reason I suggested has() is because I see a
possibility of using abstract factories as peering or delegate containers
in pimple or other container implementations.

As to the reintroduction of the old interfaces, note that they are all
marked deprecated. They literally exist only to provide a migration
strategy, and we can even indicate a release and or date by which they'll
be removed. But yes; they are an essential strategy for the migration
story. I would love for us NOT to need to release v3 versions of all
components just to accommodate the SM changes; if we can provide forward
compatible v2 releases, we avoid version inflation, and allow those
components to bump only when they have actual BC breaking changes.

bakura10:
I definitely prefer "canCreate". This better represents what this method is supposed to do.

Okay, no problem for the compat, but as you said, as all the components will need a plugin manager, it will be very hard to create a new version of those components without tagging a v3 (and if we need, then we would not need those compat interfaces). I suppose creating and maintaining a v3 and v4 of Service manager is out of the question?

weierophinney:
Ping @zendframework/community-review-team — need a review! This patch is largish, but I've tried to document it in both the patch description and the documentation. Essentially, this patch provides a way for us to prevent the need to tag v3 versions of components that use the service manager; they should, with minor additions, be compatible with v3 following this changeset (as well as the one proposed in #60, which is for the v2 series).

sandrokeil:
@weierophinney Make it sense to create a bridge repo `zend-servicemanager-2x-bridge` instead? I like this approach, but I would prefer an extra repo if possible.

weierophinney:
@sandrokeil Putting the code in a different repository complicates the migration strategy significantly. To clarify, consider the steps a user must take with that approach:

- User makes code changes to existing code to make it forwards compatible
- User upgrades to v3
- User runs tests/runs CI/runs application/whatever
- Fatal errors occur
- User must search for the source of the errors, and hopefully discovers they're due to the upgrade
- User must find migration documentation to determine what needs to happen
- User must add the migration package to their application
- Eventually, user needs to make code changes in order to work cleanly with v3
- User must remove migration package from their application once migration is done

With the proposed changes, this becomes:

- User makes code changes to existing code to make it forwards compatible
- User upgrades to v3
- User runs tests/runs CI/runs application/whatever; it *just works*. Some features raise deprecation notices indicating what changes will need to be made *in the future*.
- Eventually, user needs to make code changes in order to work cleanly with v3
- An eventual update of the servicemanager package removes the features for them.

Half the steps, and a process that puts less onus on the user to *search* for what to do; it instead *guides* them.

By marking the code as deprecated, we provide a clear signal that users cannot and should not rely on deprecated features. The only way I think we can improve on that is to give a release and/or date by which those features will be removed so that they know the deadline for upgrading their code.

sandrokeil:
Yes, it's much simpler if the code is in this repo. A deadline is a good idea, because the classes are deprecated on release and people can create an issue with a deadline in their issue tracker. :+1:

weierophinney:
I've provided additional documentation detailing specific migration steps you can take to make your version 2 factories, intializers, and plugin managers forwards compatible with version 3 via 5fc58f5.

I think with this set of changes, and the last set of changes in #60, this stuff is ready to merge. Could I get somebody from @zendframework/community-review-team to do a once-over for me, however?

akrabat:
I like this a lot. Nicely done and clear.

PR #60, opened 2016-01-05
Forwards-compatibility changes

weierophinney:
This patch provides forwards compatibility features for the v2 series to allow users to update their code for use with v3 prior to an actual migration.

## AbstractPluginManager

This patch modifies the `AbstractPluginManager` constructor to allow the following combination of arguments:

- No arguments (current behavior).
- Single `ConfigInterface` argument (current behavior); used to configure the instance.
- Single `ContainerInterface` argument (v3 behavior); sets the "parent" locator/"creation context"
- `ContainerInterface` as first argument, array as second (v3 behavior); sets the "parent" locator/"creation context", and uses the second argument to configure the instance.

These changes allow developers the ability to continue to use existing plugin managers when upgrading to v3 (though they will also need to implement a `validate()` method, which will be in the migration guide).

The changes presented are backwards compatible (loosens typehint on initial argument; new, second argument is optional), and are based on changes being implemented in #59.

## InvokableFactory

The `InvokableFactory` introduced in the develop branch is compatible with v2 as well, and this patch adapts it for use with v2. This will allow existing component refactors to retain changes that introduce the `InvokableFactory` + aliases.

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -9,6 +9,7 @@
 
 namespace Zend\ServiceManager;
 
+use Interop\Container\ContainerInterface;
 use Exception as BaseException;
 
 /**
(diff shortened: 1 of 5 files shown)


akrabat:
Nicely done. Merged to master only as I'm pretty sure that it's not relevant to develop.

PR #61, opened 2016-01-11
InvokableFactory behavior for v2 series

weierophinney:
This patch provides a couple of fixes for behavior/incompatibilities observed in the v2 series when adapting existing components to work against both v2 and v3:

- First, it adds `InvalidServiceException`. This particular exception exists on v3, but not v2, and is thrown by the v3 implementation of `InvokableFactory`. As such, v2 requires it.
- Second, it alters the behavior of `InvokableFactory::createService()` to accommodate the two possible behaviors that exist in v2 (but not v3):
  - The original behavior, in which a `name => factory` configuration is used, results in `$canonicalName` being a normalized name, and `$requestedName` being the name as passed to `get()` — which would be the class name to use.
  - In v2, if an alias was used, `$canonicalName` is the *resolved alias*, which will be the *qualified class name*. As such, this patch introduces a check to see if `$canonicalName` is an existing class name, and, if so, it uses that name when calling `__invoke()`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ matrix:
     - php: 7
     - php: hhvm 
   allow_failures:
-    - php: 7
     - php: hhvm
 
 notifications:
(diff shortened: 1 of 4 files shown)


weierophinney:
Note this goes against the newly created release-2.7 branch.

PR #62, opened 2016-01-11
Fix benchmarks assets

snapshotpl:


Ocramius:
Already included in #64

PR #64, opened 2016-01-12
Performance optimization

snapshotpl:
I have deeply reviewed code of v3 and I have found some interesting performance improvements.

* we don't need to resolve name from aliases before search in existing services.
* we don't need to resolve aliases any time - only when they are change
* and my favourite: improve construction of ServiceManager. We don't need to validate overrides if we configure manager first time, so we can omit it. It improve creation time about 20-25%!


snapshotpl:
Strange. I don't know my master's tests pass, because composer missing `Zend\Stdlib`

composer.json
```diff
@@ -15,7 +15,8 @@
     },
     "require": {
         "php": "^5.5 || ^7.0",
-        "container-interop/container-interop": "~1.0"
+        "container-interop/container-interop": "~1.0",
+        "zendframework/zend-stdlib": "^2.4"
```
Maks3w: If it used for make the test run then should be added to require-dev

snapshotpl: It's used here https://github.com/zendframework/zend-servicemanager/blob/master/src/Config.php#L62

Ocramius: That class looks like a good candidate for optimization and for dropping a
dependency ;-)

snapshotpl: :+1:

snapshotpl: Fixed in #68

snapshotpl:
#68 should fix tests

src/ServiceManager.php
```diff
@@ -164,7 +171,6 @@ public function getServiceLocator()
     public function get($name)
     {
         $requestedName = $name;
-        $name = isset($this->resolvedAliases[$name]) ? $this->resolvedAliases[$name] : $name;
```
bakura10: I'm not sure this will work. If you have an alias set, asking the alias the first time, then the resolved name the second time will create the instance two times instead of one.

snapshotpl: @bakura10 this line is just move after next if statement. We don't need `$name` before this check

bakura10: You're right, sorry! :)

snapshotpl:
I have rebased to current stable master and all tests pass!

bakura10:
Looks good to me now :). Thanks for the improvement!

src/ServiceManager.php
```diff
@@ -172,6 +178,8 @@ public function get($name)
             return $this->services[$requestedName];
         }
 
+        $name = isset($this->resolvedAliases[$name]) ? $this->resolvedAliases[$name] : $name;
```
Ocramius: Dang, wish we had PHP7's `??` for this.

benchmarks/FetchServiceManager.php
```diff
@@ -0,0 +1,39 @@
...
+namespace ZendBench\ServiceManager;
+
+use Athletic\AthleticEvent;
+use Zend\ServiceManager\ServiceManager;
+
+class FetchServiceManager extends AthleticEvent
+{
+    const NUM_SERVICES = 1000;
+
+    protected $config;
```
Ocramius: Missing hint

src/ServiceManager.php
```diff
@@ -575,9 +584,9 @@ private function resolveAlias($alias)
     /**
      * Resolve all aliases to their canonical service names.
      */
-    private function resolveAliases()
+    private function resolveAliases(array $aliases)
     {
-        foreach ($this->aliases as $alias => $service) {
+        foreach ($aliases as $alias => $service) {
             $this->resolvedAliases[$alias] = $this->resolveAlias($alias);
```
Ocramius: Can `resolveAlias` be inlined?

benchmarks/FetchServiceManager.php
```diff
@@ -0,0 +1,39 @@
...
+
+    protected function getConfig()
```
Ocramius: Private

src/ServiceManager.php
bakura10: I just tried, it brings no benefits.

Ocramius: @bakura10 I'd expect the method overhead to be quite noticeable here. This particular patch tries to avoid the initialization of aliases overall, which is why you don't see that at all.

bakura10: Having said that, I think we can inline the method call. resolveAlias is private anyway, it is small and is only used in this context.

Ocramius: I'll send a patch and write a test for it shortly, gimme a few :-)

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

Ocramius:
I just noticed that the entire benchmark setup is completely broken. Not good.

bakura10:
Yep. See that: https://github.com/zendframework/zend-servicemanager/pull/62

Ocramius:
@bakura10 no, it's not that. See https://github.com/zendframework/zend-servicemanager/blob/218eb28069a59c9aadea611c399049f29fce2404/benchmarks/FetchServices.php#L38-L43

How many of those are actually new services? How many were already instantiated?

bakura10:
Haaaa. You're right!! Didn't spot that. Good catch.

Ocramius:
I'm currently migrating to phpbench to have some actual usable benchmark outputs (mainly because the output is too different between runs).

snapshotpl:
You have absolutely right. I had plan to fix that soon

Ocramius:
@snapshotpl hold yer horses: I'm sending a patch against your branch in a few.

Ocramius:
@snapshotpl see snapshotpl/zend-servicemanager#1 - feel free to review/merge, it will be included in this PR afterwards.

snapshotpl:
@Ocramius looks great. Nice performance boost coming in next patch ;)

bakura10:
Great to see we still managed to squeeze a bit of performance (I honestly didn't think it was possible :-). Feel free to come with a 20% performance improvement to Zend Expressive too ^^.

Ocramius:
@bakura10 I see a lot of space for micro-optimizations with great impact in here, although it really needs phpbench/phpbench#252 before we even think of making the code harder to read :-P

Ocramius:
@snapshotpl merged, thanks!

Issue #66, opened 2016-01-13
Cannot create using alias and InvokableFactory


snapshotpl:
I try to get service using alias. Alias is linked to factory with InvokableFactory. I get exception:
```
InvokableFactory requires that the requested name is provided on invocation; please update your tests or consuming container
```

After debug I have found that requested name in InvokableFactory is alias name.

It's occur in 2.7.

weierophinney:
Which version - v2.7.2 or v3?

snapshotpl:
@weierophinney 2.7. I have provided falling test for this case

weierophinney:
That's *really* odd; the changes made were introduced due to testing plugin managers under identical circumstances (e.g. [zend-barcode](https://github.com/zendframework/zend-barcode/pull/16)). However, I've imported your test case locally, and can verify it. Looking into a solution now.

weierophinney:
Oof, figured it out; it's due to the fact that v2 canonicalizes the names when storing the keys for the factories. In this particular case, the `InvokableFactory` is receiving `zendtestservicemanagertestassetinvokableobject` as the `$canonicalName`, which, if it weren't canonicalized, would be correct.

The reason the change worked in zend-barcode is due to having the `$autoAddInvokableClass` flag enabled; all the various plugins in that component are invokable classes, and, as such, when the plugin manager cannot resolve to a workable factory, that logic kicks in.

A patch will be tricky, but it's a priority for our migration, so I'll be working on it today.

snapshotpl:
Thanks, it makes a sense

Ocramius:
[this](https://github.com/rstgroup/zend-servicemanager/commit/ba006d69236dd5e6ef53b1c5bda557b8c89616c9#diff-4b732fbb667feb3b23722ff3ce7dd061R1237) seems wrong:

```php

    public function testCanGetServiceUsingAliasAndBuildByInvokableFactory()
    {
        $this->serviceManager->setAlias('foo', 'ZendTest\ServiceManager\TestAsset\InvokableObject');
        $this->serviceManager->setFactory('ZendTest\ServiceManager\TestAsset\InvokableObject', 'Zend\ServiceManager\Factory\InvokableFactory');
        $result = $this->serviceManager->get('foo');
        // following assertion was corrected: (you were comparing against a string)
        $this->assertInstanceOf('ZendTest\ServiceManager\TestAsset\InvokableObject', $result);
    }
```

weierophinney:
[...]

PR #69, opened 2016-01-13
Alias resolution fixes

weierophinney:
This patch fixes alias resolution for version 2 to ensure (a) forwards compatibility with version 3, and (b) proper resolution of aliases with regards to services resolved by abstract factories.

The original report in #66 was that doing the following did not work correctly:

```php
$sm->setAlias('foo', InvokableObject::class);
$sm->setFactory(InvokableObject::class, InvokableFactory::class);
$result = $sm->get('foo');
```

When the above code was executed, `InvokableFactory::createService()` was receiving arguments that it could not use, and thus raising an exception. On inspection, what I discovered was that the `$cName` provided as a normalized version of the resolved alias — and due to normalization, would not resolve to a fully qualified class name. This was due to the fact that the SM was storing a normalized version of the alias target.

As such, the patch does the following:

- Alias targets are no longer canonicalized in the internal `$aliases` array.  This means that circular detection must canonicalize the targets for purposes of comparison when iterating aliases.

  It also means that when an alias is resolved, the name returned is the name as registered with the alias — which becomes important when considering fully qualified class names.

- `get()` now resets the `$name` to the *resolved alias target* if an alias was detected. Combined with the above change, this means that factories and abstract factories now receive the resolved alias target as the requested name, and its canonicalized version as the `$cName`.

One test broke with this change: `testGetAbstractFactoryWithAlias()`.

Previously, it read as follows:

```php
    public function testGetAbstractFactoryWithAlias()
    {
        $this->serviceManager->addAbstractFactory('ZendTest\ServiceManager\TestAsset\FooAbstractFactory');
        $this->serviceManager->setAlias('foo', 'ZendTest\ServiceManager\TestAsset\FooAbstractFactory');
        $this->assertInstanceOf('ZendTest\ServiceManager\TestAsset\Foo', $this->serviceManager->get('foo'));
    }
```

My assertion is that the test expectation is incorrect. The resolved alias of 'foo' is `ZendTest\ServiceManager\TestAsset\FooAbstractFactory`, *which that same abstract factory cannot resolve!* The behavior when retrieving a service via alias should be identical to fetching the alias target (which is the case with version 3).

As such, this test rewrites the test to demonstrate what I consider the intended expectation:

```php
    public function testGetAbstractFactoryWithAlias()
    {
        $expected = new TestAsset\Foo;
        $abstractFactory = $this->prophesize(AbstractFactoryInterface::class);
        $abstractFactory->canCreateServiceWithName(
            $this->serviceManager,
            'foo',
            'Foo'
        )->willReturn(true);
        $abstractFactory->createServiceWithName(
            $this->serviceManager,
            'foo',
            'Foo'
        )->willReturn($expected);
        $this->serviceManager->addAbstractFactory($abstractFactory->reveal());
        $this->serviceManager->setAlias('bar', 'Foo');
        $this->assertSame($expected, $this->serviceManager->get('bar'));
    }
```

The one consideration at this point is that @Ocramius has indicated that the DoctrineModule may be relying on the previous behavior (where the alias name was passed to the abstract factory, and not the resolved alias); I'm waiting for him to check into this before we consider it for merging.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -418,8 +418,7 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
     {
         $cName = $this->canonicalizeName($name);
 
-        if (
-            !isset($this->invokableClasses[$cName])
+        if (!isset($this->invokableClasses[$cName])
             && !isset($this->factories[$cName])
             && !$this->canCreateFromAbstractFactory($cName, $name)
         ) {
@@ -505,7 +504,8 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
 
         if ($this->hasAlias($cName)) {
             $isAlias = true;
-            $cName = $this->resolveAlias($cName);
+            $name = $this->resolveAlias($cName);
+            $cName = $this->canonicalizeName($name);
         }
 
         $instance = null;
@@ -524,8 +524,7 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
 
         if (!$instance) {
             $this->checkNestedContextStart($cName);
-            if (
-                isset($this->invokableClasses[$cName])
+            if (isset($this->invokableClasses[$cName])
                 || isset($this->factories[$cName])
                 || isset($this->aliases[$cName])
                 || $this->canCreateFromAbstractFactory($cName, $name)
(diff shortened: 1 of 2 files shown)


weierophinney:
I've cloned the DoctrineModule repo, added my local copy of this repository as a repository, and updated the servicemanager dependency to refer to my branch as 2.7.3, and then run the tests; they all pass.

@Ocramius if you can sign off on this, I'll merge and cut the 2.7.3 release.

Ocramius:
"works on my computer" seal of approval!

:+1:

PR #77, opened 2016-01-21
service can be a Closure

dgoosens:
`ServiceManager::setService($name, $service)` should allow the usage of Closures... Should it not?

In Zend-Expressive, adding a service with a closure results in a Fatal Error.

```php
// /config/autoload/services.global.php

use Zend\Authentication\AuthenticationService;
use Zend\Authentication\Storage\Session;
use Zend\Authentication\Adapter\DbTable;

return [
    'dependencies' => [
        'services' => [
            AuthenticationService::class => function() {
                return new AuthenticationService(new Session(/*...*/), new DbTable(/*...*/));
            }
        ]
    ]
]
```

Or is this dumb?

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -403,6 +403,10 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
             $this->unregisterService($cName);
         }
 
+        if($service instanceof \Closure) {
+            $service = $service();
+        }
+
         $this->instances[$cName] = $service;
 
         return $this;


Ocramius:
If you want the closure to be executed, then your closure is not the service per-se: it is a factory.

dgoosens:
right...  
that makes sense

thanks

pine3ree:
@dgoosens
i apologize for having referenced this PR in my issue, the real target was my PR with the same number 77 in zend-form repo.
kind regards

Issue #78, opened 2016-01-23
getFactory() passes $this to abstract factory canCreate()


kynx:
See:
https://github.com/zendframework/zend-servicemanager/blob/develop/src/ServiceManager.php#L611

... but when the factory is invoked, it get's `$this->creationContext`.

I'm not sure if this is intentional or not, but it causes infinite loops when the abstract factory tries to do a `$services->get('foo')` inside it's `canCreate()`. See the fail at: https://github.com/zendframework/zend-inputfilter/pull/86

PR #79, opened 2016-01-23
Fixed creationContext not being passed to abstract factory canCreate()

kynx:
Fix for #78

### DIFF
--- a/test/AbstractPluginManagerTest.php
+++ b/test/AbstractPluginManagerTest.php
@@ -16,6 +16,7 @@ use Zend\ServiceManager\ConfigInterface;
 use Zend\ServiceManager\Exception\InvalidArgumentException;
 use Zend\ServiceManager\Exception\InvalidServiceException;
 use Zend\ServiceManager\Exception\ServiceNotFoundException;
+use Zend\ServiceManager\Factory\AbstractFactoryInterface;
 use Zend\ServiceManager\Factory\FactoryInterface;
 use Zend\ServiceManager\Factory\InvokableFactory;
 use Zend\ServiceManager\ServiceManager;
@@ -348,5 +349,18 @@ class AbstractPluginManagerTest extends TestCase
         $pluginManager->configure(['services' => [
             stdClass::class => new stdClass(),
         ]]);
+    }
+
+    public function testAbstractFactoryGetsCreationContext()
+    {
+        $serviceManager = new ServiceManager();
+        $pluginManager = new TestAsset\SimplePluginManager($serviceManager);
+        $abstractFactory = $this->prophesize(AbstractFactoryInterface::class);
+        $abstractFactory->canCreate($serviceManager, 'foo')
+            ->willReturn(true);
+        $abstractFactory->__invoke($serviceManager, 'foo', null)
+            ->willReturn(new InvokableObject());
+        $pluginManager->addAbstractFactory($abstractFactory->reveal());
+        $this->assertInstanceOf(InvokableObject::class, $pluginManager->get('foo'));
     }
 }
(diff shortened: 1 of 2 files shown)


kynx:
Ping @weierophinney @ezimuel

Ocramius:
@kynx merged, thanks!

test/AbstractPluginManagerTest.php
```diff
@@ -349,4 +350,17 @@ public function testPassingServiceInstanceViaConfigureShouldRaiseExceptionForInv
             stdClass::class => new stdClass(),
         ]]);
     }
+
+    public function testAbstractFactoryGetsCreationContext()
+    {
+        $serviceManager = new ServiceManager();
+        $pluginManager = new TestAsset\SimplePluginManager($serviceManager);
+        $abstractFactory = $this->prophesize(AbstractFactoryInterface::class);
+        $abstractFactory->canCreate($serviceManager, 'foo')
```
bakura10: shouldBeCalled() should be added (as far as I remember, without shouldBeCalled this assertion does nothing actually, and changing foo to something completely different does not throw any error... but I may be wrong. Mind testing @Ocramius ?

test/AbstractPluginManagerTest.php
```diff
@@ -349,4 +350,17 @@ public function testPassingServiceInstanceViaConfigureShouldRaiseExceptionForInv
...
+            ->willReturn(true);
+        $abstractFactory->__invoke($serviceManager, 'foo', null)
+            ->willReturn(new InvokableObject());
```
bakura10: Same, ->shouldBeCalled() is necessary.

Ocramius: `shouldBeCalled` is not necessary, as the purpose of the test is not to verify that the factory was called, but that the `InvokableObject` was returned (and the factory received those parameters in between).

`shouldBeCalled` is `self::atLeastOnce()`, IIRC.

Ocramius: Also: every time I read prophecy code I cringe >.<

test/AbstractPluginManagerTest.php
kynx: True. But the intention was to catch the situation where `canCreate()` was called with the `$pluginManager` instead of the `$serviceManager`, which it does (in a rather ugly fashion ;)

test/AbstractPluginManagerTest.php
bakura10: Haha @danizord absolutely wanted we used Prophecy and I actually quite like it now :). I've tested, you're right. I don't know what I thought that :/.

danizord: `shouldBeCalled()` is not mandatory here, but it is useful for debugging. If the method does not get called, then I'd prefer the test to fail early telling that it was not called instead instad of failing on `assertInstanceOf()` where I can't see why it failed.

Ocramius: Send another patch.

@danizord stop brain-washing @bakura10 :-P

Dynamic methods are bad, mmkay? :-P

bakura10: If only this young dude would brain-wash me only on this... :p

danizord: >Dynamic methods are bad, mmkay? :-P

\*cough\*, I've learned from GandalPHP that magic is good when it works :)

Ocramius: @danizord that's why I wrote a 2 Mb library to avoid having magic methods ;-)

danizord: I see, but they don't hurt in this context (tests) IMHO. It just makes the test a lot more readable than:

```php
$this->getMock(MyStuff::class, [], [], '', false)->expects($this->once())->method('doSomething')->with($this->equalTo('something'));
```

Ocramius: Yeah, I like (and read) that syntax better, heh.

danizord: Rly? Damn, that was my unique argument. :joy:

PR #81, opened 2016-01-24
Test for v2-v3 PluginManager migration

kynx:
I'm sure I'm doing something really, really dumb, but I'm trying to follow the [migration docs](https://github.com/zendframework/zend-servicemanager/blob/develop/doc/book/migration.md#migration-example) and from what I've read this test should pass. Instead I get:

```
1) ZendTest\ServiceManager\AbstractPluginManagerTest::testV2v3PluginManager
Zend\ServiceManager\Exception\ServiceNotFoundException: An alias "ZendTest\ServiceManager\TestAsset\InvokableObject" was requested but no service could be found.

/Users/matt/www/zend-servicemanager/src/ServiceManager.php:551
/Users/matt/www/zend-servicemanager/src/AbstractPluginManager.php:161
/Users/matt/www/zend-servicemanager/test/AbstractPluginManagerTest.php:342
```

From what I can tell ServiceManager v2 is still looking up factories by their canonicalised name, which would work with:

```php
$serviceManager->setAlias('foo', Foo::class);
$serviceManager->setFactory(Foo::class, Bar::class)
```

...but not when defined in the `alias` and `factories` properties.

Please tell me I'm going mad.

### DIFF
--- a/test/AbstractPluginManagerTest.php
+++ b/test/AbstractPluginManagerTest.php
@@ -17,7 +17,9 @@ use Zend\ServiceManager\Exception\InvalidArgumentException;
 use Zend\ServiceManager\Exception\RuntimeException;
 use Zend\ServiceManager\ServiceManager;
 use ZendTest\ServiceManager\TestAsset\FooPluginManager;
+use ZendTest\ServiceManager\TestAsset\InvokableObject;
 use ZendTest\ServiceManager\TestAsset\MockSelfReturningDelegatorFactory;
+use ZendTest\ServiceManager\TestAsset\V2v3PluginManager;
 
 class AbstractPluginManagerTest extends \PHPUnit_Framework_TestCase
 {
@@ -332,5 +334,11 @@ class AbstractPluginManagerTest extends \PHPUnit_Framework_TestCase
     {
         $this->setExpectedException(InvalidArgumentException::class);
         new FooPluginManager($arg);
+    }
+
+    public function testV2v3PluginManager()
+    {
+        $pluginManager = new V2v3PluginManager(new ServiceManager());
+        $this->assertInstanceOf(InvokableObject::class, $pluginManager->get('foo'));
     }
 }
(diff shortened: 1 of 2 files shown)


Ocramius:
> Please tell me I'm going mad.

"I'm going mad."

Jokes apart, seems indeed to be the case :-\ https://travis-ci.org/zendframework/zend-servicemanager/jobs/104510819#L316

The entire aliases system in 2.x was quite of a mess. Is this issue just about the portability components?

kynx:
Yes - just trying to follow the docs.

Well, I can see two solutions - migrators alias the 'namespaceclassname' as well as the full name, or SM looks up factories twice. As someone doing migrations I know which I'd prefer :wink:

svycka:
@kynx, `Foo::class` this never worked when defined in the `alias` and `factories` properties.
related: https://github.com/zendframework/zend-servicemanager/issues/50
but unfortunately some zf2 components indeed trying to use this and it partially works because of auto invocables `$sm->has(Foo::class)` does not work, but` $sm->get(Foo::class)` works, because of `autoInvocable`.

kynx:
OK, think I've found the magic sauce. In v2.7 and v3, the plugin manager is never configured. If I override the constructor to do that, it works:

```php
    public function __construct($configInstanceOrParentLocator = null, array $config = [])
    {
        parent::__construct($configInstanceOrParentLocator, $config);
        $config = new Config(['aliases' => $this->aliases, 'factories' => $this->factories]);
        $config->configureServiceManager($this);
    }
```

In v3, it would be something like:

```php
    public function __construct($configInstanceOrParentLocator = null, array $config = [])
    {
        parent::__construct($configInstanceOrParentLocator, $config);
        $this->configure(['aliases' => $this->aliases, 'factories' => $this->factories]);
    }
```

Should that be happening by default in `AbstractPluginManager`?

kynx:
Scratch the above - I've added the canonicalized entry for the factory to the test and it passes. Yay!

Now if only it worked as well in v3. See #87

PR #82, opened 2016-01-24
Fix phpdoc in the migration of initializers

mnapoli:
Hope I didn't get it wrong.


Ocramius:
@mnapoli merged, thanks!

PR #84, opened 2016-01-25
Fixed release date

svycka:
someone is still living in 2015 :D


Ocramius:
@svycka dang, I will get it wrong for at least another month >.<

Ocramius:
@svycka thanks!

PR #86, opened 2016-01-25
FIX stdlib dependency

bfolliot:


bakura10:
We removed the dependency to stdlib : https://github.com/zendframework/zend-servicemanager/pull/68

bfolliot:
Sorry, i'm late...

PR #87, opened 2016-01-25
Resolve aliases defined as properties + migration docs + test

kynx:


kynx:
This passes on 2.7, but fails here. Stepping through it looks like `has()` checks the resolved aliases, not the ailiases. When aliases are defined as properies these have not been set.

kynx:
I think this is the last of my plugin manager muddles cleared up :)

kynx:
@Ocramius @weierophinney it would be helpful if this got merged.

kynx:
I've added a trait to make it easier for migrators to test their plugin managers and expanded the docs to cover `$sharedByDefault`, `validate()` exceptions and testing.

kynx:
I'm going to split this into two separate PRs: one for the bug fix, one for the documentation changes.

PR #90, opened 2016-01-26
Fix code in doc/book which produce fatal errors

tasmaniski:
Without third argument  (array $options = null)  the code produce error
```
Fatal error: __invoke() must be compatible with Zend\ServiceManager\Initializer\InitializerInterface
```


weierophinney:
Cherry-picked to master for release with 3.0.3. In the future, please issue hotfixes/documentation fixes against the master branch; the develop branch is intended for new *features* specifically.

Thanks for the patch!

PR #91, opened 2016-01-27
Added setCreationOptions() to InvokableFactory

kynx:
I've run into a problem where multiple calls to `get('foo', $options)` on a plugin manager are not setting the options on subsequent calls. I'm an using an alias to `InvokableFactory`, as recommended in the v2-v3 [migration docs](https://github.com/zendframework/zend-servicemanager/blob/master/doc/book/migration.md#migration-example).

This PR addresses that for v2 by adding the `MutableCreationOptionsInterface` to `InvokableFactory`. 

Note that this _only_ works when the plugin manager has been marked `$shareByDefault = false`. I am unable to determine if that is expected behaviour.

### DIFF
--- a/src/Factory/InvokableFactory.php
+++ b/src/Factory/InvokableFactory.php
@@ -12,6 +12,7 @@ namespace Zend\ServiceManager\Factory;
 use Interop\Container\ContainerInterface;
 use Zend\ServiceManager\Exception\InvalidServiceException;
 use Zend\ServiceManager\FactoryInterface;
+use Zend\ServiceManager\MutableCreationOptionsInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
 
 /**
@@ -26,7 +27,7 @@ use Zend\ServiceManager\ServiceLocatorInterface;
  * service manager, and can also be used in v2 code for forwards compatibility
  * with v3.
  */
-final class InvokableFactory implements FactoryInterface
+final class InvokableFactory implements FactoryInterface, MutableCreationOptionsInterface
 {
     /**
      * Options to pass to the constructor (when used in v2), if any.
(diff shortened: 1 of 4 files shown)


kynx:
Sigh. This might work for `InvokableFactory` where we can have different versions in v2 and v3, but not for other packages: `MutableCreationOptionsInterface` has been removed in v3.

Any suggestions?

bakura10:
Nooooo not this one ;(. Don't bring back this one.

Any way, you need to call build instead of get to create new instances at each call.

kynx:
@bakura10  Unfortunately SM2 doesn't have a build method. I'm trying to make this forward compatibility thing work.

kynx:
Right, for forward compatibility I've changed AbstractServiceManager to duck-type the old `MutableCreationOptionsInterface`. If found it sets the `creationOptions` on the factory. The migration process is now:

* If your factory implements `MutableCreationOptionsInterface`, remove the interface from the class but leave the `setCreationOptions()` method. This will ensure the factory works in both v2 and v3.
* Post migration remove `setCreationOptions()` and `createService()` methods

Not pretty, but the best I can think of.

bakura10:
/me want to create a v4 without compatibility :o.

kynx:
Y'know after struggling with this for the last few days I couldn't agree more :)

weierophinney:
> Note that this only works when the plugin manager has been marked $shareByDefault = false. I am unable to determine if that is expected behaviour.

It is. If `$shareByDefault` is enabled, then only the options that are passed for the first retrieval will be honored.

> /me want to create a v4 without compatibility

Migration *is* important. But the nice part is that we only have to think about the current major version to the next, so it should only come up rarely.

src/Factory/InvokableFactory.php
```diff
@@ -115,4 +103,24 @@ public function createService(ServiceLocatorInterface $serviceLocator, $canonica
...
+            $creationOptions = iterator_to_array($creationOptions);
+        }
+
+        if (! is_array($creationOptions)) {
+            throw new InvalidServiceException(sprintf(
+                '%s cannot use non-array, non-traversable creation options; received %s',
+                __CLASS__,
+                (is_object($creationOptions) ? get_class($creationOptions) : gettype($creationOptions))
+            ));
+        }
```
weierophinney: The above are unnecessary, as `$creationOptions` is type-hinted to `array` in the signature. Since that's what's defined in the interface, we can just remove the functionality.

I'll take care of it on merge.

weierophinney:
@bakura10 Also, this is not bringing back the `MutableCreationOptionsInterface`; this PR is against the release-2.7 branch. ;-)

PR #92, opened 2016-01-28
[v3] configure() now resolves aliases defined as properties

kynx:
This supersedes #87 and fixes the problem with aliases defined in the `$aliases` property not being resolved.


kynx:
@weierophinney @Ocramius once you've recovered from the Expressive release party (congrats, Btw ;) it'd be good to get this in. Not many v2-v3 PRs involving plugin managers are going to work without it.

Same goes for #91, but I'm not 100% happy with that one.

weierophinney:
Cherry-picked to master for release with 3.0.3.

@kynx — in the future, bugfixes can go against master; develop is for *new features*.

Thanks for the great patches!

PR #93, opened 2016-01-28
[v3] Updated migration documentation, added trait for testing migratrations

kynx:
This supersedes #87 fleshing out the migration documentation to include:

- examples of adding normalised FQCN entries for factories
- clear up the changes required for `$shareByDefault` / `$sharedByDefault`
- provide example of handling plugin manager validation exceptions

It includes a trait that can be used when testing migrations to ensure aliases resolve, etc.

Note that tests will fail until #92 is merged

### DIFF
--- a/test/CommonPluginManagerTrait.php
+++ b/test/CommonPluginManagerTrait.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace ZendTest\ServiceManager;
+
+use ReflectionClass;
+use ReflectionProperty;
+use Zend\ServiceManager\Exception\InvalidServiceException;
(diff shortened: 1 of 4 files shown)


test/CommonPluginManagerTrait.php
```diff
@@ -0,0 +1,110 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace ZendTest\ServiceManager;
```
weierophinney: This should be in the `src/Test/` directory, and have the namespace `Zend\ServiceManager\Test`. Why? Because Composer typically tries to install distribution builds, and our `.gitattributes` file sets up distribution builds to omit the `test/` directory. As such, the trait will not be present for most users unless then install with the `--prefer-source` flag.

Move the trait, and update the docs, and I can merge. :+1:

kynx:
Thanks - I've moved the test trait as requested. I'll do a separate PR against release-2.7 so any tests using it run in both.

weierophinney:
Cherry-picked to master for release with 3.0.3.

PR #98, opened 2016-02-24
fixed coveralls config

svycka:
fixed coveralls config and restricted satooshi/php-coveralls to ^1.0 stable version

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,3 +1,2 @@
 coverage_clover: clover.xml
 json_path: coveralls-upload.json
-src_dir: src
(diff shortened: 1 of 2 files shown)


samsonasik:
:+1: yes, the `src_dir` config is not needed in latest php-coveralls ^1.0, another components needs to be updated as well

samsonasik:
@svycka would you mind to provide PR to other components as well ?

svycka:
@samsonasik maybe, but just after this is merged because I don't know if there is everything ok.

samsonasik:
I used same approach for my personal repos, and that's how it works, see travis log

Warm regards,

Abdul Malik Ikhsan

Pada 31 Mar 2016, pukul 12.18, Vytautas Stankus <notifications@github.com> menulis:

svycka:
@samsonasik I know it works but not sure zf team will approve it. Maybe they want to specify source folder or does not want `-v` option or something I don't know so I am waiting until this is merged.

Ocramius:
@svycka thanks, included into 3.1.0 via zendframework/zend-servicemanager@8129d6ebe7cef495a24e40b52c79cdede66b3a84

PR #103, opened 2016-03-13
Allow proxymanager v2 installation

Ocramius:
This patch allows installation of ProxyManager v2.*

The changes are not breaking, but ProxyManager v2 has PHP 7 type-hints everywhere, which means that PHPUnit 5.* is required in order to mock them (specifically, PHPUnit 5.2.9+). This patch adds the required changes, as well as the polyfills required to have ProxyManager v2 behave like ProxyManager v1, configuration-wise

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -14,7 +14,9 @@ use Interop\Container\ContainerInterface;
 use Interop\Container\Exception\ContainerException;
 use ProxyManager\Configuration as ProxyConfiguration;
 use ProxyManager\Factory\LazyLoadingValueHolderFactory;
+use ProxyManager\FileLocator\FileLocator;
 use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
+use ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy;
 use Zend\ServiceManager\Exception\ContainerModificationsNotAllowedException;
 use Zend\ServiceManager\Exception\CyclicAliasException;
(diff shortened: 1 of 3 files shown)


src/ServiceManager.php
```diff
@@ -750,6 +752,10 @@ private function createLazyServiceDelegatorFactory()
 
         if (! isset($this->lazyServices['write_proxy_files']) || ! $this->lazyServices['write_proxy_files']) {
             $factoryConfig->setGeneratorStrategy(new EvaluatingGeneratorStrategy());
```
Ocramius: This block becomes superfluous when ProxyManager v1 support is dropped

src/ServiceManager.php
```diff
@@ -750,6 +752,10 @@ private function createLazyServiceDelegatorFactory()
...
+        } else {
+            $factoryConfig->setGeneratorStrategy(new FileWriterGeneratorStrategy(
```
Ocramius: Note: no tests introduced for this change: this is already covered by the current test suite

sandrokeil:
@weierophinney Any reasons why this PR is not already merged. The lazy_services setting `write_proxy_files` has no effect with ProxyManager 2.0.

weierophinney:
@ocramius Patch looks good to me; feel free to merge and tag a release.

Ocramius:
Included into 3.1.0 via zendframework/zend-servicemanager@d27e1092a7ed20f6894257168be33009175ef928

Issue #104, opened 2016-03-15
Repeated AbstractFactory in ServiceManager may occur.


Moln:
I use `zend-mvc`. When module A & B may be have the same AbstractFactory in `module.config.php`. Can remove duplicate AbstractFactory in ServiceManager ?

Exapmle:
```php
use Zend\ServiceManager\ServiceManager;
use Zend\ServiceManager\Factory\AbstractFactoryInterface;
use Interop\Container\ContainerInterface;

class TestAbstractFactory implements AbstractFactoryInterface
{
    public static $calls = 1;
    public function canCreate(ContainerInterface $container, $requestedName)
    {
        echo "Call: " . self::$calls . "\n";
        self::$calls++;
        return false;
    }
    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
    {
    }
}

$services = new ServiceManager();
$services->addAbstractFactory('TestAbstractFactory');
$services->addAbstractFactory('TestAbstractFactory');
$services->addAbstractFactory('TestAbstractFactory');

$services->has('test');

//Print:
//Call: 1
//Call: 2
//Call: 3

//Expected: 
//Call: 1
```

PR #106, opened 2016-03-31
Repeated AbstractFactory may occur in ServiceManager.

Moln:
See https://github.com/zendframework/zend-servicemanager/issues/104

### DIFF
--- a/test/CommonServiceLocatorBehaviorsTrait.php
+++ b/test/CommonServiceLocatorBehaviorsTrait.php
@@ -21,7 +21,7 @@ use Zend\ServiceManager\Factory\FactoryInterface;
 use Zend\ServiceManager\Factory\InvokableFactory;
 use Zend\ServiceManager\Initializer\InitializerInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
-use ZendTest\ServiceManager\TestAsset\CallOnlyOnceAbstractFactory;
+use ZendTest\ServiceManager\TestAsset\CallTimesAbstractFactory;
 use ZendTest\ServiceManager\TestAsset\FailingAbstractFactory;
 use ZendTest\ServiceManager\TestAsset\FailingFactory;
 use ZendTest\ServiceManager\TestAsset\InvokableObject;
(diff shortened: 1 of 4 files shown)


Ocramius:
Needs a test case before being applied.

Moln:
Added a unit test.

samsonasik:
ping @Ocramius

Ocramius:
I just realized that this change is incorrect.

Following code is perfectly legit, for example:

```php
$abstractFactory1 = new MyConfiguredAbstractFactory($config1);
$abstractFactory2 = new MyConfiguredAbstractFactory($config2);

$sm = new ServiceManager([
    'abstract_factories' => [
        $abstractFactory1,
        $abstractFactory2,
    ],
]);
```

This scenario seems legit to me, and is actually even used within stuff like DoctrineModule, now that I look closer.

I think the key used to store the factory internally should be based on `spl_object_hash()` instead.

A test for this scenario is also needed, IMO...

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -139,6 +140,22 @@ public function testCanCreateServiceWithAbstractFactory()
...
+        $serviceManager = $this->createContainer([
+            'abstract_factories' => [
+                new CallOnlyOnceAbstractFactory(),
+                new CallOnlyOnceAbstractFactory(),
+            ]
+        ]);
+        $serviceManager->addAbstractFactory(CallOnlyOnceAbstractFactory::class);
+        $serviceManager->has(stdClass::class);
+
+        $this->assertEquals(1, CallOnlyOnceAbstractFactory::getCallTimes());
```
Ocramius: This should be `2`.

Ocramius: A new test with the same instance of `CallOnlyOnceAbstractFactory` should be provided. In that scenario, only one factory should remain.

Ocramius: And even there, strong reasoning should be given as to why a duplicate abstract factory is actually a problem...

Moln:
Ocramius is right.
It is my new pull.

ServiceManager:
1. The key used `spl_object_hash()` instead.
2. Add `$cachedAbstractFactories` property for cached some instances.

Unit test:
1. Remove `testCallOnlyOnceWithMultipleIdenticalAbstractFactory`
2. Add `testCallTwiceWithDifferentInstanceAbstractFactories` and `testCallOnlyOnceWithSameInstanceAbstractFactory`.

Ocramius:
Ok, now that the patch is correct (and sorry if it feels like I'm dragging you around, I really get to check this stuff when it pops up in my inbox): what kind of problem does #104 cause? Is it just a performance issue?

Assuming idempotent factories (big assumptions, so I might be wrong here), I don't see a problem with having two factories referencing the same instance inside the `ServiceManager`...

Moln:
I thought for a while. May be I am too sensitive to performance.
Yes, it a performance issue. May be very little performance impact.

At that time I wrote two ZF modules (Module A & Module B). They all have the same configuration.

```php
'service_manager' => array(
 'abstract_factories' => array(
    'Zend\Db\Adapter\AdapterAbstractServiceFactory',
    'Zend\Cache\Service\StorageCacheAbstractServiceFactory',
    '...',
 ),
),
```

I just want to get another service, but call every `AbstractFactory::canCreateServiceWithName()` twice. I just feel no need.

Now this is another solution.

```php
class Module
{
    public function init(ModuleManager $manager)
    {
        $manager->getEventManager()->attach(ModuleEvent::EVENT_MERGE_CONFIG, function (ModuleEvent $event) {
            /** @var \Zend\ModuleManager\Listener\ConfigListener $configListener */
            $configListener =$event->getParam('configListener');
            $configs = $configListener->getMergedConfig(false);

            $configs['service_manager']['abstract_factories'] =
                array_unique($configs['service_manager']['abstract_factories']);

            $configListener->setMergedConfig($configs);
        });
    }
}
```

:smile: :smile:

snapshotpl:
It can be problem when `canCreate` will be more expensive and will return false.

Moln:
Or may be it `zend-modulemanager` issue, not `zend- servicemanager`.

weierophinney:
Assigning to 3.2.0 release, as the feature is technically new behavior.

weierophinney:
Merged to develop for release with 3.2.0

PR #111, opened 2016-04-01
Fixes #107 AbstractPluginManager plugin validation of empty serverice

### LINKED ISSUE #107: ViewHelperPlugin manager configure problem
ViewHelperManager is special type of ServiceManager. There is problem with flushing instances with configure (SM 3.0). It is needed for still alive applications e.g on ReactPHP
```php
['HelperName' => null] + $this->services
```
does not work, because there is plugin validation.

visitek:


Ocramius:
This patch cannot be applied without a test case validating it.

src/AbstractPluginManager.php
```diff
@@ -150,6 +150,9 @@ public function get($name, array $options = null)
      */
     public function validate($instance)
     {
+        if ($instance === null) {
+            return;
+        }
```
weierophinney: I'm going to argue that this is incorrect.

If we do not have an instance, this means no plugin was identified, which *should* be an invalid situation.

If this feature is required by you, you should implement the `validate()` method to allow it in your own plugin manager, or in an extension of the plugin manager where you need the feature.

weierophinney:
Closing due to lack of activity, but, more importantly, due to problems with the feature.

Issue #112, opened 2016-04-01
Slow aliases and factory set methods


malinink:
In new SM3 there is couple of methods that are more slowly than in SM2

Difference in time is great (about 8x slower on [this](https://github.com/zendframework/zend-form/blob/master/src/View/HelperConfig.php#L287) method)
All other comparisons I provide exactly on that function.  

The reason is that is it a lot of factories and aliases that pass to SM with functions ```setAlias``` and ```setFactory```. Behind the scene one key value pair meld with config in ```configure``` method.

I see 2 ways to solve problem:

- add functions such as ```setAliases``` and etc...

```
public function setAliases($aliasTargetPairs)
{
    $this->configure(['aliases' => $aliasTargetPairs]);
}
```

- Modify function ```setAlias``` to something like that: **it was a bad idea, skip that**

```
public function setAlias($alias, $target)
{
    $this->validateOverrides(['aliases' => [$alias => $target]]);
    $this->aliases[$alias] = $target;
}
```

First one is more elegant and have maximum speed(10x faster then in SM2, 75x faster current SM3). But we need to rewrite other components.
Second one is twice slower than SM2, but there is no need to modify other components. 

Of course we could apply both of them.
I think that first one is best idea, I would like to provide PR for this issue, if it is resonable.

Also I could provide PR to to 2.7 branch of SM witch allows modify other components to use that methods.

malinink:
After some checks I found that:
 - main problem is in calling resolveAliases in configure method at bad condition
 - resolveAliases could be optimized for case when new aliases do not intersect with aliases that already present in SM.

I'll provide PR that consists both of changes.
Also I add one more test for recursive aliases and benchmark for set methods.

All changes that provide my PR improve set methods speeding and does not affect the speed of the other SM methods.

setFactory is now 9x faster
setAlias (for not intersecting case) is 3x faster

malinink:
Also it is still a good question is it a good idea to add
```
public function setAliases($aliasTargetPairs)
{
    $this->configure(['aliases' => $aliasTargetPairs]);
}
```
and ```setFactories``` methods.

Now only two classes will have improvements of that:
 - [first](https://github.com/zendframework/zend-form/blob/master/src/View/HelperConfig.php#L287)
 - [second](https://github.com/zendframework/zend-i18n/blob/master/src/View/HelperConfig.php#L75)

As a result we will have much faster load of application.

snapshotpl:
Great improvement!

[...]

PR #114, opened 2016-04-06
Add setAliases and setFactories methods

malinink:
Supposed to use that functions here:
 - [zend-form](https://github.com/zendframework/zend-form/blob/master/src/View/HelperConfig.php#L287)
 - [zend-i18n](https://github.com/zendframework/zend-i18n/blob/master/src/View/HelperConfig.php#L75)
 - [zend-mvc](https://github.com/zendframework/zend-mvc/blob/master/src/Service/ViewHelperManagerFactory.php#L99)

More information could be found here #112

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -420,6 +420,16 @@ class ServiceManager implements ServiceLocatorInterface
     }
 
     /**
+     * Specify a factories for a given service names.
+     *
+     * @param array $nameFactoryPairs
+     */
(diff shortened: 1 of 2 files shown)


Ocramius:
Same remarks as #115 - we will need to make this API private or at least marked as internal

Ocramius:
Closing as per discussion

Moving on to:

 * zendframework/zend-form#77
 * zendframework/zend-i18n#47

PR #115, opened 2016-04-06
Add setAliases and setFactories methods branch 2.7

malinink:
We should add methods to  branch 2.7 in order to change other components that to use them before we migrate to SM3.

More information could be found here #112

All components listed in symmetric PR #114

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -303,6 +303,22 @@ class ServiceManager implements ServiceLocatorInterface, ContainerInterface
     }
 
     /**
+     * Set factories
+     *
+     * @param  array $nameFactoryPairs
+     * @return ServiceManager
+     * @throws Exception\InvalidArgumentException
+     * @throws Exception\InvalidServiceNameException
+     */
+    public function setFactories($nameFactoryPairs)
+    {
+        foreach ($nameFactoryPairs as $name => $factory) {
+            $this->setFactory($name, $factory);
+        }
+        return $this;
+    }
+
+    /**
      * Add abstract factory
      *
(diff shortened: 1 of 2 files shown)


Ocramius:
Disagree with this approach. This API should at least be private and used only by internal `zendframework/zend-servicemanager` components. `public` is a no-go in my opinion, especially after all the effort put into reducing the surface of interaction with the ServiceManager

malinink:
@Ocramius 
If that methods will be protected or private then we could not use them outside class.
Both of methods made for the same as a function setFactory and setAlias (and that method are public).

The reason of creation new methods is new SM3, where it is better to add all aliases or factories at once due to performance issues.

Maybe there are any other ways to do that? (all discussion is here #112 and here #122)

Ocramius:
> The reason of creation new methods is new SM3, where it is better to add all aliases or factories at once due to performance issues.

I understand that. I wonder if we have a way to avoid that, and instead merge configs upfront (which would remove an entire set of problems in this space completely)

weierophinney:
Use configure() is you want to do multiple aliases or factories at once.
It's not BC, bit works for v3.

malinink:
@Ocramius @weierophinney 
As I see, these PR already exist in all depend components. 
 - https://github.com/zendframework/zend-form/pull/77
 - https://github.com/zendframework/zend-i18n/pull/47

So it is better to close both my PR (#114 #115), and issue (#112).

Ocramius:
Closing as per discussion

Moving on to:

 * zendframework/zend-form#77
 * zendframework/zend-i18n#47

Issue #126, opened 2016-06-14
2.7 InvokableFactory options from AbstractPluginManager affect multiple unshared instantiations


AndrolGenhald:
I ran into this when using a NotEmpty validator, if I specify the options for one of them, all of them get those options rather than the defaults.

```
public function getInputFilterSpecification()
{
	return [
		'elementA' => [
			'required' => true,
			'validators' => [
				[
					'name' => 'notEmpty',
					'options' => [
						'type' => NotEmpty::NULL,
					],
				],
			],
		],
		'elementB' => [
			'required' => true,
			//Injected validator will have type NotEmpty::NULL
		],
	];
}
```

AndrolGenhald:
I added a test and tried to fix it by having AbstractPluginManager call setCreationOptions with an empty array, but that causes testMutableMethodNeverCalledWithoutCreationOptions to fail.
AndrolGenhald/zend-servicemanager@c56f494d0948697ac2867d0b2aa4eb877a395c37

weierophinney:
Fixed with #127.

PR #127, opened 2016-06-15
Bugfix/invokable factory options affect multiple instantiations

AndrolGenhald:
Fix issue #126 
I don't really like checking instanceof InvokableFactory, but at least it works.

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -11,6 +11,7 @@ namespace Zend\ServiceManager;
 
 use Interop\Container\ContainerInterface;
 use Exception as BaseException;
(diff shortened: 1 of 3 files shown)


AndrolGenhald:
Any word on this? It's still an issue.

weierophinney:
Thanks, @AndrolGenhald\! Released with 2.7.7

thestanislav:
It seem this breaks default initialization of Zend\Filter\Boolean
First parameter of constructor must be null, not an empty array for default initialization 
https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L62

adamlundrigan:
@weierophinney ping (BC break)

PR #136, opened 2016-07-12
InitializerInterface already in use

kokspflanze:
i found on my fail pull request:

https://travis-ci.org/bjyoungblood/BjyAuthorize/jobs/144009025

Cannot use Zend\ServiceManager\Initializer\InitializerInterface as InitializerInterface because the name is already in use in `ServiceManager.php` on line 27

im not realy sure about the alias name to resolve problem, so im open for other solutions.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -24,7 +24,7 @@ use Zend\ServiceManager\Exception\ServiceNotCreatedException;
 use Zend\ServiceManager\Exception\ServiceNotFoundException;
 use Zend\ServiceManager\Factory\AbstractFactoryInterface;
 use Zend\ServiceManager\Factory\DelegatorFactoryInterface;
-use Zend\ServiceManager\Initializer\InitializerInterface;
+use Zend\ServiceManager\Initializer\InitializerInterface as InitializerInitializerInterface;
 
 /**
  * Service Manager.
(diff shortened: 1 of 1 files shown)


src/ServiceManager.php
```diff
@@ -24,7 +24,7 @@
 use Zend\ServiceManager\Exception\ServiceNotFoundException;
 use Zend\ServiceManager\Factory\AbstractFactoryInterface;
 use Zend\ServiceManager\Factory\DelegatorFactoryInterface;
-use Zend\ServiceManager\Initializer\InitializerInterface;
+use Zend\ServiceManager\Initializer\InitializerInterface as InitializerInitializerInterface;
```
svycka: maybe better:
```php
use Zend\ServiceManager\Initializer;
//and then 
Initializer\InitializerInterface::class
```

svycka:
```php
use Zend\ServiceManager\Initializer;
```
this is not required not sure we need this. vote :D

weierophinney:
I just ran into this when working on zf-apigility as well. What's really odd is that the component itself raises no errors, and I've tested literally dozens of components and modules against zend-servicemanager v3 versions in the past months with no issue. I'm wondering it's a reflection issue?

There are actually three interfaces affected:

- `InitializerInterface`
- `AbstractFactoryInterface` (this was what I ran into)
- `DelegatorFactoryInterface`

I'm going to see if I can find a way around it first before looking at merging this.

PR #137, opened 2016-07-15
Fix: Remove unused imports

localheinz:
This PR

* [x] removes unused imports

💁 Ran

```
$ composer global require friendsofphp/php-cs-fixer:2.0.0-alpha
$ php-cs-fixer fix --using-cache=no --rules=no_unused_imports .
```


weierophinney:
Thanks, @localheinz

PR #142, opened 2016-08-23
[docs] Don't return $instance in initializer callable doc

akrabat:
`Zend\ServiceManager\Initializer\InitializerInterface::__invoke()`'s docblock says that it does not return the instance, so the migration documentation should not either.

FWiW, the V2 example probably should't return `$instance` either. However, the V2 interface says that it returns `mixed`, even though v2's `ServiceManager` doesn't do anything with it in `doCreate()`. Wasn't sure what to do with the V2 one, so left it alone.


weierophinney:
Thanks, @akrabat\!

Issue #128, opened 2016-06-19
Сonfigurable factories


skar:
Hi, all,

I have tourbles with tons of factories (controllers/actions/simple services/etc) and I think to do simple dependencies configuration like:

```php
'dep-name' => [
    'class' => MyService::class,
    'deps' => [
        Dep1::class,
        Dep2::class,
    ],
],
```
It's will be alternative for `factories`, `invokables` etc and works like a factory sample:

```php
'dep-name' => function($container) {
    return new MyService($container->get(Dep1::class), $container->get(Dep2::class));
},
```

In the future it can be used for annotations/reflection support maybe.

I want to implement it.

But I need help from you!
First, I need discussions and ideas for implementation (config examples, naming etc).
Second, I need support (moral).

Ocramius:
As discussed on the mailing list, this is a good idea, but:

 * we need to enforce string keys for the parameters, as it is impossible to override them otherwise (merged config)
 * we need to type-check all parameters, and throw meaningful exceptions in case of problems
 * we may (further down the line) compile these definitions into factories
 * we need benchmarks for this, even if the first version will not actually be performing very well
 * we may look at symfony/dependency-injection to avoid writing our own incompatible DSL for no good reason

skar:
> we need to enforce string keys for the parameters, as it is impossible to override them otherwise (merged config)

Stdlib\ArrayUtils::merge [can](https://github.com/zendframework/zend-stdlib/blob/master/src/ArrayUtils.php#L277) merge numeric keys without override, but not array_merge.
I think it will be the will of the end user, but it should be described in the documentation.

> we need to type-check all parameters, and throw meaningful exceptions in case of problems

Maybe typehint can help us?

> we need benchmarks for this, even if the first version will not actually be performing very well

Any examples?


> Avoid this bit: has been only pain so far ;-)

Keyword 'maybe' =)

Koopzington:
type-check parameters?
If the constructor of the Service has typehinted it's parameters you'd automatically get meaningful errors when trying to pass instances to it that aren't what the constructors expect.
```php
final class Dep2
{
}

final class MyService
{
    public function __construct(Dep1 $dep1, Dep2 $dep2)
    {
    }
}

$myService = new MyService(new Dep2(), new Dep3());

Fatal error: Uncaught TypeError: Argument 1 passed to MyService::__construct() must be an instance of Dep1, instance of Dep2 given

[...]

PR #143, opened 2016-08-24
Added new ConfigAbstractFactory per issue #128

GeeH:
Configurable abstract factory allows creation of services via config key, see tests for usage.

Feedback greatfully received.

### DIFF
--- a/benchmarks/FetchNewServicesBench.php
+++ b/benchmarks/FetchNewServicesBench.php
@@ -5,6 +5,7 @@ namespace ZendBench\ServiceManager;
 use PhpBench\Benchmark\Metadata\Annotations\Iterations;
 use PhpBench\Benchmark\Metadata\Annotations\Revs;
 use PhpBench\Benchmark\Metadata\Annotations\Warmup;
+use Zend\ServiceManager\AbstractFactory\ConfigAbstractFactory;
 use Zend\ServiceManager\ServiceManager;
 
 /**
@@ -30,6 +31,11 @@ class FetchNewServicesBench
             ],
             'services' => [
                 'service1' => new \stdClass(),
+                'config' => [
+                    ConfigAbstractFactory::class => [
+                        BenchAsset\Bar::class => [],
+                    ],
+                ],g
             ],
             'aliases' => [
                 'factoryAlias1'          => 'factory1',
@@ -37,7 +43,8 @@ class FetchNewServicesBench
                 'recursiveFactoryAlias2' => 'recursiveFactoryAlias1',
             ],
             'abstract_factories' => [
-                BenchAsset\AbstractFactoryFoo::class
+                BenchAsset\AbstractFactoryFoo::class,
+                ConfigAbstractFactory::class,
             ],
         ]);
     }
@@ -144,5 +151,11 @@ class FetchNewServicesBench
         $sm = clone $this->sm;
 
         $sm->build('foo');
+    }
+
+    public function benchFetchConfigAbstractFactoryBar()
+    {
+        $sm = clone $this->sm;
+        $sm->get(BenchAsset\Bar::class);
     }
 }
--- a/benchmarks/FetchNewServicesBench.php
+++ b/benchmarks/FetchNewServicesBench.php
@@ -35,7 +35,7 @@ class FetchNewServicesBench
                     ConfigAbstractFactory::class => [
                         BenchAsset\Bar::class => [],
                     ],
-                ],g
+                ],
             ],
             'aliases' => [
                 'factoryAlias1'          => 'factory1',
--- a/src/AbstractFactory/ConfigAbstractFactory.php
+++ b/src/AbstractFactory/ConfigAbstractFactory.php
@@ -11,15 +11,13 @@ namespace Zend\ServiceManager\AbstractFactory;
 
 use Zend\ServiceManager\Factory\AbstractFactoryInterface;
 
-class ConfigAbstractFactory implements AbstractFactoryInterface
+final class ConfigAbstractFactory implements AbstractFactoryInterface
 {
 
     /**
-     * Can the factory create an instance for the service?
+     * Factory can create the service if there is a key for it in the config
      *
-     * @param  \Interop\Container\ContainerInterface $container
-     * @param  string $requestedName
-     * @return bool
+     * {@inheritdoc}
      */
     public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
     {
@@ -33,16 +31,7 @@ class ConfigAbstractFactory implements AbstractFactoryInterface
     }
 
     /**
-     * Create an object
-     *
-     * @param  \Interop\Container\ContainerInterface $container
-     * @param  string $requestedName
-     * @param  null|array $options
-     * @return object
-     * @throws \Zend\ServiceManager\Exception\ServiceNotFoundException if unable to resolve the service.
-     * @throws \Zend\ServiceManager\Exception\ServiceNotCreatedException if an exception is raised when
-     *     creating a service.
-     * @throws \Interop\Container\Exception\ContainerException if any other error occurs
+     * {@inheritDoc}
      */
     public function __invoke(\Interop\Container\ContainerInterface $container, $requestedName, array $options = null)
     {
--- a/src/AbstractFactory/ConfigAbstractFactory.php
+++ b/src/AbstractFactory/ConfigAbstractFactory.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\ServiceManager\AbstractFactory;
+
+use Zend\ServiceManager\Factory\AbstractFactoryInterface;
+
+class ConfigAbstractFactory implements AbstractFactoryInterface
+{
+
+    /**
+     * Can the factory create an instance for the service?
+     *
+     * @param  \Interop\Container\ContainerInterface $container
+     * @param  string $requestedName
+     * @return bool
+     */
+    public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
+    {
+        if (!$container->has('config') || !array_key_exists(self::class, $container->get('config'))) {
+            return false;
+        }
+        $config = $container->get('config');
+        $dependencies = $config[self::class];
+
+        return array_key_exists($requestedName, $dependencies);
+    }
+
+    /**
+     * Create an object
+     *
+     * @param  \Interop\Container\ContainerInterface $container
+     * @param  string $requestedName
+     * @param  null|array $options
+     * @return object
+     * @throws \Zend\ServiceManager\Exception\ServiceNotFoundException if unable to resolve the service.
+     * @throws \Zend\ServiceManager\Exception\ServiceNotCreatedException if an exception is raised when
+     *     creating a service.
+     * @throws \Interop\Container\Exception\ContainerException if any other error occurs
+     */
+    public function __invoke(\Interop\Container\ContainerInterface $container, $requestedName, array $options = null)
+    {
+        $config = $container->get('config');
+        $dependencies = $config[self::class][$requestedName];
+
+        // class has no dependencies, just create it and return
+        if (empty($dependencies)) {
+            return new $requestedName();
+        }
+
+        $arguments = [];
+        foreach ($dependencies as $dependency) {
+            $arguments[] = $container->get($dependency);
+        }
+
+        return new $requestedName(...$arguments);
+    }
+}
--- a/src/AbstractFactory/ConfigAbstractFactory.php
+++ b/src/AbstractFactory/ConfigAbstractFactory.php
@@ -27,7 +27,9 @@ final class ConfigAbstractFactory implements AbstractFactoryInterface
         $config = $container->get('config');
         $dependencies = $config[self::class];
 
-        return array_key_exists($requestedName, $dependencies);
+        return is_array($dependencies) && array_key_exists($requestedName, $dependencies) && is_array(
+            $dependencies[$requestedName]
+        );
     }
 
     /**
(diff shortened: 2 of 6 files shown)


GeeH:
It's worth noting that this only works on PHP 5.6+ but as it's an optional abstract factory (I don't think we would enable it by default but open to that debate) it should be fine.

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+
+class ConfigAbstractFactoryTest extends \PHPUnit_Framework_TestCase
+{
+
+    public function testCanCreateShortCircuits()
+    {
+        $abstractFactory = new ConfigAbstractFactory();
+        $serviceManager = new ServiceManager();
+
+        self::assertFalse($abstractFactory->canCreate($serviceManager, 'OcramiusSucks'));
```
vaclavvanik: > OcramiusSucks

lol

GeeH:
Benchmarks look "OK", but it is definitely slower:

3d7ab5

I'm trying to figure out why this is so much slower - my limited knowledge thinks it shouldn't be, as it's only array lookups...

GeeH:
The slow point is argument unpacking, as I learned from this benchmark - https://gist.github.com/GeeH/ad03958521c8481be5e94e1a22556182

Personally, I put this in the realms of acceptable, I would take the performance hit for the convenience (I'm also aware I'm talking to myself).

DASPRiD:
The code looks sane to me, and I really like the feature. It will save one from writing a bunch of factories which usually just pull more dependencies from the container. The performance loss is rather minimal and neglectable.

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+        $serviceManager->setService(
+            'config',
+            [
+                ConfigAbstractFactory::class => [
+                    InvokableObject::class => [],
+                    FailingFactory::class => [],
+                    SimpleDependencyObject::class => [
+                        InvokableObject::class,
+                        FailingFactory::class,
+                    ],
```
macnibblet: I think we have a Java problem here

test/AbstractFactory/ConfigAbstractFactoryTest.php
Ocramius: I approve of this.

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Can the factory create an instance for the service?
+     *
+     * @param  \Interop\Container\ContainerInterface $container
+     * @param  string $requestedName
+     * @return bool
+     */
+    public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
+    {
+        if (!$container->has('config') || !array_key_exists(self::class, $container->get('config'))) {
```
Ocramius: Unless `ConfigAbstractFactory` is `final`, use `get_class($this)`

Ocramius: Note that the class can be made `final`, since it implements the entirety of `AbstractFactoryInterface`

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+            return false;
+        }
+        $config = $container->get('config');
+        $dependencies = $config[self::class];
```
Ocramius: `get_class($this)` if it is not `final`

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Zend\ServiceManager\AbstractFactory;
+
+use Zend\ServiceManager\Factory\AbstractFactoryInterface;
+
+class ConfigAbstractFactory implements AbstractFactoryInterface
+{
+
+    /**
+     * Can the factory create an instance for the service?
```
Ocramius: That's `{@inheritDoc}`. Consider adding your own explanation of what this is looking for before the `{@inheritDoc}`

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+            return false;
+        }
+        $config = $container->get('config');
+        $dependencies = $config[self::class];
+
+        return array_key_exists($requestedName, $dependencies);
+    }
+
+    /**
+     * Create an object
```
Ocramius: That's `{@inheritDoc}`. Consider adding your own explanation of what this is looking for before the `{@inheritDoc}`

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+     *     creating a service.
+     * @throws \Interop\Container\Exception\ContainerException if any other error occurs
+     */
+    public function __invoke(\Interop\Container\ContainerInterface $container, $requestedName, array $options = null)
+    {
+        $config = $container->get('config');
+        $dependencies = $config[self::class][$requestedName];
+
+        // class has no dependencies, just create it and return
+        if (empty($dependencies)) {
```
Ocramius: This block is redundant, and I don't think that it has any purpose, unless `$dependencies` can be anything but `array`. The performance improvement is also not noticeable

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
+    {
+        if (!$container->has('config') || !array_key_exists(self::class, $container->get('config'))) {
+            return false;
+        }
+        $config = $container->get('config');
+        $dependencies = $config[self::class];
+
+        return array_key_exists($requestedName, $dependencies);
```
Ocramius: You also want to check if the dependencies list is an `array`

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,64 @@
...
+        if (empty($dependencies)) {
+            return new $requestedName();
+        }
+
+        $arguments = [];
+        foreach ($dependencies as $dependency) {
+            $arguments[] = $container->get($dependency);
+        }
+
+        return new $requestedName(...$arguments);
```
Ocramius: `return new $requestedName(...array_map([$container, 'get'], $dependencies))`

benchmarks/FetchNewServicesBench.php
```diff
@@ -30,14 +31,20 @@ public function __construct()
             ],
             'services' => [
                 'service1' => new \stdClass(),
+                'config' => [
+                    ConfigAbstractFactory::class => [
+                        BenchAsset\Bar::class => [],
```
Ocramius: Please add one scenario with a non-empty set of dependencies here

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+use Zend\ServiceManager\ServiceManager;
+use ZendTest\ServiceManager\TestAsset\ComplexDependencyObject;
+use ZendTest\ServiceManager\TestAsset\FailingFactory;
+use ZendTest\ServiceManager\TestAsset\InvokableObject;
+use ZendTest\ServiceManager\TestAsset\SimpleDependencyObject;
+
+class ConfigAbstractFactoryTest extends \PHPUnit_Framework_TestCase
+{
+
+    public function testCanCreateShortCircuits()
```
Ocramius: You need a scenario for this that short-circuits if the dependencies list is not an array of strings

Ocramius: Which expands to two tests:

 * not an array (null, string, etc)
 * an array not containing only strings

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        $abstractFactory = new ConfigAbstractFactory();
+        $serviceManager = new ServiceManager();
+        $serviceManager->setService(
+            'config',
+            [
+                ConfigAbstractFactory::class => [
+                    InvokableObject::class => [],
+                    FailingFactory::class => [],
+                    SimpleDependencyObject::class => [
```
Ocramius: I wouldn't expect this to work, since neither `InvokableObject` nor `FailingFactory` are configured services

Ocramius: Also, please don't use the factories as services, makes for very confusing naming in the tests. Please look for alternate test assets

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+            LazyLoadingValueHolderFactory::class . '.config',
+            [
+                'Proxies' => 'Suck',
+            ]
+        );
+        $serviceManager->setService(
+            'config',
+            [
+                ConfigAbstractFactory::class => [
+                    InvokableObject::class => [],
```
Ocramius: Same as above: was confused by this, please define them as services via `setService`

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+                    FailingFactory::class => [],
+                    SimpleDependencyObject::class => [
+                        InvokableObject::class,
+                        FailingFactory::class,
+                    ],
+                    LazyLoadingValueHolderFactory::class => [],
```
Ocramius: Please don't reuse these as services - makes for really confusing tests

Ocramius:
The idea is neat, but there is one problem with it (and I do thing that it just needs testing):

Module 1:

```php
$config = ['services' => ['config' => [ConfigAbstractFactory::class => ['foo' => ['bar', 'baz']]]]]
```

Module 2:

```php
$config = ['services' => ['config' => [ConfigAbstractFactory::class => ['foo' => ['taz', 'tar']]]]]
```

With the current config merging semantics, this ends up being:

```php
$config = ['services' => ['config' => [ConfigAbstractFactory::class => ['foo' => ['bar', 'baz', 'taz', 'tar']]]]]
```

Therefore, supporting string keys would be ideal:

Module 1:

```php
$config = ['services' => ['config' => [ConfigAbstractFactory::class => ['foo' => ['bar' => 'bar', 'baz' => 'baz']]]]]
```

Module 2:

```php
$config = ['services' => ['config' => [ConfigAbstractFactory::class => ['foo' => ['bar' => 'taz', 
'bar' => 'tar']]]]]
```

With this config merging semantics, this ends up being (expected):

```php
$config = ['services' => ['config' => [ConfigAbstractFactory::class => ['foo' => ['bar' => 'taz', 
'bar' => 'tar']]]]]
```

GeeH:
@Ocramius I really dislike the idea of multiple ways to instantiate the same class with different parameters, for me, in that situation you would not use this abstract factory but write a "real" factory in the same way as you do currently. I feel this adds a layer of complexity that's not needed. This is the 99% use case solution, anything outside of that you wouldn't use this method.

Ocramius:
> instantiate the same class with different parameters

No, the idea is that Module 2 overrides Module 1, same as previously in ZF2. If I want to overwrite a single parameter, I need parameters to have a key.

GeeH:
Surely config merging would just handle this? If I override `ComplexDependency::class` key in the `ConfigAbstractFactory::class` key then config merging would override it?

Ocramius:
> Surely config merging would just handle this? 

Yes, but your current implementation does not work with string config keys, only with numeric array keys (which config merging does *add*, not *override*)

src/AbstractFactory/ConfigAbstractFactory.php
Ocramius: Class was made `final`

GeeH:
Can you write me a test for that please @Ocramius?

Ocramius:
```php

    public function testInvokeWithStringKeyArguments()
    {
        $abstractFactory = new ConfigAbstractFactory();
        $serviceManager = new ServiceManager();
        $serviceManager->setService(
            'config',
            [
                ConfigAbstractFactory::class => [
                    InvokableObject::class => [],
                    FailingFactory::class => [],
                    SimpleDependencyObject::class => [
                        InvokableObject::class => InvokableObject::class,
                        FailingFactory::class => FailingFactory::class,
                    ],
                ]
            ]
        );
        $serviceManager->addAbstractFactory($abstractFactory);
        self::assertInstanceOf(
            SimpleDependencyObject::class,
            $abstractFactory($serviceManager, SimpleDependencyObject::class)
        );
    }
```

Please remember to use different test assets

GeeH:
This is really horrible having to have an array key makes my eyes bleed and makes me feel sick to my stomach. I don't want to do this.

Ocramius:
That's how config merging works, @GeeH, and it's been around for at least 4 years.

GeeH:
I know how config merging works, I'm just thinking, @Ocramius, that it may be that if you want to overload something you just define a normal factory for it, it keeps this system clean and simple.

Ocramius:
@geeh for you, it's a 2 LOC change. The alternative is throwing an exception when the array is not numerically indexed. As it currently stands, this test causes a fatal to be raised

GeeH:
@Ocramius  It's a 2 line change in the code, but makes the config massively more complex and unweildy.

Ocramius:
@GeeH I'm fine with that decision, but then `canCreate` must fail when non-numeric indexes are given.

You'd need a check like following in `canCreate`:


```php
$dependencies === array_values(array_map('strval', $dependencies));
```

GeeH:
Roger that, thanks for you input @Ocramius!

GeeH:
Closing this PR, I'm going to do some more work and reopen.

weierophinney:
@GeeH Why not just mark it as a WIP? :smile:

GeeH:
@weierophinney TBH the comments all over the place are confusing my tiny mind

weierophinney:
@GeeH reasonable. Ping me when the new PR is ready, or you've re-opened.

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,46 @@
...
+{
+
+    /**
+     * Factory can create the service if there is a key for it in the config
+     *
+     * {@inheritdoc}
+     */
+    public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
+    {
+        if (!$container->has('config') || !array_key_exists(self::class, $container->get('config'))) {
```
snapshotpl: `isset` is faster than `array_key_exists` and it should still works, right?

PR #146, opened 2016-08-30
Adds a new Config Abstract Factory that allows service creation by config

GeeH:
As per issue #128.

This PR fixes everything suggested in PR #143 but opened a new PR for clarity of new comments.

Ping @weierophinney @Ocramius.


GeeH:
Travis failing is fully expected as this is 5.6+ only.

benchmarks/FetchNewServicesBench.php
```diff
@@ -145,4 +152,18 @@ public function benchBuildAbstractFactoryFoo()
...
+        $sm = clone $this->sm;
+
+        $sm->get(BenchAsset\Bar::class);
+    }
+
+    public function benchBuildConfigAbstractFactoryBar()
+    {
+        $sm = clone $this->sm;
+
+        $sm->get(BenchAsset\Bar::class);
```
Ocramius: These two benchmarks are duplicate

Ocramius: I remember asking for a different bench, but can't see the previous comments here

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,77 @@
...
+
+final class ConfigAbstractFactory implements AbstractFactoryInterface
+{
+
+    /**
+     * Factory can create the service if there is a key for it in the config
+     *
+     * {@inheritdoc}
+     */
+    public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
```
Ocramius: Please import these symbols

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        if (!$container->has('config') || !array_key_exists(self::class, $container->get('config'))) {
+            return false;
+        }
+        $config = $container->get('config');
+        $dependencies = $config[self::class];
+
+        // config must be array, and have a key for the requested name that's value is also an array
+        if (!is_array($dependencies)
```
Ocramius: You can use `! (isset($dependencies[$requestedName]) && is_array($dependencies[$requestedName]))`

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,77 @@
...
+            throw new ServiceNotCreatedException('Config must be an array');
+        }
+
+        if (!array_key_exists(self::class, $config)) {
+            throw new ServiceNotCreatedException('Cannot find a `' . self::class . '` key in the config array');
+        }
+
+        $dependencies = $config[self::class];
+
+        if (!is_array($dependencies)
```
Ocramius: Instead of duplicating the logic, you can simply call `$this->canCreate()` internally

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,77 @@
...
+            || !array_key_exists($requestedName, $dependencies)
+            || !is_array($dependencies[$requestedName])
+            || $dependencies[$requestedName] !== array_values(array_map('strval', $dependencies[$requestedName]))
+        ) {
+            throw new ServiceNotCreatedException('Dependencies config must exist and be an array of strings');
```
Ocramius: This error message could need a revamp of the types of the various values (indexed by their key)

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,77 @@
...
+        }
+        $dependencies = $dependencies[$requestedName];
```
Ocramius: Don't overwrite a variable if the semantic meaning is different. Simply inline it with the next call instead.

test/TestAsset/ComplexDependencyObject.php
```diff
@@ -0,0 +1,19 @@
...
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace ZendTest\ServiceManager\TestAsset;
+
+class ComplexDependencyObject
```
Ocramius: Thanks for adding these new assets that I requested :-)

src/AbstractFactory/ConfigAbstractFactory.php
GeeH: No, these are prescribed in the interface that way.

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,77 @@
...
+        $arguments = array_map([$container, 'get'], $dependencies);
+
+        return new $requestedName(...$arguments);
```
Ocramius: This call will fail if `$arguments` is not a packed array (indexed starting from 0, only adjacent integers).

To check for that, you can use `array_keys($dependencies) === array_values(array_keys($dependencies))`

src/AbstractFactory/ConfigAbstractFactory.php
GeeH: By design, would rather give specific error messages.

src/AbstractFactory/ConfigAbstractFactory.php
GeeH: I don't understand what you are saying here.

src/AbstractFactory/ConfigAbstractFactory.php
Ocramius: Yes, but importing them is still supported (in the `use` statement)

src/AbstractFactory/ConfigAbstractFactory.php
Ocramius: Looks good then

src/AbstractFactory/ConfigAbstractFactory.php
Ocramius: If you have an error message like `'Dependencies config must exist and be an array of strings'`, you will just run in circles trying to find what the broken value is. If you add something like `json_encode(array_map('gettype', $dependencies[$requestedName]))` to the exception, you make it much easier to understand what has gone wrong.

src/AbstractFactory/ConfigAbstractFactory.php
GeeH: Simplified the canCreate and left the logic in __invoke to ensure these error messages fire. Thanks.

GeeH:
Added documentation, I'm actually happy with this now @weierophinney.

froschdesign:
For documentation with code example: 👍 👍 👍

weierophinney:
> Travis failing is fully expected as this is 5.6+ only.

Since this is for the next minor release, feel free to bump the minimum supported PHP version to 5.6. :smile:

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -0,0 +1,73 @@
...
+{
+
+    /**
+     * Factory can create the service if there is a key for it in the config
+     *
+     * {@inheritdoc}
+     */
+    public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
+    {
+        if (!$container->has('config') || !array_key_exists(self::class, $container->get('config'))) {
```
snapshotpl: Why not using `isset` instead of `array_key_exists` ?

weierophinney: @snapshotpl This is more declarative, and the performance impact is negligible in modern PHP versions.

doc/book/config-abstract-factory.md
```diff
@@ -0,0 +1,106 @@
+# Config Abstract Factory
+
+You can simplify the process of creating factories by adding the 
+`ConfigAbstractFactory` to your service manager. This allows you to define
+services using a configuration map, rather than having to create separate 
+factories for all your services. 
+
+## Enabling
```
weierophinney: Please add an additional note indicating that, as an abstract factory, it can also be mapped as a factory:

```php
return [
    'service_manager' => [
        'factories' => [
            SomeCustomClass::class => ConfigAbstractFactory::class,
        ],
    ],
];
```

This approach also has a slight performance improvement, as the SM will not need to loop through abstract factories when discovering the service for the first time.

doc/book/config-abstract-factory.md
```diff
@@ -0,0 +1,106 @@
...
+$serviceManager->addAbstractFactory(new ConfigAbstractFactory());
+```
+
+Or within any config provider using:
+
+```php
+return [
+    'service_manager' => [
+        'abstract_factories' => [
+            ConfigAbstractFactories::class,
```
weierophinney: `s/ies/y/`.

doc/book/config-abstract-factory.md
```diff
@@ -0,0 +1,106 @@
...
+        Logger::class => [
+            Handler::class,
+        ],
+    ],
+];
+```
+
+The definition tells the service manager how this abstract factory should manage dependencies in
+the classes defined. In the above example, `MySimpleClass` has a single dependency on a `Logger`
+instance. The abstract factory will simply look to fulfil that dependency by calling a `get`
```
weierophinney: Please remove weasel words like "simply" and "just".

weierophinney:
Looks good with the exception of a few documentation changes I've requested. Ping me when done, and I'll merge!

Thanks, @GeeH!

doc/book/config-abstract-factory.md
GeeH: This is not an instance of using a "simply" weasel word, as the user isn't "simply" doing something, but the abstract factory is doing something in a technically "simple" way. These words are weaselly only when you are implying that it's easy for the user to do something, or that the user "just" or "only" has to do something. Using these words to explain what technical behaviour happens is OK (in my opinion of course).

It's easy to edit this if you disagree.

GeeH:
Ping @weierophinney.

doc/book/config-abstract-factory.md
weierophinney: "Simply" has become a trigger word for me, but you're absolutely correct about this specific usage.

One other, unrelated, point: `s/fulfil/fulfill/`. :smile:

doc/book/config-abstract-factory.md
```diff
@@ -0,0 +1,120 @@
...
+        'abstract_factories' => [
+            ConfigAbstractFactory::class,
+        ],
+    ],
+];
+```
+
+It is also possible to use the config abstract factory in the traditional way; by registering
+it as a factory for a specific class. This marginally improves performance but loses
+the benefit of not needing to create a factory key for all of you configured factories:
```
weierophinney: But also makes it explicit what code is responsible for creating the instance.

Essentially, we have this:

Approach | Pros | Cons
-------- | ---- | ----
Abstract factory | One-time setup | Performance; discovery of code responsible for creating instance
Factory | Performance; explicit mapping to factory responsible | Additional (duplicate) setup

I think denoting both approaches, along with their benefits and drawbacks, allows developers to choose which approach makes the most sense for their own application needs.

weierophinney:
Okay, one last comment, and then we can wrap up: It might be a good idea to demonstrate in the documentation a migration path from this abstract factory to an explicit factory. One of the frequent comments I've seen from folks about the deprecation of `ServiceLocatorAware` is that developers are unaware of how to write factories. Including some explicit steps here will help them when they have the question of, "I need to do something that falls outside the scope of this abstract factory, but I don't know how."

(And yes, I'm aware you're planning on writing a tool for auto-creation of factories from this configuration and/or to create the configuration, but having docs now is never a bad thing.)

doc/book/config-abstract-factory.md
GeeH: I just think it confuses the message here. I'd much rather see this added into the Abstract Factory documentation and linked to here than adding this to this particular abstract factory. It's a really good point, but hiding that point in this documentation is not the best for me.

GeeH:
I really don't want to look like I'm ducking doing anything more because both points you have made are valid, but I think these points outline an overarching problem with the documentation in general. For me, the component documentation should be dry instructions with examples on how to use this component either stand-alone or in the context of other components such as the MVC. But of course, we also need more "tutorial" style documentation that says "if you outgrow this, you can go here next" or "this is the pros and cons of using abstract factories". I'm 100% onboard that this extra information is needed, but I'm not convinced that this document is the right place for them.

@froschdesign @weierophinney I would love to hear your thoughts on this.

weierophinney:
@GeeH I'll go ahead and merge, and then see if I can figure out good locations for that information. In some of the other components, we have "cookbook" sections, and adding such a section might be the path we should take here. If so, it is beyond the scope of the feature, as you note, so we can do it separately.

GeeH:
I'll make sure there are issues created to capture this.

froschdesign:
@GeeH and @weierophinney 
The "tutorial style documentation" is the "cookbook"! So the documentation is the right and only place.

doc/book/config-abstract-factory.md
weierophinney: Fair enough; let's relegate this to #148. Do you want me to remove when I merge, or do you want to push another commit?

weierophinney:
Awesome feature, @GeeH\! Merged to develop for the 3.2.0 release.

Issue #148, opened 2016-09-01
Add more tutorial documentation for AbstractFactory vs Factory


GeeH:
Essentially, we have this:

| Approach | Pros | Cons |
| --- | --- | --- |
| Abstract factory | One-time setup | Performance; discovery of code responsible for creating instance |
| Factory | Performance; explicit mapping to factory responsible | Additional (duplicate) setup |

I think denoting both approaches, along with their benefits and drawbacks, allows developers to choose which approach makes the most sense for their own application needs.

weierophinney:
See #150 :smile:

PR #150, opened 2016-09-01
Wrote document resolving #148

weierophinney:
This patch adds a cookbook to the documentation, and provides a document outlining the pros and cons of abstract factories vs factories (and/or using the former as the latter).


weierophinney:
Marked as WIP, as I would like to revise it once #153 and #154 are merged.

PR #154, opened 2016-09-14
Feature/add cli tool for config

GeeH:
Creates a set of new command line tools that allow developers to generate their factories or configuration automatically.


src/Tool/FactoryCreator.php
```diff
@@ -0,0 +1,144 @@
...
+                return true;
+            }
+        );
+
+        if (empty($constructorParameters)) {
+            return [];
+        }
+
+        return array_map(function ($parameter) {
+            return $parameter->getType();
```
weierophinney: Discovered a problem here: `ReflectionParameter::getType()` only exists in PHP 7. As such, we need to do one of two things:

- Make this an optional feature, and raise an exception if the class is invoked in a PHP version less than 7.
- Add zend-code as a dependency, and use its reflection features (which use `ReflectionParameter::getType()` on PHP 7, but otherwise fall back to reflecting the docblock). If we take this route, we could make zend-code a *suggested* dependency, for those wanting to use the factory generation tooling.

What are your thoughts, @GeeH ?

src/Tool/FactoryCreator.php
```diff
@@ -0,0 +1,144 @@
...
+            $class,
+            $this->createArgumentString($className)
+        );
+    }
+
+    /**
+     * @param $className
+     * @return string
+     */
+    private function getClassName($className):string
```
weierophinney: I've sent a PR for this, @GeeH — essentially, a scalar return type hint crept in here.

test/TestAsset/SampleFactory.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
```
weierophinney: This class is missing a licence docblock.

test/TestAsset/ObjectWithScalarDependency.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
```
weierophinney: Set the repo link to zend-servicemanager, and set the copyright to just 2016, please.

weierophinney: And do this throughout on any new class files.

test/TestAsset/config/invalid.config.php
```diff
@@ -0,0 +1,2 @@
+<?php
+return null;
```
weierophinney: My bad, but this one needs a file-level docblock.

test/TestAsset/config/test.config.php
```diff
@@ -0,0 +1,2 @@
+<?php
+return [];
```
weierophinney: This file needs a file-level docblock.

test/TestAsset/factories/ComplexDependencyObject.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
```
weierophinney: For those reading through the review, this class file, and the others in this directory, do **not** need file level docblocks, as they are used for comparing the output of the `FactoryCreator`.

src/Tool/FactoryCreator.php
danizord: @weierophinney we can fallback to `ReflectionParameter::isArray()`, `ReflectionParameter::isCallable()` and `ReflectionParameter::getClass()` if PHP < 7.

weierophinney: @danizord I totally missed that `ReflectionParameter::getClass()` will work. That simplifies the concept dramatically. @GeeH — take note! That would make it far easier for us to support both PHP versions, as it would not require an additional dependency.

weierophinney:
@GeeH In case you missed it buried in the replies: we don't need zend-code after all, as `ReflectionParameter::getClass()` will work for us in PHP versions &lt; 7.

src/Tool/FactoryCreator.php
danizord: There's a good example here: https://github.com/phpspec/prophecy/blob/a749c19621ba047a85e73b713e6e10836dd1f5c1/src/Prophecy/Doubler/Generator/ClassMirror.php#L213-L232

weierophinney:
Ironically, we use it in zend-code as well, just before falling back to
docblock annotations. :-)

GeeH:
@weierophinney we still haven't added the cli tool for generating the class => factory mapping in the config, although the tool is there

src/Tool/FactoryCreator.php
```diff
@@ -0,0 +1,144 @@
+<?php
+/**
+ *  Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
```
samsonasik: s/zf2/zend-servicemanager

GeeH: I suggest we leave this in, and then I will submit a different PR to replace all of these annotations across the entire repository.

weierophinney:
Thanks, @GeeH\!

PR #157, opened 2016-09-15
QA tools, licensing, etc.

weierophinney:
- Updates all license docblocks to reference this repo (not the zf2 repo), and the correct year range (when known).
- Adds scripts to composer:
  - cs-check
  - cs-fix
  - test
  - test-coverage
  - upload-coverage
- Updates the Travis integration to use the above scripts.
- Updates the contributing guide to reference these scripts instead of tooling when possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,12 @@ matrix:
   include:
     - php: 5.6
       env:
-        - EXECUTE_TEST_COVERALLS=true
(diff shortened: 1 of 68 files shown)


samsonasik:
@malukenho lib ( malukenho/docheader ) can be used for license header check like proposed at https://github.com/zendframework/zend-expressive/pull/353

PR #158, opened 2016-09-20
Removing class_exists($canonicalName) call from InvokableFactory v2

aliliinskii:
This pull request removes the `class_exists()` call with `$canonicalName` argument from v2 `Zend\ServiceManager\Factory\InvokableFactory::createService()` method (#61).

Since #69 changed the service alias resoultion, InvokableFactory recieves FQCN via $requestedName parameter.
`class_exists($canonicalName)` produces an extra autoloader call which is either redundant (as $canonicalName now contains a canonicalized service name if actual name or alias is given), or may lead to an unexpected behaviour, as `$canonicalName` may be an actual name of another class. In that case it can be pulled by autoloader, making the factory produce an object of that class.

A positive test case for creating an object of $canonicalName class is converted to a negative one, expecting a `Zend\ServiceManager\Exception\InvalidServiceException` expection to be thrown.

### DIFF
--- a/src/Factory/InvokableFactory.php
+++ b/src/Factory/InvokableFactory.php
@@ -65,22 +65,8 @@ final class InvokableFactory implements FactoryInterface
     /**
      * Create an instance of the named service.
      *
-     * First, it checks if `$canonicalName` resolves to a class, and, if so, uses
-     * that value to proxy to `__invoke()`.
-     *
-     * Next, if `$requestedName` is non-empty and resolves to a class, this
-     * method uses that value to proxy to `__invoke()`.
-     *
-     * Finally, if the above each fail, it raises an exception.
-     *
-     * The approach above is performed as version 2 has two distinct behaviors
-     * under which factories are invoked:
-     *
-     * - If an alias was used, $canonicalName is the resolved name, and
-     *   $requestedName is the service name requested, in which case $canonicalName
-     *   is likely the qualified class name;
-     * - Otherwise, $canonicalName is the normalized name, and $requestedName
-     *   is the original service name requested (typically the qualified class name).
+     * If `$requestedName` resolves to a class, this method uses that value
+     * to proxy to `__invoke()`; otherwise, raises an exception.
      *
      * @param ServiceLocatorInterface $serviceLocator
      * @param null|string $canonicalName
@@ -90,11 +76,7 @@ final class InvokableFactory implements FactoryInterface
      */
     public function createService(ServiceLocatorInterface $serviceLocator, $canonicalName = null, $requestedName = null)
     {
-        if (class_exists($canonicalName)) {
-            return $this($serviceLocator, $canonicalName, $this->creationOptions);
-        }
-
-        if (is_string($requestedName) && class_exists($requestedName)) {
+        if (class_exists($requestedName)) {
             return $this($serviceLocator, $requestedName, $this->creationOptions);
         }
 
--- a/src/Factory/InvokableFactory.php
+++ b/src/Factory/InvokableFactory.php
@@ -65,8 +65,13 @@ final class InvokableFactory implements FactoryInterface
     /**
      * Create an instance of the named service.
      *
-     * If `$requestedName` resolves to a class, this method uses that value
-     * to proxy to `__invoke()`; otherwise, raises an exception.
+     * First, it checks if `$requestedName` is non-empty and resolves to a class, and, if so, uses
+     * that value to proxy to `__invoke()`.
+     *
+     * Next, if `$canonicalName` resolves to a class, this method uses that value
+     * to proxy to `__invoke()`.
+     *
+     * Finally, if the above each fail, it raises an exception.
      *
      * @param ServiceLocatorInterface $serviceLocator
      * @param null|string $canonicalName
@@ -76,8 +81,12 @@ final class InvokableFactory implements FactoryInterface
      */
     public function createService(ServiceLocatorInterface $serviceLocator, $canonicalName = null, $requestedName = null)
     {
-        if (class_exists($requestedName)) {
+        if (is_string($requestedName) && class_exists($requestedName)) {
             return $this($serviceLocator, $requestedName, $this->creationOptions);
+        }
+
+        if (class_exists($canonicalName)) {
+            return $this($serviceLocator, $canonicalName, $this->creationOptions);
         }
 
         throw new InvalidServiceException(sprintf(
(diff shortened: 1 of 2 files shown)


Ocramius:
> A positive test case for creating an object of $canonicalName class is converted to a negative one, expecting a `Zend\ServiceManager\Exception\InvalidServiceException` expection to be thrown.

This is a BC break, and it would require a version bump, which is simply 3.0 (where this behavior is not present, AFAIK).

weierophinney:
@aliliinskii Does this solve an observed problem, or is it only an optimization? If the latter, the change in tests is concerning, as it represents a potential (if not real) BC break, as @Ocramius has noted. #69 was fixing a BC break in the 2.7 series, which we don't want to revert.

Also, you state:

> class_exists($canonicalName) produces an extra autoloader call

My understanding of the PHP engine is that once looked up, the result is stored in a hash table, which means no additional autoloader call occurs on subsequent invocations.

aliliinskii:
Yes, this might introduce a BC break in case of an explicit `Zend\ServiceManager\Factory\InvokableFactory::createService()` method usage from userland code. Perhaps a switch of priority in favour of $requestedName would be a better solution?

@weierophinney This does both. It solves the probability of using a wrong class name instead of a requested one as ServiceManager passes the resolved FQCN via $requestedName. Due to the way service name is canonicalized, factory might pick a class from a global namespace instead of a requested one. As a consequence of preferring a requested name it avoids an autoloader miss.

> My understanding of the PHP engine is that once looked up, the result is stored in a hash table, which means no additional autoloader call occurs on subsequent invocations.

That is true, though that's not the case when a class is missing, which will most likely to happen with a canonicalized name. For example, when asking for 2 instances of an unshared `\Foo\Bar` service, the autoloader will be triggered twice to load a class `foobar`.

Ocramius:
If a class is missing on an invokable, that will lead to an exceptional
case that isn't performance-sensitive

aliliinskii:
@Ocramius 

> If a class is missing on an invokable, that will lead to an exceptional case that isn't performance-sensitive

Could you please explain this?

aliliinskii:
Reimplemented the patch by preferring `$requestedName` over `$canonicalName`.
This still fixes the possible inconsistency in `InvokableFactory` and optimizes the autoloader usage for a standard use case. Yet it provides a fallback to `$canonicalName` in case no `$requestedName` is provided or the class is missing. As a result, the original tests were restored.

When using only a composer autoload function, the difference in service instance creation time is about 15-20%; it increases with the length of autoload function stack: for example, 20-25% when appending a proxy manager autoloader. PHP 7.0.11 is used for benchmarks.

<details>
<summary>Benchmark code (click to expand):</summary>
```php
<?php
use Zend\ServiceManager\Factory\InvokableFactory;
use Zend\ServiceManager\Proxy\LazyServiceFactoryFactory;
use Zend\ServiceManager\Config;

$timer = function($title, callable $test) {
    $startTime = microtime(true);
    $test();
    printf($title.": %s\n", microtime(true) - $startTime);
};
$classTemplate = <<<'TEMPLATE'
<?php
namespace Bench;

class Test{index}
{
}
TEMPLATE;
$classDirPath = sys_get_temp_dir().'/bench';
mkdir($classDirPath);
$composerAutoloader = require __DIR__.'/vendor/autoload.php';
$composerAutoloader->addPsr4('Bench\\', $classDirPath);

$serviceManagerConfig = [
    'factories' => [
        LazyServiceFactoryFactory::class => LazyServiceFactoryFactory::class,
        stdClass::class => InvokableFactory::class,
    ],
    'shared' => [],
    'delegators' => [
        stdClass::class => [
            LazyServiceFactoryFactory::class
        ],
    ],
];

for ($i = 0; $i <= 1000; ++$i) {
    file_put_contents($classDirPath."/Test{$i}.php", strtr($classTemplate, ['{index}' => $i]));
    $fqcn = "Bench\Test{$i}";
    $serviceManagerConfig['factories'][$fqcn] = InvokableFactory::class;
    $serviceManagerConfig['shared'][$fqcn] = false;
}

$serviceManager = new ServiceManager(new Config($serviceManagerConfig));
$serviceManager->setService(
    'Config',
    [
        'lazy_services' => [
            'class_map' => [
                stdClass::class => stdClass::class,
            ],
        ],
    ]
);

// adds an extra autoloader function
$serviceManager->get(stdClass::class);

$timer(
    'Creating same class instances',
    function() use ($serviceManager) {
        for ($i = 0; $i < 1000; ++$i) {
            $serviceManager->get('Bench\Test1000');
        }
    }
);
$timer(
    'Creating unique class instances (each class triggers autoloader)',
    function() use ($serviceManager) {
        for ($i = 0; $i < 1000; ++$i) {
            $serviceManager->get("Bench\Test{$i}");
        }
    }
);

array_map('unlink', glob($classDirPath.'/*'));
rmdir($classDirPath);

```
</details>

weierophinney:
@aliliinskii Can you write up the benchmarks using our existing suite, please? (See https://github.com/zendframework/zend-servicemanager/tree/release-2.7/benchmarks)

Otherwise, looking like a much more acceptable change, as no more BC breaks! Thanks!

src/Factory/InvokableFactory.php
```diff
@@ -65,23 +65,14 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
...
+     * to proxy to `__invoke()`.
      *
      * Finally, if the above each fail, it raises an exception.
      *
-     * The approach above is performed as version 2 has two distinct behaviors
-     * under which factories are invoked:
-     *
-     * - If an alias was used, $canonicalName is the resolved name, and
-     *   $requestedName is the service name requested, in which case $canonicalName
-     *   is likely the qualified class name;
```
MKhalusova: fdhdsjr

aliliinskii:
@weierophinney Should i add the benchmark as a part of pull request?

aliliinskii:
Benchmark code using athletic framework:
```php
<?php

namespace ZendBench\ServiceManager;

use Athletic\AthleticEvent;
use Zend\ServiceManager\Factory\InvokableFactory;
use Zend\ServiceManager\Config;

class InvokableFactoryFetchService extends AthleticEvent
{
    /**
     * @var ServiceManager
     */
    protected $sm;

    protected function getConfig()
    {
        return [
            'factories' => [
                BenchAsset\Foo::class => InvokableFactory::class,
            ],
            'shared' => [
                BenchAsset\Foo::class => false,
            ],
        ];
    }

    public function classSetUp()
    {
        $this->sm = new ServiceManager(new Config($this->getConfig()));
    }

    /**
     * Fetch the service using invokable factory
     *
     * @iterations 5000
     */
    public function fetchInvokableFactoryService()
    {
        $result = $this->sm->get(BenchAsset\Foo::class);
    }
}

```

weierophinney:
Thanks, @aliliinskii

PR #160, opened 2016-09-21
Allow the generate-deps-for-config-factory to create the specified config file

weierophinney:
Modifies the tool slightly to allow it to create the specified config file if it does not currently exist.


weierophinney:
Ping @GeeH — minor usability improvement to allow creating a config file, and not require an existing one.

GeeH:
On it.

PR #164, opened 2016-10-04
Fix #159 by resetting creation options to NULL

### LINKED ISSUE #159: Creation options auto passing empty array breaks Boolean Filter
With the change of #127 empty creation options are now passed as empty arrays rather than NULL to the constructors of services.
This results in a breaking change affecting the Zend\Filter\Boolean.

https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L64

Before the PR, NULL would have been passed to the constructor by default, defaulting the "type" value to TYPE_PHP which would do regular PHP-style casts. Now that an empty array is passed instead, it assumes 0 as the type and keeps it there, because there is no 'type' override in the "creation options".

AKA this gets executed:
https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L73
https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L95
https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L104
https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L117

Setting the 'type' to 0, and as such _everything_ is casted to TRUE due to 'casting' being TRUE by default

https://github.com/zendframework/zend-filter/blob/master/src/Boolean.php#L289

So even e.g. NULL gets filtered to TRUE

EDIT: This issue was FIXED in this PR #164 and was released in version 2.7.8

amcsi:
I'm afraid you might consider this a BC break because of changing the signature of `setCreationOptions()`

See #159

### DIFF
--- a/src/Factory/InvokableFactory.php
+++ b/src/Factory/InvokableFactory.php
@@ -107,7 +107,7 @@ final class InvokableFactory implements FactoryInterface
     /**
      * {@inheritdoc}
      */
-    public function setCreationOptions(array $creationOptions)
+    public function setCreationOptions(array $creationOptions = null)
     {
         $this->creationOptions = $creationOptions;
     }
--- a/test/AbstractPluginManagerTest.php
+++ b/test/AbstractPluginManagerTest.php
@@ -151,6 +151,18 @@ class AbstractPluginManagerTest extends \PHPUnit_Framework_TestCase
         $plugin2 = $pluginManager->get(Baz::class);
 
         $this->assertNotEquals($plugin1, $plugin2);
+    }
+
+    public function testInvokableFactoryNoOptionsResetsCreationOptions()
+    {
+        /** @var $pluginManager AbstractPluginManager */
+        $pluginManager = $this->getMockForAbstractClass('Zend\ServiceManager\AbstractPluginManager');
+        $pluginManager->setFactory(Baz::class, InvokableFactory::class);
+        $pluginManager->setShared(Baz::class, false);
+        $creationOptions = ['key1' => 'value1'];
+        $plugin1 = $pluginManager->get(Baz::class, $creationOptions);
+        $plugin2 = $pluginManager->get(Baz::class);
+
         $this->assertSame($creationOptions, $plugin1->getOptions());
         $this->assertNull($plugin2->getOptions());
     }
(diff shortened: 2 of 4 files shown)


test/AbstractPluginManagerTest.php
```diff
@@ -151,6 +151,8 @@ public function testInvokableFactoryOptionsAffectMultipleInstantiations()
         $plugin2 = $pluginManager->get(Baz::class);
 
         $this->assertNotEquals($plugin1, $plugin2);
+        $this->assertSame($creationOptions, $plugin1->getOptions());
+        $this->assertNull($plugin2->getOptions());
```
Ocramius: You need a separate test case for setting creation options once to an array, then back to `null`.

amcsi: Can it be part of this same test method?

Ocramius: Just add a new method with a matching name

amcsi: @Ocramius all right, done

test/TestAsset/Baz.php
```diff
@@ -17,4 +17,9 @@ public function __construct($options = null)
     {
         $this->options = $options;
     }
+
+    public function getOptions()
```
weierophinney: This isn't technically needed, as `$options` is public. As such, the tests could just look at `$instance->options`. I can make that change during merge, if necessary.

src/Factory/InvokableFactory.php
```diff
@@ -107,7 +107,7 @@ public function createService(ServiceLocatorInterface $serviceLocator, $canonica
     /**
      * {@inheritdoc}
      */
-    public function setCreationOptions(array $creationOptions)
+    public function setCreationOptions(array $creationOptions = null)
```
weierophinney: My first inclination on reading this is that it's a BC break and/or incompatible with existing interfaces. However, on more thorough review of the code, I find this is not the case:

- `InvokableFactory` does not implement `MutableCreationOptionsInterface` directly, but instead duck-types it. As such, we can change the signature here without worries of breakage due to inconsistent implementation.
- Further, `InvokableFactory` is marked `final`, which means we do not need to worry about extending classes. This also means that the changes in `AbstractPluginManager` will not have ramifications beyond this class.

As such, I'm going to clear the "BC Break" label, as there is none at this point.

PR #167, opened 2016-12-03
bugfix PDOException return string as code 

kokspflanze:
bugfix PDOException return string as code, also doc say thats getCode return mixed, so we have to caste to int, to avoid FatalErrors

doc http://php.net/manual/en/exception.getcode.php#refsect1-exception.getcode-returnvalues

i will also create the fix for version `2.7`

# Exception

```
PHP Fatal error:  Uncaught Error: Wrong parameters for Zend\ServiceManager\Exception\ServiceNotCreatedException([string $message [, long $code
[, Throwable $previous = NULL]]]) in C:\Tools\pserverCMSFull\vendor\zendframework\zend-servicemanager\src\ServiceManager.php:758
```

# Stack trace:

```
#0 C:\Tools\pserverCMSFull\vendor\zendframework\zend-servicemanager\src\ServiceManager.php(758): Exception->__construct('Service with na...', '
IMSSP', Object(Doctrine\DBAL\Driver\PDOException))
#1 C:\Tools\pserverCMSFull\vendor\zendframework\zend-servicemanager\src\ServiceManager.php(195): Zend\ServiceManager\ServiceManager->doCreate('
doctrine.connec...')
#2 C:\Tools\pserverCMSFull\vendor\doctrine\doctrine-orm-module\src\DoctrineORMModule\Service\EntityManagerFactory.php(38): Zend\ServiceManager\
ServiceManager->get('doctrine.connec...')
#3 C:\Tools\pserverCMSFull\vendor\doctrine\doctrine-orm-module\src\DoctrineORMModule\Service\EntityManagerFactory.php(55): DoctrineORMModule\Se
rvice\EntityManagerFactory->__invoke(Object(Zend\ServiceManager\Service in C:\Tools\pserverCMSFull\vendor\zendframework\zend-servicemanager\src
\ServiceManager.php on line 758
```

# Doctrine\DBAL\Driver\PDOException details

```
array(4) {
  ["get_class"]=>
  string(33) "Doctrine\DBAL\Driver\PDOException"
  ["getMessage"]=>
  string(242) "SQLSTATE[IMSSP]: This extension requires the Microsoft ODBC Driver 11 for SQL Server to communicate with SQL Server. Access the
following URL to download the ODBC Driver 11 for SQL Server for x64: http://go.microsoft.com/fwlink/?LinkId=163712"
  ["getCode"]=>
  string(5) "IMSSP"
  ["instanceof_throwable"]=>
  bool(true)
}
```


dstockto:
Doesn't this imply that whatever error message or alpha code that was sent back with the exception would be cast to 0? There's information in those codes that will be thrown away.

kokspflanze:
@dstockto the change is just for the code, not for the message. yes a code like "IMSSP" will be changed to 0, if the cast not exist it will give following exception, than you have no changs to know what is wrong, expected to look in the code directly on that line.

````
PHP Fatal error:  Uncaught Error: Wrong parameters for Zend\ServiceManager\Exception\ServiceNotCreatedException([string $message [, long $code
[, Throwable $previous = NULL]]]) in C:\Tools\pserverCMSFull\vendor\zendframework\zend-servicemanager\src\ServiceManager.php:758
````

dstockto:
I understand it's not changing the message, but it forces people down the path of error message string matching as opposed to code matching. 

In the example in the first post, the message is "SQLSTATE[IMSSP]: This extension requires the Microsoft ODBC Driver 11 for SQL Server to communicate with SQL Server. Access the
following URL to download the ODBC Driver 11 for SQL Server for x64: http://go.microsoft.com/fwlink/?LinkId=163712"

While it's certainly possible to continue to look for 'IMSSP' in this instance, not all errors returned this way duplicate the string code in the error message. In those cases, the programmer is relying on the error string not changing. Conversely, it's much less likely that the code would change for the same error.

kokspflanze:
@dstockto did you have the problem with a `PDOException` or a other `Exception` with a custome error-code? i understand your problem but a normal `Exception` should have a integer as error-code. http://de2.php.net/manual/en/class.exception.php just the `PDOException` do something special, see https://bugs.php.net/bug.php?id=41064.

you can add also the code before the reason in https://github.com/kokspflanze/zend-servicemanager/blob/63b0109c7fc45b7e0fd047226f44d9b1e85c2e52/src/ServiceManager.php#L755 in the zf error message

Ocramius:
Fwiw, this was widely discussed in internals, and pdo_sqlsrv is the bad
apple, not respecting API signature. That code should be re-thrown as
integer to prevent further issues.

dstockto:
I think there is more than just the pdo_sqlsrv driver that will do this. In postgres, custom errors come out as P0001, etc and I seem to recall various H* for MySQL as well. Adding the code before the message seems like a decent compromise that doesn't throw away potentially useful information.

The reason I even ended up over here is going back through a lot of code I have that catches InvalidQueryExceptions which I believe are wrapping PDOException is that they are doing == checks against the error codes. The error codes tend to be something like 23505 or 23503 or whatever, but it turns out that they are actually strings. I was going to cast and do a === comparison but then ran into some code that has to deal with a P0001 error which means a cast would wipe out that information. So for now I'm doing `(int)$e->getCode() === 23505` if we're talking about a numeric error code and not casting for alphanumeric codes.

If pdo_* needs to re-throw as integer codes then it means documentation of this conversion table becomes critical. As it stands now, the codes match the error codes from the database so googling them will point you to the meaning of the error. If numeric codes are unchanged and alphanumeric codes are rethrown as numbers then with the vendor codes being canonical, I'm not sure there's a way to guarantee that a mapped code wouldn't be used later by the vendor. Perhaps PDOException needs to have a string code spot to accommodate the original code?

Ocramius:
This is still about PDO not respecting the signature. Don't try breaking
the signature everywhere because that's a bad apple.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

PR #169, opened 2016-12-21
Mutation tests

snapshotpl:
I ran humbug and fixed what I can ;)

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -206,14 +206,6 @@ class ServiceManager implements ServiceLocatorInterface
             $this->services[$name] = $object;
         }
 
-        // Also do so for aliases; this allows sharing based on service name used.
-        if ($requestedName !== $name
-            && (($this->sharedByDefault && ! isset($this->shared[$requestedName]))
-                || (isset($this->shared[$requestedName]) && $this->shared[$requestedName]))
-        ) {
-            $this->services[$requestedName] = $object;
-        }
-
         return $object;
     }
 
--- a/src/AbstractFactory/ConfigAbstractFactory.php
+++ b/src/AbstractFactory/ConfigAbstractFactory.php
@@ -7,6 +7,7 @@
 
 namespace Zend\ServiceManager\AbstractFactory;
 
+use Interop\Container\ContainerInterface;
 use Zend\ServiceManager\Exception\ServiceNotCreatedException;
 use Zend\ServiceManager\Factory\AbstractFactoryInterface;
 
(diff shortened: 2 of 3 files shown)


src/ServiceManager.php
```diff
@@ -206,14 +206,6 @@ public function get($name)
             $this->services[$name] = $object;
         }
 
-        // Also do so for aliases; this allows sharing based on service name used.
-        if ($requestedName !== $name
```
Ocramius: What condition was unreachable here according to mutation testing?

Ocramius: There are differences in these conditionals. Still, can you check if the code was covered before the PR? If not, can you check if it is possible to cover it at all?

snapshotpl: I analized it and I think this conditions make same thing, even that looks different. How I understand that is works: it caches requested name if alias is used. This make it after creation object, but code below make it after second use.

weierophinney: I think a better approach is to determine what scenario would allow this section of code to be exercised, and write a test (or tests) to cover it. Removing it looks like it would remove potentially expected behaviors.

snapshotpl: @weierophinney if any exists - yes. But now we don't have any in unit tests

weierophinney: @snapshotpl That's my point: we need to **add** unit tests to cover this functionality. Removing it is a potential BC break if the behavior is expected by end users. Based on the code comment, this code is for allowing aliases to share the same instances as their targets. (Just because the functionality is not covered does not mean it's not intentional.)

Ocramius: @weierophinney let's get to a technical agreement here, since opinions will not get us far (especially with mutation testing):

 * we need to ensure 100% coverage on this class
 * we need to ensure 0% escaped mutants on this class

When we hit that, we can basically delete any code. What @snapshotpl said so far (if I interpreted it correctly) is that this particular code path is unreachable or redundant. If we have 100% coverage with tests and then run the mutation testing, and removing the code still makes all tests pass (and removes the mutation), we're fine.

weierophinney: @Ocramius — Considering that the code in question relies on specific state in the instance (specifically the `$shared` and `$services` properties, which would be based on initial configuration), I'm arguing that it's possible that humbug cannot exercise it with any mutants due to never operating against a _state_ that would cause the code to be exercised.

Does that make sense?

Ocramius: > Does that make sense?

No, if there is coverage, humbug will create a mutant and report a mutation if there isn't sufficient branch coverage

Ocramius:
@snapshotpl back to you - sorry if it basically took 1 year and 1 month to even get the discussion rolling :S

weierophinney:
Closing due to lack of activity. If somebody wants to pick this up, please submit a new PR.

Ocramius:
@weierophinney if this merge applied cleanly, it can be merged as-is.

PR #170, opened 2016-12-27
Adds file-based caching to ReflectionBasedAbstractFactory

kirkmadera:
Adds the ability to use file based caching for the ReflectionBasedAbstractFactory. This could be used in Production and would automatically generate factories on demand. This could reduce deployment time when compared to a script that has to generate all classes and would have little impact on performance to do the initial generation as needed.

This does require declaring ReflectionBasedAbstractFactory explicitly in bootstrap in order to be able to pass parameters to it. There is no AbstractFactory factory. Explaining this is a bit difficult, but this documentation does it well (https://zendframework.github.io/zend-servicemanager/reflection-abstract-factory/). 

Note, you would still need a process for comparing timestamps or removing old cache files on deployments, so I don't know if this is really a valid strategy. A better strategy may still be to generate all classes on deployment. I wanted to push this as a potential improvement though anyways.

### DIFF
--- a/src/AbstractFactory/ReflectionBasedAbstractFactory.php
+++ b/src/AbstractFactory/ReflectionBasedAbstractFactory.php
@@ -12,6 +12,7 @@ use ReflectionClass;
 use ReflectionParameter;
 use Zend\ServiceManager\Exception\ServiceNotFoundException;
 use Zend\ServiceManager\Factory\AbstractFactoryInterface;
+use Zend\ServiceManager\Tool\FactoryCreator;
 
 /**
  * Reflection-based factory.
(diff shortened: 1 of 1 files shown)


kirkmadera:
I also noticed that the logic in `Zend\ServiceManager\Tool\FactoryCreator` is different from that in `Zend\ServiceManager\AbstractFactory\ReflectionBasedAbstractFactory` in that it is missing the special case logic around handling array parameters named "config".

Ocramius:
Idea is good, but few things need consideration:

 * Caching should not be added to the original class, but to a new factory that eventually consumes the original one
 * Caching should either use a cache adapter, or abstract the file cache concept away, since it is prone to security and concurrency issues 
 * It is much more efficient to dump a set of service definitions to a local config file than caching on a per-service scenario: consider collecting services and dumping at shutdown
 * This kind of feature has many open-ended questions: it doesn't fit the core service manager, IMO
 * It is fine to have a slow abstract factory that is reflection-based, and then have developers write factories for performance-critical services
 * Tests are absolutely required

Ocramius:
As it currently stands, this feature is a good candidate for a new package that just provides a service dumper: wouldn't implement it here though...

kirkmadera:
Thanks for the thoughtful consideration on this. I'll take your suggestions and ponder on this some more. Closing this PR.

PR #174, opened 2017-01-24
Allows the use of ArrayObject config service with ConfigAbstractFactory

bfolliot:


bfolliot:
With Expressive skeleton, "config" service is an ArrayObject https://github.com/zendframework/zend-expressive-skeleton/blob/master/config/config.php

xtreamwayz:
The ArrayObject is caused by using the zend-expressive-skeleton to setup the application. It is only there for the Aura.Di and is fixed and removed in the [develop branch](https://github.com/zendframework/zend-expressive-skeleton/commit/29f20449b8c660b2b02dbe1c94f915476feb4cfb).

You can easily replace that last line in config/config.php with `return $config;`.

If this causes issues in a zend-expressive-skeleton based application, I think it should be reported to zend-expressive-skeleton. So far no issues have been reported. It would be interesting to know what steps you took during installation in order to get that error.

bfolliot:
Yes, but i think it may be good to have this compatibility with ArrayObject.

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -40,8 +41,8 @@ public function __invoke(\Interop\Container\ContainerInterface $container, $requ
 
         $config = $container->get('config');
 
-        if (! is_array($config)) {
-            throw new ServiceNotCreatedException('Config must be an array');
+        if (! is_array($config) and (! is_object($config) or ! $config instanceof \ArrayObject)) {
```
weierophinney: Change this line to:

```php
if (! (is_array($config) || $config instanceof ArrayObject)) {
```

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -45,6 +46,17 @@ public function testCanCreateReturnsTrueIfDependencyNotArrays()
         $serviceManager->setAllowOverride(true);
         $serviceManager->setService(
             'config',
+            new ArrayObject([
+                ConfigAbstractFactory::class => [
+                    InvokableObject::class => 42,
+                ]
+            ])
+        );
+        self::assertTrue($abstractFactory->canCreate($serviceManager, InvokableObject::class));
```
weierophinney: Remove this. This particular test case is covering the _values_ related to configured classes. You have a more specific test for the `ArrayObject` behavior below which this is duplicating anyways.

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -76,6 +88,23 @@ public function testCanCreate()
         self::assertFalse($abstractFactory->canCreate($serviceManager, ServiceManager::class));
     }
 
+    public function testCanCreateWithArrayObject()
```
weierophinney: Rename this to `testCanCreateReturnsTrueWhenConfigIsAnArrayObject`.

weierophinney: Also, add an additional test `testFactoryCanCreateInstancesWhenConfigIsAnArrayObject`, demonstrating that the factory correctly creates the instance when the `config` service is an `ArrayObject`.

bfolliot:
@weierophinney : It's good for you now?

PR #175, opened 2017-01-25
Fix storing shared status of peering service manager services

derschatta:
When a service is retrieved from the peering service manager the shared entry is stored with a non canonicalized name. This can lead to unwanted behaviour of services not treated as non-shared later.

This fix is only for these who need to use the peering functionality in the 2.x version.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -1037,7 +1037,8 @@ class ServiceManager implements ServiceLocatorInterface
             $peeringServiceManager->serviceManagerCaller = $this;
 
             if ($peeringServiceManager->has($name)) {
-                $this->shared[$name] = $peeringServiceManager->isShared($name);
+                $cname = $this->canonicalizeName($name);
+                $this->shared[$cname] = $peeringServiceManager->isShared($name);
                 $instance = $peeringServiceManager->get($name);
                 $peeringServiceManager->serviceManagerCaller = null;
                 return $instance;
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -463,7 +463,7 @@ class ServiceManager implements ServiceLocatorInterface
      * @return bool|null
      * @throws Exception\ServiceNotFoundException
      */
-    public function getShared($name, $usePeeringServiceManagers = true)
+    private function getShared($name, $usePeeringServiceManagers = true)
     {
         $cName = $this->canonicalizeName($name);
 
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -467,14 +467,6 @@ class ServiceManager implements ServiceLocatorInterface
     {
         $cName = $this->canonicalizeName($name);
 
-        if (!$this->has($name)) {
-            throw new Exception\ServiceNotFoundException(sprintf(
-                '%s: A service by the name "%s" was not found',
-                get_class($this) . '::' . __FUNCTION__,
-                $name
-            ));
-        }
-
         if (!isset($this->shared[$cName]) && $usePeeringServiceManagers) {
             $caller = $this->serviceManagerCaller;
             foreach ($this->peeringServiceManagers as $peeringServiceManager) {
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -456,6 +456,53 @@ class ServiceManager implements ServiceLocatorInterface
     }
 
     /**
+     * returns null if shared value is not set at all otherwise it returns the value
+     *
+     * @param  string $name
+     * @param  bool   $usePeeringServiceManagers
+     * @return bool|null
+     * @throws Exception\ServiceNotFoundException
+     */
+    public function getShared($name, $usePeeringServiceManagers = true)
+    {
+        $cName = $this->canonicalizeName($name);
+
+        if (!$this->has($name)) {
+            throw new Exception\ServiceNotFoundException(sprintf(
+                '%s: A service by the name "%s" was not found',
+                get_class($this) . '::' . __FUNCTION__,
+                $name
+            ));
+        }
+
+        if (!isset($this->shared[$cName]) && $usePeeringServiceManagers) {
+            $caller = $this->serviceManagerCaller;
+            foreach ($this->peeringServiceManagers as $peeringServiceManager) {
+                // ignore peering service manager if they are the same instance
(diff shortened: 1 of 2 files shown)


Ocramius:
This needs a test case

Ocramius:
@derschatta be aware that ServiceManager v3 dropped peering support.

src/ServiceManager.php
```diff
@@ -456,6 +456,53 @@ public function isShared($name)
...
 
     /**
+     * returns null if shared value is not set at all otherwise it returns the value
+     *
+     * @param  string $name
+     * @param  bool   $usePeeringServiceManagers
+     * @return bool|null
+     * @throws Exception\ServiceNotFoundException
+     */
+    public function getShared($name, $usePeeringServiceManagers = true)
```
Ocramius: can be `private` since only other service managers are going to interact with this method

derschatta: It can't be private when it's called from other service manager instance, right?

derschatta: made it private ;)

Ocramius: @derschatta yes, it can ;-)

derschatta: @Ocramius You never stop learning ;).

derschatta: @Ocramius All changes okay for you?

src/ServiceManager.php
```diff
@@ -456,6 +456,53 @@ public function isShared($name)
...
+     * @param  string $name
+     * @param  bool   $usePeeringServiceManagers
+     * @return bool|null
+     * @throws Exception\ServiceNotFoundException
+     */
+    private function getShared($name, $usePeeringServiceManagers = true)
+    {
+        $cName = $this->canonicalizeName($name);
+
+        if (!$this->has($name)) {
```
Ocramius: This check doesn't need to be repeated, since the API is private

src/ServiceManager.php
```diff
@@ -456,6 +456,53 @@ public function isShared($name)
...
+                get_class($this) . '::' . __FUNCTION__,
+                $name
+            ));
+        }
+
+        if (!isset($this->shared[$cName]) && $usePeeringServiceManagers) {
+            $caller = $this->serviceManagerCaller;
+            foreach ($this->peeringServiceManagers as $peeringServiceManager) {
+                // ignore peering service manager if they are the same instance
+                if ($caller === $peeringServiceManager) {
```
Ocramius: Urgh, cyclic peering checking code :S

This needs accurate testing, and the tests provided so far are not sufficient

src/ServiceManager.php
```diff
@@ -456,6 +456,53 @@ public function isShared($name)
...
+                    continue;
+                }
+
+                $peeringServiceManager->serviceManagerCaller = $this;
```
Ocramius: This flag is fishy: what are you doing here?

src/ServiceManager.php
```diff
@@ -456,6 +456,53 @@ public function isShared($name)
...
+
+                $shared = $peeringServiceManager->getShared($cName);
```
Ocramius: Note that you can just access `$peeringServiceManager->shared` here, since you have access to private scope

src/ServiceManager.php
```diff
@@ -456,6 +456,53 @@ public function isShared($name)
...
+
+                $peeringServiceManager->serviceManagerCaller = null;
+            }
+        }
+
+        if (isset($this->shared[$cName])) {
+            return $this->shared[$cName];
+        }
+
+        return;
```
Ocramius: `return null` should be explicit, since the API is not `void`

src/ServiceManager.php
```diff
@@ -1037,7 +1084,12 @@ protected function loopPeeringServiceManagers($name)
             $peeringServiceManager->serviceManagerCaller = $this;
 
             if ($peeringServiceManager->has($name)) {
-                $this->shared[$name] = $peeringServiceManager->isShared($name);
+                $cname = $this->canonicalizeName($name);
+                $shared = $peeringServiceManager->getShared($name);
+                if ($shared === null) {
+                    $shared = $this->shareByDefault();
```
Ocramius: Is the `shareByDefault()` o the other container considered?

src/ServiceManager.php
```diff
@@ -1037,7 +1084,12 @@ protected function loopPeeringServiceManagers($name)
...
+                }
+                $this->shared[$cname] = $shared;
```
Ocramius: Why is this being set locally? I wouldn't do that, IMO :-\

src/ServiceManager.php
derschatta: yeah but the point is to go recursively through the peering managers therefore I need to call the method

src/ServiceManager.php
derschatta: I would if php-cs-fixer wouldn't complain if I write return null :(

src/ServiceManager.php
derschatta: the architecture makes it hard to determine that because if the other peering managers have it set I can't distinguish between a set value or default value. So I decided to use the flag from the uppermost service manager. No idea how to improve that to be honest.

src/ServiceManager.php
derschatta: me neither but it was like that before. Also if it's not shared the instance is stored locally.

src/ServiceManager.php
derschatta: As I understood it from the other methods it provides a way to skip the current instance in the recursive check. Otherwise it can happen that we end up in an endless loop.

derschatta: btw. it's done like that already so I just reuse it in the same way.

src/ServiceManager.php
derschatta: well the tests seemed enough for other methods like ->get or ->has for example and I just use the logic already used in the other methods. It might be ugly but the way the whole peering thing is implemented is ... well, not that perfect maybe ;)

src/ServiceManager.php
Ocramius: That's... weird. This is one of the scenarios where `null` makes more sense. Oh well :-\

src/ServiceManager.php
derschatta: removed it

PR #176, opened 2017-02-01
Generate deps ignoring unresolved dependencies

thomasvargiu:
*Trying to resolve missed cache using `ReflectionBasedAbstractFactory` generating config mappings for a whole module/project.*

### Skipping classes with unresolved dependencies

Added ability to ignore only classed with direct unresolved dependencies in ConfigDumper command.

At this moment, `generate-deps-for-config-factory` command throws an exception if just one dependency in the the dependency tree could not be resolved.

Adding a parameter `-i|--ignore-unresolved` it will not add the definition of the class with an unresolved dependency, but continuing the execution adding other dependency definitions.

Consider these classes:

```php
namespace TestAsset;

class ObjectWithObjectScalarDependency
{
    public function __construct(
        SimpleDependencyObject $simpleDependencyObject,
        ObjectWithScalarDependency $dependency,
        TestInterface $interface
)  {
    }
}

class SimpleDependencyObject
{
    public function __construct(InvokableObject $invokableObject)
    {
    }
}

class ObjectWithScalarDependency
{
    public function __construct($aName)
    {
    }
}

class InvokableObject
{
    public function __construct(array $options = [])
    {
        $this->options = $options;
    }
}

interface TestInterface
{
}

```

Dependency tree:
```
+ ObjectWithObjectScalarDependency
|
+---> SimpleDependencyObject
|     |
|     +---> InvokableObject
|
+---> ObjectWithScalarDependency
|     |
|     +---> $aName (scalar)
|
+---> TestInterface (interface)
```

Command:
```
$ generate-deps-for-config-factory ./config.php TestAsset\ObjectWithObjectScalarDependency
```

This command will write nothing. An exception will throw.

Command:
```
$ generate-deps-for-config-factory --ignore-unresolved ./config.php TestAsset\ObjectWithObjectScalarDependency
```

This command will generate a config with the following mapping config:

```php
$config = [
    ObjectWithObjectScalarDependency::class => [
        SimpleDependencyObject::class,
        ObjectWithScalarDependency::class,
        TestInterface:: class,
    ],
    SimpleDependencyObject::class => [
        InvokableObject::class,
    ],
    InvokableObject::class => [],
];
```

Like for interfaces, classes with unresolved dependencies will be included in other classes dependencies, but no definitions will be written.

This will be useful to generate something similar a cache for a whole project/module.

The idea is to do something like this:

```
$ find ./module/Auth/src/ -name "*.php" | xargs ./bin/classes-from-file | xargs ./bin/generate-deps-for-config-factory ./config.php
```

### ContainerInterface dependecy in ConfigDumper

I added the ability to inject a `ContainerInterface` in the `ConfigDumper` class.

It's used in `ConfigDumper::createDependencyConfig()`.

If a container is injected, before to throw an exception, it checks if the class with an unresolved dependency is already defined in the container. In this case we can safely skip the dependency definition for this class.

### Considerations

I preferred to don't add a command to get classes in files (scanning a directory) to don't include another dependency in this project.

### DIFF
--- a/src/Tool/ConfigDumper.php
+++ b/src/Tool/ConfigDumper.php
@@ -7,6 +7,7 @@
 
 namespace Zend\ServiceManager\Tool;
 
+use Interop\Container\ContainerInterface;
 use ReflectionClass;
 use ReflectionParameter;
 use Traversable;
@@ -26,12 +27,34 @@ return %s;
 EOC;
 
     /**
+     * @var ContainerInterface
+     */
+    protected $container;
+
+    /**
+     * @param ContainerInterface $container
+     */
+    public function __construct(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @param ContainerInterface $container
+     */
+    public function setContainer(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
+
+    /**
      * @param array $config
      * @param string $className
+     * @param bool $ignoreUnresolved
      * @return array
      * @throws InvalidArgumentException for invalid $className
      */
-    public function createDependencyConfig(array $config, $className)
+    public function createDependencyConfig(array $config, $className, $ignoreUnresolved = false)
     {
         $this->validateClassName($className);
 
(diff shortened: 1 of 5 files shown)


src/Tool/ConfigDumper.php
```diff
@@ -26,12 +27,34 @@ class ConfigDumper
 EOC;
 
     /**
+     * @var ContainerInterface
+     */
+    protected $container;
```
weierophinney: `private` visibility, please.

src/Tool/ConfigDumper.php
```diff
@@ -26,12 +27,34 @@ class ConfigDumper
...
+     */
+    public function __construct(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @param ContainerInterface $container
+     */
+    public function setContainer(ContainerInterface $container)
```
weierophinney: Is this setter really necessary? If it _is_, then use it from `__construct()`.

test/TestAsset/ObjectWithObjectScalarDependency.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
```
weierophinney: 2017, not 2016. :smile:

weierophinney:
Thanks, @thomasvargiu\!

PR #180, opened 2017-02-15
Add PSR-11 support to `ServiceManager`

weierophinney:
This patch updates the `container-interop/container-interop` constraint to `^1.2`, and explicitly adds `psr/container` with a constraint of `^1.0`; this immediately allows `Zend\ServiceManager\ServiceManager` to be used as a PSR-11 container.

Moreover, this patch provides a test validating that assertion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,16 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Added
 
-- Nothing.
+- [#180](https://github.com/zendframework/zend-servicemanager/pull/180) adds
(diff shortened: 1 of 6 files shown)


byan:
How we should deal with \Interop\Container\ContainerInterface in implements of the following factories interface like [Factory](https://github.com/zendframework/zend-servicemanager/blob/master/src/Factory/FactoryInterface.php#L37), [InvokableFactory](https://github.com/zendframework/zend-servicemanager/blob/master/src/Factory/InvokableFactory.php#L28), etc?

weierophinney:
@byan Read the proposed psr-11.md file and changelog - I cover the topic in those.

PR #196, opened 2017-07-04
Optimize PHP function calls

smuggli:
Those optimizations target improving OPCODES.

### DIFF
--- a/src/AbstractFactory/ConfigAbstractFactory.php
+++ b/src/AbstractFactory/ConfigAbstractFactory.php
@@ -21,13 +21,13 @@ final class ConfigAbstractFactory implements AbstractFactoryInterface
      */
     public function canCreate(\Interop\Container\ContainerInterface $container, $requestedName)
     {
-        if (! $container->has('config') || ! array_key_exists(self::class, $container->get('config'))) {
+        if (! $container->has('config') || ! \array_key_exists(self::class, $container->get('config'))) {
             return false;
         }
         $config = $container->get('config');
         $dependencies = $config[self::class];
 
-        return is_array($dependencies) && array_key_exists($requestedName, $dependencies);
+        return \is_array($dependencies) && \array_key_exists($requestedName, $dependencies);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


smuggli:
Here are two showcases which demonstrate the optimazations. 

is_string:
https://3v4l.org/sSjcd/vld#output
https://3v4l.org/Ufg8r/vld#output

array_keys:
https://3v4l.org/S8Bse/vld#output
https://3v4l.org/qRcHaJ/vld#output


I ran the PHPBench scripts but I am not sure if the report is necessary for that. 

Also such optimazations already have been made:
https://github.com/zendframework/zend-servicemanager/blob/master/src/ServiceManager.php#L395
https://github.com/zendframework/zend-servicemanager/blob/master/src/ServiceManager.php#L396

snapshotpl:
Run PHPBench scripts to show performance improvement

smuggli:
The reports vary overall a bit, some are better and some worse.
[before.txt](https://github.com/zendframework/zend-servicemanager/files/1122080/before.txt)
[optimized.txt](https://github.com/zendframework/zend-servicemanager/files/1122081/optimized.txt)

Ocramius:
Could you replace those `isset()` checks with `??` overall? Also, variadics may help too in private API

smuggli:
@Ocramius Yes, I can replace them but this implies that support for PHP < 7 would be dropped. Has this been decided yet? For variadics I will see what I can do. Thanks for the comment.

MichaelGooden:
@smuggli The plan moving forward is to start pinning major components against PHP 7.1.

Source: https://framework.zend.com/blog/2017-06-06-zf-php-7-1.html

stefanotorresi:
Honestly, I don't find the performance argument very compelling.
This is a change that would need to be propagated on every component for consistency, and it would also be appropriate to put a coding standards rule in place.
Is it all truly worth it?

Ocramius:
This is generally about internal language optimisations coming up later on:
the engine will optimise more of these, but only if the FQN is used.

smuggli:
@MichaelGooden Thanks completely missed that.

@stefanotorresi It is not a big gain but still something. For now I just started with ServiceManager to see if changes like that are welcome ;). Haven't made any thoughts on a coding standard rule yet.

@Ocramius Just replaced the issets with ?? for now because I am running out of time for today.

webimpress:
@smuggli This PR should target `develop`, not `master`.

smuggli:
@webimpress Changed it to develop. Thanks!

webimpress:
@smuggli I've created separate PR to drop PHP 5.6 and 7.0 support - #197. I will work on PHPCS sniff (in `zend-coding-standard` library) to detect function calls without `\`. It shouldn't be complicated.

Ocramius:
commit smuggli/zend-servicemanager@fc81dee is unrelated, and the patch needs to be rebased to get rid of it

kynx:
Would `use function array_key_exists` achieve the same thing? It would avoid 100s of lines of code changes and be much easier on the eyes ;) Though maybe harder to sniff.

MichaelGooden:
@kynx It does appear to achieve the same outcome:

With `use function` statement: https://3v4l.org/QeVpj/vld#output
Without: https://3v4l.org/sSjcd/vld#output

Ocramius:
@kynx nice one!

kynx:
Cheers @Ocramius ! 

On a side note I just found out PHPStorm can [auto-import from global scope](https://blog.jetbrains.com/phpstorm/2017/04/importing-improvements-in-phpstorm-2017-1/). Nice!

webimpress:
Here is my sniff to import all used internal functions:
https://github.com/zendframework/zend-coding-standard/pull/1/commits/19e353bb65a32774e9fd40d901587f95e777572b

I've created PR #198 after applying the sniff on develop branch.

Should we do the same with constants?
https://3v4l.org/Hlh9d/vld#output
https://3v4l.org/uSN1i/vld#output

Ocramius:
> Should we do the same with constants?

I don't think that we use many internal constants except for:

 * exceptions
 * filesystem-related operations
 * array_map operations

All of the above are usually improved by reducing the operation amount itself, rather than these micro-optimisations...

smuggli:
@Ocramius Sorry had no time working on that. To keep it simple, I would not do any variadics changes yet. Anything else what should be done before merging?

Ocramius:
@smuggli this is good to go as-is, thanks!

PR #197, opened 2017-07-05
Drop PHP 5.6 and 7.0

webimpress:
Dropped PHP 5.6 and 7.0 support.
PR based on #195. It should be merged before #196, because #196 contains changes which requires PHP 7.

Ping @Ocramius


test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -184,8 +184,8 @@ public function testExceptsWhenConfigNotSet()
     {
         $abstractFactory = new ConfigAbstractFactory();
         $serviceManager = new ServiceManager();
-        self::expectException(ServiceNotCreatedException::class);
-        self::expectExceptionMessage('Cannot find a config array in the container');
+        $this->expectException(ServiceNotCreatedException::class);
```
Ocramius: These should stay as `self`

webimpress: Why? Methods `expectException*` are not static. Please see:
https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L569

Ocramius: Oh, totally misread this, sorry :-\

Ocramius:
LGTM, thanks @webimpress!

PR #198, opened 2017-07-09
Import PHP internal functions

webimpress:
As @kynx noticed in #196 we can import internal PHP functions to get the same optimisation in function calls. Changes in this PR are made by sniff from dev of [`zend-coding-standard`](https://github.com/zendframework/zend-coding-standard/pull/1/commits/19e353bb65a32774e9fd40d901587f95e777572b) library and it supersedes #196.

/cc @smuggli @kynx @MichaelGooden @Ocramius @weierophinney


Ocramius:
LGTM :ship:

PR #199, opened 2017-08-08
replace count() usage to empty() when possible

samsonasik:
ref: http://maettig.com/code/php/php-performance-benchmarks.php#empty

### DIFF
--- a/src/Tool/ConfigDumperCommand.php
+++ b/src/Tool/ConfigDumperCommand.php
@@ -124,7 +124,7 @@ EOH;
      */
     private function parseArgs(array $args)
     {
-        if (! count($args)) {
+        if (empty($args)) {
             return $this->createHelpArgument();
         }
 
@@ -140,7 +140,7 @@ EOH;
             $arg1 = array_shift($args);
         }
 
-        if (! count($args)) {
+        if (empty($args)) {
             return $this->createErrorArgument('Missing class name');
         }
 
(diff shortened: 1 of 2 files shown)


src/Tool/ConfigDumperCommand.php
```diff
@@ -124,7 +124,7 @@ public function __invoke(array $args)
      */
     private function parseArgs(array $args)
     {
-        if (! count($args)) {
+        if (empty($args)) {
```
Ocramius: Can be replaced with `if ($args) {`

src/Tool/ConfigDumperCommand.php
```diff
@@ -140,7 +140,7 @@ private function parseArgs(array $args)
             $arg1 = array_shift($args);
         }
 
-        if (! count($args)) {
+        if (empty($args)) {
```
Ocramius: Can be replaced with `if ($args) {`

src/Tool/FactoryCreatorCommand.php
```diff
@@ -104,7 +104,7 @@ public function __invoke(array $args)
      */
     private function parseArgs(array $args)
     {
-        if (! count($args)) {
+        if (empty($args)) {
```
Ocramius: Can be replaced with `if ($args) {`

samsonasik:
@Ocramius I've updated `empty($args)` with `! $args`

Ocramius:
LGTM :+1:

PR #201, opened 2017-08-20
added php72 travis

kokspflanze:
added php72 travis in build matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,18 +47,8 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
-    - php: nightly
-      env:
-        - DEPS=lowest
-    - php: nightly
-      env:
-        - DEPS=locked
-    - php: nightly
-      env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -38,6 +38,15 @@ matrix:
     - php: 7.1
       env:
         - DEPS=latest
+    - php: 7.2
```
webimpress: IMHO, PHP 7.2 should be added instead of "nightly"

xtreamwayz: It has 7.2 AND nightly. Nightly refers to 7.3 at the moment. Not sure if 7.3 is already a travis php tag.

webimpress: Oh, if it refers to PHP 7.3 it's good to have then !

xtreamwayz: https://travis-ci.org/zendframework/zend-servicemanager/jobs/266571573#L267

PR #209, opened 2017-11-28
Updated Travis CI configuration - develop fix

webimpress:
- fixed env variable - `CS_CHECK` instead of `CHECK_CS`
- removed `--no-plugins` flag as it is not needed anymore
- updated actions (to conform maintainers template)


weierophinney:
Thanks, @webimpress

PR #210, opened 2017-12-05
Ensure a `null` value is passed to `setCreationOptions` instead of empty array

weierophinney:
Per https://github.com/zendframework/zend-servicemanager/issues/159#issuecomment-349339218, the updates in #205 broke a previously fixed edge case whereby `AbstractPluginManager::createServiceViaCallback()` would inject a `null` value to a factory's `setCreationOptions()` method if no instance options were provided; it instead passed an empty array, which broke a number of plugins that relied on a null value being present.

This patch does the following:

- Adds a regression test for the behavior.
- Modifies `AbstractPluginManager::createServiceViaCallback()` to have three distinct behaviors surrounding creation options, in the following order:
  - if the factory is an `InvokableFactory`, it tests if the creation options are an array or non-empty; if not, it passes a `null` value to `setCreationOptions()`.
  - if the factory is an `MutableCreationOptionsInterface`, it tests if the creation options are an array or non-empty; if not, it passes an empty array to `setCreationOptions()`.
  - if the factory implements the `setCreationOptions()` and the creation options are a non-array or non-empty, it reflects the method to determine if the options argument has a default value, using that if present, or an empty array if not.

### DIFF
--- a/test/AbstractPluginManagerTest.php
+++ b/test/AbstractPluginManagerTest.php
@@ -209,8 +209,8 @@ class AbstractPluginManagerTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @group 205
+     * @codingStandardsIgnoreStart
      */
-    // @codingStandardsIgnoreStart
     public function testRetrievingServicesViaFactoryThatUsesCreationOptionsShouldReturnNewInstanceEveryTimeOptionsAreProvided()
     {
         // @codingStandardsIgnoreEnd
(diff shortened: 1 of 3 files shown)


amcsi:
How complex :o
Thanks a lot for this quick fix. I can confirm that this fix fixed our integration tests in our projects 👍

amcsi:
One thing I don't quite get... I thought I already wrote a unit test to cover this regression. How come it didn't catch this?
https://github.com/zendframework/zend-servicemanager/pull/164/files#diff-46506c2cbbf939781cbaa62bc7771ec8R156

weierophinney:
@amcsi Because the behavior with an `InvokableFactory` is different than that for a duck-typed factory. Oddly, I had to change the order of conditionals, as `InvokableFactory` fulfills the same duck-type, and I got failures once I started passing a `null` value for the `$options` with a number of our duck-typed factories (which don't have a default `null` value for the `$options`).

Anyways, there's now three issues worth of tests capturing different edge cases, so hopefully it's fixed for the rest of the 2.7 lifetime.

test/AbstractPluginManagerTest.php
```diff
@@ -451,4 +451,16 @@ public function testInvokableFactoryHasMutableOptions()
         $object = $pluginManager->get('foo', $options);
         $this->assertEquals($options, $object->getOptions());
     }
+
+    public function testIfCreationOptionsAreEmptySetThemThemToNullWhenCreatingAService()
+    {
+        // @codingStandardsIgnoreEnd
```
webimpress: @weierophinney Do we need this comment above here? I can't see where is start... ?

weierophinney: Odd; that must be a copy-paste artifact from a test I used as a starting point.

No, not needed, but already released.

PR #211, opened 2017-12-14
Updated to php-coveralls/php-coveralls

webimpress:
With version 2 package has been renamed from `satooshi/php-coveralls` to `php-coveralls/php-coveralls`, and the script has been renamed from `coveralls` to `php-coveralls`


Ocramius:
@webimpress I'm merging this into `develop` only, since it is an improvement.

Ocramius:
Merged via zendframework/zend-servicemanager@c7fd498d0b75f867256007057d31529a4543700e

webimpress:
@Ocramius, this is not an improvement, this is a hotfix. The current version (on master) is not working properly.

Ocramius:
@webimpress that only affects builds, not consumers, so I'd rather have a non-working `master` than backporting something that is only relevant to project development

webimpress:
@Ocramius right, it affects builds, so in case of new "hotfix" to master, we can't see coverage. This is why I think it should be merged to master straight away as a hotfix.

PR #213, opened 2017-12-27
Test and Fix for ServiceManager::get erronously returning shared services

fhein:
This PR provides a test and a fix: Under certain conditions the service manager returns a shared service allthough it should not.

To invoke this bug, set shared_by_default to false. Add a factory for a class. Set shared for this class to true. Define an alias for that class. Get an object using the class id. Get an alias (1) using the alias id. Get another alias (2) using the alias id.

Acceptable result would be: object !== alias1 && object !== alias2 && alias1 !== alias2.

Current result is: object === alias1 === alias2, i.e. the returned service is shared.

Note: With a fresh install from master unit tests do not pass. PHPUnit reports 9 errors. With my test case added, it reports 10 errors (most of them related to \n and \r\n string endings, I guess).

### DIFF
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -286,25 +286,23 @@ class ServiceManagerTest extends TestCase
 
     public function testShouldNotShareAliasesWhichAreNotConfiguredToBeShared()
     {
-        $config =
-        [
-            'factories' =>
+        $sm = new ServiceManager(
             [
-                \stdClass::class => InvokableFactory::class,
-            ],
-            'aliases' =>
-            [
-                'alias' => \stdClass::class,
-            ],
-            'shared_by_default' => false,
-            'shared' =>
-            [
-                \stdClass::class => true,
+                'factories' =>
+                [
+                    \stdClass::class => InvokableFactory::class,
+                ],
+                'aliases' =>
+                [
+                    'alias' => \stdClass::class,
+                ],
+                'shared_by_default' => false,
+                'shared' =>
+                [
+                    \stdClass::class => true,
+                ],
             ]
-        ];
-        $sm = new ServiceManager($config);
-        $alias1 = $sm->get('alias');
-        $alias2 = $sm->get('alias');
-        self::assertNotSame($alias1, $alias2);
+        );
+        self::assertNotSame($sm->get('alias'), $sm->get('alias'));
     }
 }
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -288,9 +288,9 @@ class ServiceManagerTest extends TestCase
     {
         $config =
         [
-            'services' =>
+            'factories' =>
             [
-                \stdClass::class => new \stdClass(),
+                \stdClass::class => InvokableFactory::class,
             ],
             'aliases' =>
             [
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -16,6 +16,7 @@ use Zend\ServiceManager\Factory\InvokableFactory;
 use Zend\ServiceManager\ServiceManager;
 use ZendTest\ServiceManager\TestAsset\InvokableObject;
 use ZendTest\ServiceManager\TestAsset\SimpleServiceManager;
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
 
 /**
  * @covers \Zend\ServiceManager\ServiceManager
@@ -304,5 +305,32 @@ class ServiceManagerTest extends TestCase
             ]
         );
         self::assertNotSame($sm->get('alias'), $sm->get('alias'));
+    }
+
+    public function testCreateNonSharedAliasServiceFromDirectlyDefinedService()
+    {
+        $sm = new ServiceManager(
+            [
+                'services' =>
+                [
+                    \stdClass::class => new \stdClass(),
+                ],
+                'aliases' =>
+                [
+                    'alias' => \stdClass::class,
+                ],
+                'shared_by_default' => false,
+                'shared' =>
+                [
+                    \stdClass::class => true,
+                ],
+            ]
+        );
+        // Obviously the service manager has to have a factory for
+        // each service which is configured not to be shared
+        // As a delicate detail this applies to alias services as
+        // well if the alias is configured not to be shared
+        self::expectException(ServiceNotFoundException::class);
+        $sm->get('alias');
     }
 }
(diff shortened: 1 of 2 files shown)


test/ServiceManagerTest.php
```diff
@@ -283,4 +283,42 @@ public function testFactoryMayBeStaticMethodDescribedByCallableString()
         $serviceManager = new SimpleServiceManager($config);
         $this->assertEquals(stdClass::class, get_class($serviceManager->get(stdClass::class)));
     }
+
+    public function testShouldNotShareAliasesWhichAreNotConfiguredToBeShared()
+    {
+        $config =
+        [
+            'factories' =>
```
Ocramius: This test looks too complex: any way to simplify it a bit more by just removing the assertion message customisation? The assertion errors already match your messages without any retouching needed.

test/ServiceManagerTest.php
```diff
@@ -283,4 +283,42 @@ public function testFactoryMayBeStaticMethodDescribedByCallableString()
         $serviceManager = new SimpleServiceManager($config);
         $this->assertEquals(stdClass::class, get_class($serviceManager->get(stdClass::class)));
     }
+
+    public function testShouldNotShareAliasesWhichAreNotConfiguredToBeShared()
+    {
+        $config =
```
Ocramius: Variable is redundant. Also (even though not specified by CS) why are your values on the line after the key?

test/ServiceManagerTest.php
```diff
@@ -283,4 +283,42 @@ public function testFactoryMayBeStaticMethodDescribedByCallableString()
...
+        [
+            'factories' =>
+            [
+                \stdClass::class => InvokableFactory::class,
+            ],
+            'aliases' =>
```
Ocramius: @weierophinney what do you (currently) think of dropping aliases overall? They're a pain and always will be...

test/ServiceManagerTest.php
```diff
@@ -283,4 +283,42 @@ public function testFactoryMayBeStaticMethodDescribedByCallableString()
...
+            $msg .= "service === alias1 but should be !==\n";
+        }
+        if ($service === $alias2) {
+            $msg .= "service === alias2 but should be !==\n";
+        }
+        if ($alias1 === $alias2) {
+            $msg .= "alias1 === alias2 but should be !==\n";
+        }
+        self::assertNotSame($service, $alias1, $msg);
+        self::assertNotSame($alias1, $alias2, $msg);
```
Ocramius: IMO they should be the same, since the service is marked as shared, and the alias is NOT a new service.

fhein: I like the original intent (as documented in the source) to allow control by name used whether a service aquired as shared or not much.  

One use case example could be requesting an iterator, where 'LocalIterator' creates a new instance (for subiteration, lookahead or whatever), and 'GlobalIterator' refers to the global iteration context. Both can iterate and  share a common environment.

I think, it should be up to the user to decide what makes a new service is. A 'rental car' looks appears to be different  from an 'owned car', allthough both are 'car'.

test/ServiceManagerTest.php
fhein: Of course you're right

test/ServiceManagerTest.php
fhein: The var was not really redundant, I think. I removed it anyway.

test/ServiceManagerTest.php
fhein: @Ocramius @weierophinney:  Why was my (and other's) reply to this question deleted? I explained in some detail, why - at least - I need and honour that particular feature. There are obvious and - from my perspective - urgent use cases. 

There are more platforms around to discuss that, if contributions to discussions here continue to automagically disappear.

Ocramius: Where was this question posed? We don't delete stuff here...

Anyway, replacing aliases with an abstract service factory is sufficient and removes all these quirks, IMO

test/ServiceManagerTest.php
Ocramius: The car is the same one. If you want a new instance you register the service twice, rather than creating multiple names for the same service.

test/ServiceManagerTest.php
Ocramius: It is not reused

test/ServiceManagerTest.php
froschdesign: @fhein 
Nothing is deleted. GitHub hides "outdated" comments after the affected lines of code have been updated.

Please click on the link "Show outdated". (see screenshot)

outdated

test/ServiceManagerTest.php
fhein: Excuse me, what variable? $config? I think I do not understand what you want me to do.

Ocramius: `$config` is only ever used once, so you can squash it into a `new ServiceManager([...])` call :-)

test/ServiceManagerTest.php
fhein: Ok. I did not know that. I apologize.

test/ServiceManagerTest.php
fhein: You're joking,  right? :) Believe it or not: My first implementation indeed was as you suggest. Then I saw that most of the other testcases use this $config variable. For better readability, I guess. So I wanted to stay in line with that. :)

test/ServiceManagerTest.php
fhein: @Ocramius : That's a question of semantics. What exactly is a service? 

The 'car' is the same, of course. But the 'car' on it's own is not a service, it is a thing. Rental or ownership is required to transform the 'car' from a thing to a service available to you.

If 'being shared' can be seen as a service property, then a service having that property is different from a service not having this property. In my picture a class is just a class. The service manager makes a service out of it.

The alias construct implicitly connects the new (alias based) service to another service definition, but this new service may be different in many means from the original. But it's core object is constructed via the same thing (factory class, the 'car'). 

The relationship between the alias and the service aliased is maintained automatically, even if the definition of the service aliased changes. That is IMO better than multiple registration.

test/ServiceManagerTest.php
```diff
@@ -305,4 +306,31 @@ public function testShouldNotShareAliasesWhichAreNotConfiguredToBeShared()
...
+                    \stdClass::class => true,
+                ],
+            ]
+        );
+        // Obviously the service manager has to have a factory for
+        // each service which is configured not to be shared
+        // As a delicate detail this applies to alias services as
+        // well if the alias is configured not to be shared
+        self::expectException(ServiceNotFoundException::class);
+        $sm->get('alias');
```
Ocramius: Were the assertions removed on purpose?

fhein: This is a second test method to illustrate that a non shared alias has to point to a factory, not to a service, obviously. 

The first test method is one above and still has the assertion.

fhein:
This fix should be applied based on #214. See there.

PR #214, opened 2017-12-29
Fix: Non-Shared alias request returns shared service

fhein:
This PR covers the same bug as #213 but provides the better implementation. See discussion below. 

In the particular case that there is a service defined in the services array (i.e. shared), an alias for that service is defined and configured not to get shared.

If you request this non-shared alias via get(), this will happen:
If a factory is available, the non-shared services gets created via the factory (letting the shared service aside). 
If a factory is not available, the non-shared alias service is cloned from the available shared service.

The only change to the behaviour of sm::get() is, that in case no factory but a service exists for a non-shared alias, the service gets cloned. As of #213 a ServiceNotFound exception was raised.

I did not touch the behaviour, that a service registered via the services member array is ALWAYS shared regardless of any sharing setup.

Is it meaningful to introduce such a fallback factory capability for aliases? It appears to me to be more consistent with the intent, that an alias can be a way to control via requested name, whether a service is shared or not. On the other hand it breaks the rule, that all services which the service manager should be able to create, require a factory definition. This proposal prefers an available factory, but does not require it.

By implementing the __clone member function of the class used to create a service you can do extremely powerful things. Even wizardry. You may request a non-shared service alias not because you need that object. Maybe just to get __clone called. All at the cost of maintability. Having fun can get very expensive.  :)

Is it useful to introduce such a fallback factory capability? I think so. The iterator use case I mentioned in the discussion of #213 and similar cases (have a shared master service and aquire disposable working copies from the service manager via alias) seem to be relevant to me. 

All, what do you think? I'd highly appreciate your opinion and advice.

Btw, I changed get() from ($name, $requestedName) scheme to ($name, $resolvedName). This saves one string assignment and appears to be more consistent with other sm::methods to me.


fhein:
Hm, if the class used for service creation has other usages (beside being the blueprint for a service provided by the service manager), and this other usages include cloning the class, and the class implements __clone, this will probably produce unexpected side effects.

This raises a need for documentation, if this PR should get accepted. In userland probably very few configurations exist which meet this very special setting.

But there may be be the one or other configuration out there, which silently accepts the bug and work with non-shared alias requests producing shared services.

Ocramius:
Overall against this: as already mentioned in #213, an alias **IS** the service. Requesting an alias is the same as requesting the original service under a different name, so configuring an alias should not even be possible (`array_intersect_key()` validation on shared/alias keys and throw exception if there is config for aliases).

Also to be added: usual hint at removing the alias feature overall.

fhein:
The current design as of the master branch does not agree with you. See implementation of `get()`, lines 209ff of servicemanager.php:
 
       // Also do so for aliases; this allows sharing based on service name used.
        if ($requestedName !== $name
            && (($this->sharedByDefault && ! isset($this->shared[$requestedName]))
                || (isset($this->shared[$requestedName]) && $this->shared[$requestedName]))
        ) {
            $this->services[$requestedName] = $object;
        }

You propose a design change which would break BC. The `array_intersect_key` approach you propose would break existing tests (`testAliasToAnExplicitServiceShouldWork()` et al.). Your approach would even break the documentation. Huge impact. There should be strong reasons to do so. What would be the particular benefit from that? Is your preference a reason strong enough to justify to reject a bug fix according to the current design?

This PR fixes a bug honoring the documented intent in the code above.  And it adds a new capability. The latter would evolve the current design a little. People who do not like it don't use it. But is it technically or conceptually bad? I pointed out some benefits this would enable.
 
ServiceManager offers several ways to accomplish the same thing. I also do not like some of the constructs. `ConfigAbstractFactory` is an example. But if I found a bug there I would fix it. I would not search there, of course. :)

Nobody who dislikes aliases must use them. Anyway, they are part of the currently agreed and documented official design. If there are bugs, they have to get fixed. 

Please reconsider. Bug fix PRs are not the appropriate platform for design changes which break BC.

Please would the author(s) of the code above and the several tests and features related to aliases comment on this, too?

fhein:
I closed #213.

Depending on the clone feature getting not accepted I will remove it here. The implementation here is cleaner, more complete and more efficient than #213.

Ocramius:
> Huge impact. There should be strong reasons to do so. What would be the particular benefit from that? Is your preference a reason strong enough to justify to reject a bug fix according to the current design?

The removal was planned for 3.0.0, but then I dropped the ball on it because I didn't follow the design due to lack of time (gotta pay my bills sometimes!).

But yes, aliases are just a royal pain in the butt, there are no advantages, just downsides in using them, and removing them would make the entire ServiceManager another couple percent points faster, while providing a backwards-compatible polyfill was already proven to be possible in https://github.com/zendframework/zendframework/pull/5500, where the performance penalty is only paid by users relying on aliases.

fhein:
I understand. Anyway. as long as the polyfill is not in place and the alias functionality lives in the core,  support and maintainance for that is mandatory overhead, I think.

fhein:
I could transparently outsource the alias functionality to an abstract factory maintaining BC.
This would leave all alias related APIs of the servicemanager intact but would remove alias considerations from has() and get(). sm configuration would be the same, i.e. alias section will stay. The abstract factory would be triggered as the first abstract factory, to maintain precedence of service resolution. 

A new setting could be but must not necessarily be introduced: enable_aliases (default: true). If false, the abstract alias factory would not get created and chained in. With this set to false there would be no alias related performance loss at all.

Without that setting or with true things would generally speed up. Only user supplied abstract factories
would pay the price, that the alias check occurs upfront.

There would not be the necessity to deprecate the feature.

May I proceed?

Ocramius:
I only voiced my own opinion so far, so I would rather wait for someone else from the @zendframework/community-review-team to first check.

fhein:
Ok. Am I right, that the royal pain in the ass you see results from resolution precedence?
Currently, immediate services are resolved first, aliases are resolved next. Alias definitions
supersede any configuration 'inherited' by for example third party modules.

This is, in fact, a major pain in the ass and, I think, a design failure. You can erronously define aliases without knowing that there already is a service or factory of that  name. Even worse, a third party module introduces a new service or factory. The name matches one of your aliases. Ouch.

Very difficult to sign support contracts with that in mind. 

Do you have this in mind? Can you subscribe to my view, that aliases have meaningful uses in an environment, where mentioned side effects can not occur?

I do not have any installations out there where the important main service manager can corrupt third party's configurations and vice versa. The 'own' service manager is ALWAYS a distinct one. Otherwise, as mentioned above, support would be a bloodbath.

With the last commit I introduced an `enable_aliases` option to the service manager configuration. This setting is true by default and grants 100% BC (including the precedence problem). All code does respect this setting. Alias configuration is completely ignored if enable_aliases is false. 

I optimized the code to perform better if no aliases are defined. Performance gain is maximized with `$enableAliases =false`, but you get an advantage if no aliases are set, also.

I did not introduce a warning in case that the sm is configured to ignore aliases and the configuration provided has aliases. Do you think, I should issue a warning in such cases? I believe, there are cases, where people consciously use the alias feature to disable things they would get injected otherwise. 
They exploit a design weakness. My opinion.

I do not longer think, that the alias functionality should be outsourced to a factory. It needs to get tamed. 

Please reconsider your plan to deprecate the feature. Keep in mind, please, that there are people out there interested in using single framework components. Please review the code by looking at it.

akrabat:
FWiW, I think aliases are a useful feature for configuration in the same way that invokables are. I don't really mind what the underlying implementation looks. The key thing to me is that the configuration makes intent clear.

I have to admit that I'm surprised that we allow aliases to be configured differently to their service; that seems bonkers to me. If the service is shared, then all the aliases should be shared too and vice versa. Seems like a misstep - or there's a use-case that I'm not aware of?

Regarding the specifics of this PR:

1. I think that it's dangerous to clone a service to create a non-shared instance. Debugging side-effects of `__clone()` will be hard. Personally, I think that if you try to get a non-shared alias of a shared service where there's no factory, then that's an error condition. The SM shouldn't try to guess intent by cloning some object who's internal state could be anything.

2. The optimisation work around alias performance should be a separate PR to allow for a separate discussion.

fhein:
The clone behaviour is scary to some degree. The use cases I see can easily be covered otherwise. I removed that. I also removed the $enableAliases option, because it is not really needed to gain a performance increase.

fhein:
Closed and replaced by #215.

PR #215, opened 2017-12-31
BUGFIX: 'shared' property of aliases is ignored under certain conditions

fhein:
This PR is a streamline of the discussions about #213 and #214. Both PRs were closed in favour of this one.

To invoke this bug, set `shared_by_default` to `false`. Add a factory for a class. Set `shared` for this class to `true`. Define an alias for that class. Get an object using the class id. Get an alias (1) using the alias id. Get another alias (2) using the alias id.

Acceptable result would be: `object !== alias1 && object !== alias2 && alias1 !== alias2`.

Current result is: `object === alias1 === alias2`, i.e. the returned service is shared.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -180,26 +180,26 @@ class ServiceManager implements ServiceLocatorInterface
         if (isset($this->services[$name])) {
             return $this->services[$name];
         }
-
+        
         // Determine if the service should be shared
         $sharedService = ($this->sharedByDefault && ! isset($this->shared[$name])
             || (isset($this->shared[$name]) && $this->shared[$name]));
-
+        
         // We achieve better performance if we can let all alias
         // considerations out
         if (empty($this->resolvedAliases)) {
             $object = $this->doCreate($name);
-
+            
             // Cache the object for later, if it is supposed to be shared.
             if (($sharedService)) {
                 $this->services[$name] = $object;
             }
             return $object;
         }
-
+        
         // Here we have to deal with requests which may be aliases
         $resolvedName = isset($this->resolvedAliases[$name]) ? $this->resolvedAliases[$name] : $name;
-
+        
         // Can only become true, if the requested service is an shared alias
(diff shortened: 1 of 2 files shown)


fhein:
@Ocramius: There was an issue with travis. It could not create session key because disk quota was exceeded. Is there a way to trigger the tests again without doing a fake commit?

Btw, if you wanted to  put aliases in a factory, I think, all you needed was already in place. A delegator factory could be used, which returned the aliased service. This could be done silently maintaining the aliases configuration key.

Whether or not alias resolution should occur at configuration time or via request is a design devision. Neither for aliases not for the delegators replacing them it is necessary to resolve them at configuraiton time. But I like the configuration time option more, because it is costly and ugly to resolve aliases and detect cycles per request. 

An abstract factory approach would break BC because resulotion precedence was changed. Using a delegator factory approach would maintain resolution precedence. And it would avoid to introduce another factory class. On the other hand, it would increase the overhead of creating a service via alias and I can not see the benefit of doing so.

If it was up to me, I would keep aliases as they are. Further, I would clarify the documentation on aliases, elaborating on the risks and chances as well. I reviewed older code bases and found, that aliases were used intentionally to disable or redirect inherited services.

But I support your idea, that users who omit the use of aliases should not have to pay the price for them.  #215 and #216 together ensure that.

Ocramius:
@fhein restarted

fhein:
That said, an alias can be seen as a lightweight delegator (one, which immediately delivers it's delegate). 
Because delegators are resolved at the same precedence as aliases (assumed aliases would be removed),
delegators can cause the same 'royal pain in the butt' aliases can.  People need to know, what they are doing.

Delegators are a great feature, though, as aliases are, if used conciously with the possible side effects in mind. Please don't deprecate any of both.

fhein:
Now it passed. I wonder, wonder ... :)

fhein:
Can be done better. I'll reconsider.

PR #221, opened 2018-01-05
Performance enhancements for has(), setService(), setFactory(), setShared(), setAlias(), setInvokableClass() and configure()

fhein:
This PR is about performance of the setter APIs (or injector APIs if you like) as well as has(). Here's the current comparison report. Thanks @dantleech for faster than light fix.

	benchmark: SetNewServicesBench
	+------------------------+--------------------+------------------+--------+
	| subject                | suite:develop:mean | suite:PR221:mean | gain   |
	+------------------------+--------------------+------------------+--------+
	| benchSetService        | 1.924µs            | 0.655µs          |  2,94x |
	| benchSetFactory        | 4.183µs            | 1.226µs          |  3,41x |
	| benchSetAlias          | 11.222µs           | 1.872µs          |  5,99x |
	| benchSetAliasOverrided | 34.617µs           | 1.876µs          | 18,45x |
	+------------------------+--------------------+------------------+--------+

This is the develop branch:

	Dumped result to SetNewServices.develop.xml
	suite: 133ec8900bdebd2cfb961e09ca103052bd993f16, date: 2018-01-05, stime: 22:12:29
	+---------------------+------------------------+--------+--------+--------+-----+------------+-------------+-------------+-------------+-------------+------------+--------+--------+
	| benchmark           | subject                | groups | params | revs   | its | mem_peak   | best        | mean        | mode        | worst       | stdev      | rstdev | diff   |
	+---------------------+------------------------+--------+--------+--------+-----+------------+-------------+-------------+-------------+-------------+------------+--------+--------+
	| SetNewServicesBench | benchSetService        |        | []     | 100000 | 100 | 1,375,992b | 1.910110µs  | 1.957413µs  | 1.946003µs  | 2.050120µs  | 0.028245µs | 1.44%  | 1.00x  |
	| SetNewServicesBench | benchSetFactory        |        | []     | 100000 | 100 | 1,375,992b | 4.180240µs  | 4.240743µs  | 4.229557µs  | 4.420250µs  | 0.030047µs | 0.71%  | 2.17x  |
	| SetNewServicesBench | benchSetAlias          |        | []     | 100000 | 100 | 1,375,992b | 11.440650µs | 11.625965µs | 11.601287µs | 12.160700µs | 0.099540µs | 0.86%  | 5.94x  |
	| SetNewServicesBench | benchSetAliasOverrided |        | []     | 100000 | 100 | 1,376,000b | 35.572030µs | 35.950056µs | 35.759907µs | 37.172120µs | 0.312706µs | 0.87%  | 18.37x |
	+---------------------+------------------------+--------+--------+--------+-----+------------+-------------+-------------+-------------+-------------+------------+--------+--------+

Here comes this PR:

	Dumped result to SetNewServices.setAlias.xml
	suite: 133ec89fee8638dd6d892c639ca0556b8090f9b8, date: 2018-01-05, stime: 21:42:44
	+---------------------+------------------------+--------+--------+--------+-----+------------+------------+------------+------------+------------+------------+--------+-------+
	| benchmark           | subject                | groups | params | revs   | its | mem_peak   | best       | mean       | mode       | worst      | stdev      | rstdev | diff  |
	+---------------------+------------------------+--------+--------+--------+-----+------------+------------+------------+------------+------------+------------+--------+-------+
	| SetNewServicesBench | benchSetService        |        | []     | 100000 | 100 | 1,347,528b | 0.640030µs | 0.654438µs | 0.650130µs | 0.680040µs | 0.010230µs | 1.56%  | 1.00x |
	| SetNewServicesBench | benchSetFactory        |        | []     | 100000 | 100 | 1,347,528b | 1.200070µs | 1.222270µs | 1.217291µs | 1.280070µs | 0.013681µs | 1.12%  | 1.87x |
	| SetNewServicesBench | benchSetAlias          |        | []     | 100000 | 100 | 1,347,528b | 1.810100µs | 1.845307µs | 1.848852µs | 1.900110µs | 0.015841µs | 0.86%  | 2.82x |
	| SetNewServicesBench | benchSetAliasOverrided |        | []     | 100000 | 100 | 1,347,536b | 1.780100µs | 1.803603µs | 1.794896µs | 1.870110µs | 0.019256µs | 1.07%  | 2.76x |
	+---------------------+------------------------+--------+--------+--------+-----+------------+------------+------------+------------+------------+------------+--------+-------+

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -121,6 +121,15 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
     }
 
     /**
+     * Override setService for additional plugin validation.
+     */
+    public function setService($name, $service)
+    {
+        $this->validate($service);
+        parent::setService($name, $service);
+    }
+
+    /**
      * {@inheritDoc}
      *
      * @param string $name Service name of plugin to retrieve.
--- a/src/Exception/ContainerModificationsNotAllowedException.php
+++ b/src/Exception/ContainerModificationsNotAllowedException.php
@@ -14,4 +14,14 @@ use DomainException;
  */
 class ContainerModificationsNotAllowedException extends DomainException implements ExceptionInterface
 {
+
+    public function __construct($service)
+    {
+        parent::__construct(sprintf(
+            'The container does not allow to replace/update a service'
+            . ' with existing instances; the following '
+            . 'already exist in the container: %s',
+            $service
+        ));
+    }
 }
--- a/src/Exception/ContainerModificationsNotAllowedException.php
+++ b/src/Exception/ContainerModificationsNotAllowedException.php
@@ -14,10 +14,9 @@ use DomainException;
  */
 class ContainerModificationsNotAllowedException extends DomainException implements ExceptionInterface
 {
-
-    public function __construct($service)
+    public static function fromExistingService($service)
     {
-        parent::__construct(sprintf(
+        return new self(sprintf(
             'The container does not allow to replace/update a service'
             . ' with existing instances; the following '
             . 'already exist in the container: %s',
--- a/src/Exception/CyclicAliasException.php
+++ b/src/Exception/CyclicAliasException.php
@@ -20,7 +20,7 @@ use function sprintf;
 class CyclicAliasException extends InvalidArgumentException
 {
     /**
-     * @param string   conflicting alias key
+     * @param string   $alias conflicting alias key
      * @param string[] $aliases map of referenced services, indexed by alias name (string)
      *
      * @return self
--- a/src/Exception/CyclicAliasException.php
+++ b/src/Exception/CyclicAliasException.php
@@ -7,11 +7,18 @@
 
 namespace Zend\ServiceManager\Exception;
 
+use function array_filter;
+use function array_keys;
+use function array_map;
+use function array_values;
+use function implode;
+use function reset;
+use function serialize;
+use function sort;
 use function sprintf;
 
 class CyclicAliasException extends InvalidArgumentException
 {
-    public static $cycle;
     /**
      * @param string   conflicting alias key
      * @param string[] $aliases map of referenced services, indexed by alias name (string)
@@ -28,16 +35,135 @@ class CyclicAliasException extends InvalidArgumentException
         }
         $cycle .= ' -> ' . $alias . "\n";
 
-        // for testing
-        self::$cycle = $cycle;
         return new self(sprintf(
             "A cycle was detected within the aliases defintions:\n%s",
             $cycle
         ));
     }
 
-    public function getCycle()
+    /**
+     * @param string[] $aliases map of referenced services, indexed by alias name (string)
+     *
+     * @return self
+     */
+    public static function fromAliasesMap(array $aliases)
     {
-        return self::$cycle;
+        $detectedCycles = array_filter(array_map(
+            function ($alias) use ($aliases) {
+                return self::getCycleFor($aliases, $alias);
+            },
+            array_keys($aliases)
+        ));
+
+        if (! $detectedCycles) {
+            return new self(sprintf(
+                "A cycle was detected within the following aliases map:\n\n%s",
+                self::printReferencesMap($aliases)
+            ));
+        }
+
+        return new self(sprintf(
+            "Cycles were detected within the provided aliases:\n\n%s\n\n"
+            . "The cycle was detected in the following alias map:\n\n%s",
+            self::printCycles(self::deDuplicateDetectedCycles($detectedCycles)),
+            self::printReferencesMap($aliases)
+        ));
+    }
+
+    /**
+     * Retrieves the cycle detected for the given $alias, or `null` if no cycle was detected
+     *
+     * @param string[] $aliases
+     * @param string   $alias
+     *
+     * @return array|null
+     */
+    private static function getCycleFor(array $aliases, $alias)
+    {
+        $cycleCandidate = [];
+        $targetName     = $alias;
+
+        while (isset($aliases[$targetName])) {
+            if (isset($cycleCandidate[$targetName])) {
+                return $cycleCandidate;
+            }
+
+            $cycleCandidate[$targetName] = true;
+
+            $targetName = $aliases[$targetName];
+        }
+
+        return null;
+    }
+
+    /**
+     * @param string[] $aliases
+     *
+     * @return string
+     */
+    private static function printReferencesMap(array $aliases)
+    {
+        $map = [];
+
+        foreach ($aliases as $alias => $reference) {
+            $map[] = '"' . $alias . '" => "' . $reference . '"';
+        }
+
+        return "[\n" . implode("\n", $map) . "\n]";
+    }
+
+    /**
+     * @param string[][] $detectedCycles
+     *
+     * @return string
+     */
+    private static function printCycles(array $detectedCycles)
+    {
+        return "[\n" . implode("\n", array_map([__CLASS__, 'printCycle'], $detectedCycles)) . "\n]";
+    }
+
+    /**
+     * @param string[] $detectedCycle
+     *
+     * @return string
+     */
+    private static function printCycle(array $detectedCycle)
+    {
+        $fullCycle   = array_keys($detectedCycle);
+        $fullCycle[] = reset($fullCycle);
+
+        return implode(
+            ' => ',
+            array_map(
+                function ($cycle) {
+                    return '"' . $cycle . '"';
+                },
+                $fullCycle
+            )
+        );
+    }
+
+    /**
+     * @param bool[][] $detectedCycles
+     *
+     * @return bool[][] de-duplicated
+     */
+    private static function deDuplicateDetectedCycles(array $detectedCycles)
+    {
+        $detectedCyclesByHash = [];
+
+        foreach ($detectedCycles as $detectedCycle) {
+            $cycleAliases = array_keys($detectedCycle);
+
+            sort($cycleAliases);
+
+            $hash = serialize(array_values($cycleAliases));
+
+            $detectedCyclesByHash[$hash] = isset($detectedCyclesByHash[$hash])
+                ? $detectedCyclesByHash[$hash]
+                : $detectedCycle;
+        }
+
+        return array_values($detectedCyclesByHash);
     }
 }
--- a/src/Exception/CyclicAliasException.php
+++ b/src/Exception/CyclicAliasException.php
@@ -7,141 +7,28 @@
 
 namespace Zend\ServiceManager\Exception;
 
-use function array_filter;
-use function array_keys;
-use function array_map;
-use function array_values;
-use function implode;
-use function reset;
-use function serialize;
-use function sort;
 use function sprintf;
 
 class CyclicAliasException extends InvalidArgumentException
 {
     /**
+     * @param string   conflicting alias key
      * @param string[] $aliases map of referenced services, indexed by alias name (string)
      *
      * @return self
      */
-    public static function fromAliasesMap(array $aliases)
+    public static function fromCyclicAlias($alias, $aliases)
     {
-        $detectedCycles = array_filter(array_map(
-            function ($alias) use ($aliases) {
-                return self::getCycleFor($aliases, $alias);
-            },
-            array_keys($aliases)
-        ));
-
-        if (! $detectedCycles) {
-            return new self(sprintf(
-                "A cycle was detected within the following aliases map:\n\n%s",
-                self::printReferencesMap($aliases)
-            ));
+        $msg = $alias;
+        $cursor = $alias;
+        while($aliases[$cursor] !== $alias) {
+            $cursor = $aliases[$cursor];
+            $msg .= ' -> '. $cursor;
         }
-
+        $msg .= ' -> ' . $alias . PHP_EOL;
         return new self(sprintf(
-            "Cycles were detected within the provided aliases:\n\n%s\n\n"
-            . "The cycle was detected in the following alias map:\n\n%s",
-            self::printCycles(self::deDuplicateDetectedCycles($detectedCycles)),
-            self::printReferencesMap($aliases)
-        ));
-    }
-
-    /**
-     * Retrieves the cycle detected for the given $alias, or `null` if no cycle was detected
-     *
-     * @param string[] $aliases
-     * @param string   $alias
-     *
-     * @return array|null
-     */
-    private static function getCycleFor(array $aliases, $alias)
-    {
-        $cycleCandidate = [];
-        $targetName     = $alias;
-
-        while (isset($aliases[$targetName])) {
-            if (isset($cycleCandidate[$targetName])) {
-                return $cycleCandidate;
-            }
-
-            $cycleCandidate[$targetName] = true;
-
-            $targetName = $aliases[$targetName];
-        }
-
-        return null;
-    }
-
-    /**
-     * @param string[] $aliases
-     *
-     * @return string
-     */
-    private static function printReferencesMap(array $aliases)
-    {
-        $map = [];
-
-        foreach ($aliases as $alias => $reference) {
-            $map[] = '"' . $alias . '" => "' . $reference . '"';
-        }
-
-        return "[\n" . implode("\n", $map) . "\n]";
-    }
-
-    /**
-     * @param string[][] $detectedCycles
-     *
-     * @return string
-     */
-    private static function printCycles(array $detectedCycles)
-    {
-        return "[\n" . implode("\n", array_map([__CLASS__, 'printCycle'], $detectedCycles)) . "\n]";
-    }
-
-    /**
-     * @param string[] $detectedCycle
-     *
-     * @return string
-     */
-    private static function printCycle(array $detectedCycle)
-    {
-        $fullCycle   = array_keys($detectedCycle);
-        $fullCycle[] = reset($fullCycle);
-
-        return implode(
-            ' => ',
-            array_map(
-                function ($cycle) {
-                    return '"' . $cycle . '"';
-                },
-                $fullCycle
-            )
-        );
-    }
-
-    /**
-     * @param bool[][] $detectedCycles
-     *
-     * @return bool[][] de-duplicated
-     */
-    private static function deDuplicateDetectedCycles(array $detectedCycles)
-    {
-        $detectedCyclesByHash = [];
-
-        foreach ($detectedCycles as $detectedCycle) {
-            $cycleAliases = array_keys($detectedCycle);
-
-            sort($cycleAliases);
-
-            $hash = serialize(array_values($cycleAliases));
-
-            $detectedCyclesByHash[$hash] = isset($detectedCyclesByHash[$hash])
-                ? $detectedCyclesByHash[$hash]
-                : $detectedCycle;
-        }
-
-        return array_values($detectedCyclesByHash);
+            "A cycle was detected within the aliases defintions:\n%s",
+            $msg
+            ));
     }
 }
--- a/src/Exception/CyclicAliasException.php
+++ b/src/Exception/CyclicAliasException.php
@@ -11,6 +11,7 @@ use function sprintf;
 
 class CyclicAliasException extends InvalidArgumentException
 {
+    public static $cycle;
     /**
      * @param string   conflicting alias key
      * @param string[] $aliases map of referenced services, indexed by alias name (string)
@@ -19,16 +20,24 @@ class CyclicAliasException extends InvalidArgumentException
      */
     public static function fromCyclicAlias($alias, $aliases)
     {
-        $msg = $alias;
+        $cycle = $alias;
         $cursor = $alias;
-        while($aliases[$cursor] !== $alias) {
+        while (isset($aliases[$cursor]) && $aliases[$cursor] !== $alias) {
             $cursor = $aliases[$cursor];
-            $msg .= ' -> '. $cursor;
+            $cycle .= ' -> '. $cursor;
         }
-        $msg .= ' -> ' . $alias . PHP_EOL;
+        $cycle .= ' -> ' . $alias . "\n";
+
+        // for testing
+        self::$cycle = $cycle;
         return new self(sprintf(
             "A cycle was detected within the aliases defintions:\n%s",
-            $msg
-            ));
+            $cycle
+        ));
+    }
+
+    public function getCycle()
+    {
+        return self::$cycle;
     }
 }
--- a/src/Exception/InvalidArgumentException.php
+++ b/src/Exception/InvalidArgumentException.php
@@ -8,10 +8,32 @@
 namespace Zend\ServiceManager\Exception;
 
 use InvalidArgumentException as SplInvalidArgumentException;
+use Zend\ServiceManager\Initializer\InitializerInterface;
+use Zend\ServiceManager\AbstractFactoryInterface;
 
 /**
  * @inheritDoc
  */
 class InvalidArgumentException extends SplInvalidArgumentException implements ExceptionInterface
 {
+    public static function fromInvalidInitializer($initializer)
+    {
+        return new self(sprintf(
+            'An invalid initializer was registered. Expected a callable, or an instance of '
+            . '(or valid class name or function name resolving to) "%s", '
+            . 'but "%s" was received',
+            InitializerInterface::class,
+            (is_object($initializer) ? get_class($initializer) : gettype($initializer))
+        ));
+    }
+
+    public static function fromInvalidAbstractFactory($abstractFactory)
+    {
+        return new self(sprintf(
+            'An invalid abstract factory was registered. Expected an instance of or a '
+            . 'valid class name resolving to an implementation of "%s", but "%s" was received.',
+            AbstractFactoryInterface::class,
+            (is_object($abstractFactory) ? get_class($abstractFactory) : gettype($abstractFactory))
+        ));
+    }
 }
--- a/src/Exception/InvalidArgumentException.php
+++ b/src/Exception/InvalidArgumentException.php
@@ -23,7 +23,7 @@ class InvalidArgumentException extends SplInvalidArgumentException implements Ex
             . '(or valid class name or function name resolving to) "%s", '
             . 'but "%s" was received',
             InitializerInterface::class,
-            (is_object($initializer) ? get_class($initializer) : gettype($initializer))
+            is_object($initializer) ? get_class($initializer) : gettype($initializer)
         ));
     }
 
@@ -33,7 +33,7 @@ class InvalidArgumentException extends SplInvalidArgumentException implements Ex
             'An invalid abstract factory was registered. Expected an instance of or a '
             . 'valid class name resolving to an implementation of "%s", but "%s" was received.',
             AbstractFactoryInterface::class,
-            (is_object($abstractFactory) ? get_class($abstractFactory) : gettype($abstractFactory))
+            is_object($abstractFactory) ? get_class($abstractFactory) : gettype($abstractFactory)
         ));
     }
 }
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -345,7 +345,6 @@ class ServiceManager implements ServiceLocatorInterface
             $factories = $this->createFactoriesForInvokables($config['invokables']);
 
             if (! empty($aliases)) {
-                // @todo: This is wrong! These aliases are 'resolved' already
                 $config['aliases'] = (isset($config['aliases']))
                     ? \array_merge($config['aliases'], $aliases)
                     : $aliases;
@@ -520,6 +519,7 @@ class ServiceManager implements ServiceLocatorInterface
      *
      * @param string[]|Initializer\InitializerInterface[]|callable[] $initializers
      *
+     * @return void
      */
     private function resolveInitializer($initializer)
     {
@@ -926,13 +926,6 @@ class ServiceManager implements ServiceLocatorInterface
 
     /**
      * Instantiate abstract factories for to avoid checks during service construction.
-     *
-     * @todo: To construct to avoid is not really an optimization, it's lazyness
-     * AbstractFactories are shared services. Make sure that has() is guarded against
-     * numerical parameters. Handle AbstractFactories as any other shared services.
-     *
-     * @todo: Implement a has() and get() logic which unifies the several arrays
-     * It is a desaster to isset(blah) several times on each request.
      *
      * @param string[]|Factory\AbstractFactoryInterface[] $abstractFactories
      *
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -413,7 +413,7 @@ class ServiceManager implements ServiceLocatorInterface
             $this->mapAliasToTarget($alias, $target);
             return;
         }
-        throw new ContainerModificationsNotAllowedException($alias);
+        throw ContainerModificationsNotAllowedException::fromExistingService($alias);
     }
 
     /**
@@ -429,7 +429,7 @@ class ServiceManager implements ServiceLocatorInterface
             $this->createAliasesAndFactoriesForInvokables([$name => $class ?? $name]);
             return;
         }
-        throw new ContainerModificationsNotAllowedException($name);
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
 
     /**
@@ -445,7 +445,7 @@ class ServiceManager implements ServiceLocatorInterface
             $this->factories[$name] = $factory;
             return;
         }
-        throw new ContainerModificationsNotAllowedException($name);
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
 
     /**
@@ -504,7 +504,7 @@ class ServiceManager implements ServiceLocatorInterface
             $this->services[$name] = $service;
             return;
         }
-        throw new ContainerModificationsNotAllowedException($name);
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
 
     /**
@@ -519,7 +519,7 @@ class ServiceManager implements ServiceLocatorInterface
             $this->shared[$name] = (bool) $flag;
             return;
         }
-        throw new ContainerModificationsNotAllowedException($name);
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
 
     /**
@@ -798,7 +798,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
 
@@ -807,7 +807,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
 
@@ -816,7 +816,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
 
@@ -825,7 +825,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
 
@@ -834,7 +834,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
 
@@ -843,7 +843,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
 
@@ -852,7 +852,7 @@ class ServiceManager implements ServiceLocatorInterface
                 if (! isset($this->services[$service]) || $this->allowOverride) {
                     continue;
                 }
-                throw new ContainerModificationsNotAllowedException($service);
+                throw ContainerModificationsNotAllowedException::fromExistingService($service);
             }
         }
     }
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -28,7 +28,6 @@ use function array_merge_recursive;
 use function class_exists;
 use function get_class;
 use function gettype;
-use function implode;
 use function is_callable;
 use function is_object;
 use function is_string;
@@ -191,43 +190,50 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function get($name)
     {
-        $requestedName = $name;
-
         // We start by checking if we have cached the requested service (this
         // is the fastest method).
-        if (isset($this->services[$requestedName])) {
-            return $this->services[$requestedName];
-        }
-
-        $name = $this->resolvedAliases[$name] ?? $name;
-
-        // Next, if the alias should be shared, and we have cached the resolved
-        // service, use it.
-        if ($requestedName !== $name
-            && (! isset($this->shared[$requestedName]) || $this->shared[$requestedName])
-            && isset($this->services[$name])
-        ) {
-            $this->services[$requestedName] = $this->services[$name];
+        if (isset($this->services[$name])) {
             return $this->services[$name];
         }
 
-        // At this point, we need to create the instance; we use the resolved
-        // name for that.
-        $object = $this->doCreate($name);
+        // Determine if the service should be shared
+        $sharedService = isset($this->shared[$name]) ? $this->shared[$name] : $this->sharedByDefault;
 
-        // Cache it for later, if it is supposed to be shared.
-        if (($this->sharedByDefault && ! isset($this->shared[$name]))
-            || (isset($this->shared[$name]) && $this->shared[$name])
-        ) {
-            $this->services[$name] = $object;
+        // We achieve better performance if we can let all alias
+        // considerations out
+        if (empty($this->resolvedAliases)) {
+            $object = $this->doCreate($name);
+
+            // Cache the object for later, if it is supposed to be shared.
+            if (($sharedService)) {
+                $this->services[$name] = $object;
+            }
+            return $object;
         }
 
-        // Also do so for aliases; this allows sharing based on service name used.
-        if ($requestedName !== $name
-            && (($this->sharedByDefault && ! isset($this->shared[$requestedName]))
-                || (isset($this->shared[$requestedName]) && $this->shared[$requestedName]))
-        ) {
-            $this->services[$requestedName] = $object;
+        // Here we have to deal with requests which may be aliases
+        $resolvedName = isset($this->resolvedAliases[$name]) ? $this->resolvedAliases[$name] : $name;
+
+        // Can only become true, if the requested service is an shared alias
+        $sharedAlias = $sharedService && isset($this->services[$resolvedName]);
+        // If the alias is configured as shared service, we are done.
+        if ($sharedAlias) {
+            $this->services[$name] = $this->services[$resolvedName];
+            return $this->services[$resolvedName];
+        }
+
+        // At this point we have to create the object. We use the
+        // resolved name for that.
+        $object = $this->doCreate($resolvedName);
+
+        // Cache the object for later, if it is supposed to be shared.
+        if (($sharedService)) {
+            $this->services[$resolvedName] = $object;
+        }
+        // Also do so for aliases, this allows sharing based on service name used.
+        // $serviceAvailable is true if and only if we have an alias
+        if ($sharedAlias) {
+            $this->services[$name] = $object;
         }
 
         return $object;
@@ -327,7 +333,7 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function configure(array $config)
     {
-        $this->validateOverrides($config);
+        $this->validateConfig($config);
 
         if (isset($config['services'])) {
             $this->services = $config['services'] + $this->services;
@@ -453,7 +459,8 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function setFactory($name, $factory)
     {
-        $this->configure(['factories' => [$name => $factory]]);
+        $this->validate($name);
+        $this->factories[$name] = $factory;
     }
 
     /**
@@ -475,7 +482,7 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function addAbstractFactory($factory)
     {
-        $this->configure(['abstract_factories' => [$factory]]);
+        $this->resolveAbstractFactory($factory);
     }
 
     /**
@@ -508,7 +515,8 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function setService($name, $service)
     {
-        $this->configure(['services' => [$name => $service]]);
+        $this->validate($name);
+        $this->services[$name] = $service;
     }
 
     /**
@@ -519,7 +527,51 @@ class ServiceManager implements ServiceLocatorInterface
      */
     public function setShared($name, $flag)
     {
-        $this->configure(['shared' => [$name => (bool) $flag]]);
+        $this->validate($name);
+        $this->shared[$name] = (bool) $flag;
+    }
+
+    private function resolveAbstractFactory($abstractFactory)
+    {
+        if (\is_string($abstractFactory) && \class_exists($abstractFactory)) {
+            //Cached string
+            if (! isset($this->cachedAbstractFactories[$abstractFactory])) {
+                $this->cachedAbstractFactories[$abstractFactory] = new $abstractFactory();
+            }
+
+            $abstractFactory = $this->cachedAbstractFactories[$abstractFactory];
+        }
+
+        if ($abstractFactory instanceof Factory\AbstractFactoryInterface) {
+            $abstractFactoryObjHash = \spl_object_hash($abstractFactory);
+            $this->abstractFactories[$abstractFactoryObjHash] = $abstractFactory;
+            return;
+        }
+
+        // Error condition; let's find out why.
+
+        // If we still have a string, we have a class name that does not resolve
+        if (\is_string($abstractFactory)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'An invalid abstract factory was registered; resolved to class "%s" ' .
+                    'which does not exist; please provide a valid class name resolving ' .
+                    'to an implementation of %s',
+                    $abstractFactory,
+                    AbstractFactoryInterface::class
+                )
+            );
+        }
+
+        // Otherwise, we have an invalid type.
+        throw new InvalidArgumentException(
+            sprintf(
+                'An invalid abstract factory was registered. Expected an instance of "%s", ' .
+                'but "%s" was received',
+                AbstractFactoryInterface::class,
+                (is_object($abstractFactory) ? get_class($abstractFactory) : gettype($abstractFactory))
+            )
+        );
     }
 
     /**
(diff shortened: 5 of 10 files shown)


fhein:
Here is an overall comparison of develop against this PR. The benchs of develop which are faster are for your convenience marked with x. The benchs where this PR outoperforms develop beyond believe are marked with !.

@dantleech: This is a showcase where it would help that the comparison report would pull suite::context (as documented) instead of suite::uuid. I had to manually reformat. Not much fun. And, if you want me to beg, I'll do: For you it's just a finger snip to hardcode a comparison column (like 1.24x as in reports=aggregate). Assume that the first --file is the base, the second --file is the - hopefully - faster implementation. Even if you would do that just for me, I bet, tons of users would use that. Please! Please! Please! Pleeeeaaaaseee? :)

@Ocramius, @weierophinney: As long as the current code base has that much room for improvement, I'd reject any BC break request which is tried to be justified with performance,. I'd say: Just go ahead and optimize that f** performance. :) As all what I say, this is not meant personally. But for me it is hard to believe, that this is actually 3.x code. 

If I am allowed to issue two wishes: First one: Please don't bother me again with complaints about redundant variables, which I introduce into test code(!) just to increase the test(!) code's readability. Second one: Please don't ask me to break down this PR (which does not touch any semantics of the ServiceManager API nor does it break bc) into thousands of distinct PRs. If I had to do, I think, this would have been 10-15 PRs yet. And I see 20 or 30 more coming up. Please permit my approach to optimize SM performance. Appreciation would be appreciated.

In C++ you are used to review your code very carefully. Particularily, if you heavily employ meta-programming, you have to deal with compile times of several minutes up to hours. If you release b**s**, even if it is open source, you'll lose acceptance, because nobody will go through death valley, if it's not ok.  (deatch valley: apply a fix and spent hours to recompile). People, who don't have time to waste, will discard immediately, if they see severe issues. Not so with PHP, it seems.

PHP on the other side is the land of the free. Abstraction in PHP merely means: Are you still a string or are you an object already? Not much skills required. But it is extremely powerful. I sincerely love it (complementing c++). I love to generate c++ code using PHP. And: I can implement/generate the code I execute next programatically (!!!). That's awesome. That's great.

The resolution of aliases and in particular the detection of cyclic alias definitions was all but fantastic. I have problems to comprehend, how such code could make it to 3.x.

I am talking about things and methods, and concepts and market acceptance and such. Not about people. Please understand and believe. On the other hand: I don't want to make friends. A want a better, faster, more reliable servicemanager. Because I need it.

I started, but I am not finished yet  with abstractfactories. That they are currently pre-initialized does not optimize anything. I don't comprehend, why people in this particular case only, (!) optimize an isset(xy) call away. ServiceManager's code heavily uses (IMO, overuses) isset(), but in this particular case this was 'optimized' away. An abstractfactory is a shared service with a default factory (something like an abstractfactoryfactory). 

On the other hand there is not even a benchmark in place to measure the cost of pre-initialization of - for example - an abstract factory with an ancestry of 27 parents, and some hundreds or  thousand of members. Not good.

I will - within this single PR - optimizre performance. If, at the end, you like it. accept and integrate. If not, not. The only thing I will do, is make this faster (and maybe streamline naming a bit (resolveAliasesFromPreviouslyResolvedAliases and such describe verbously problems that don't exist. I dropped that). 

I will not break BC, I will not break any semantics. I, in fact,  question some of the semantics right now (why is it possible to set an alias for something which is unknown and can not be produced? Are there use cases for that? I doubt. But there are test cases to make sure, that this works. So I continue to support that. Even better performace improvements were possible, if aliases were only allowed to alias things that exist. The probability of catching unintended side effects would be reduced. (i.e. the need for exception messages which explain god and the world would get reduced or eliminated)).

Back to the code. I did not touch anything regarding invokables yet ,,, because I believed, that they were low prio because deprecated. Btw, my approach originally set delegators high prio, because they are new. I dropped delegator prio. They are not covered by any test. If nobody cares, I will not start caring. 

Anyway, I am surprised what is possible in the land of the free. Delegators were a strike. They silently bypassed all measurements and are in and released. That's a fact. How could that happen? 3.x? Nevertheless, I love them. They are Aliases de Luxe. But anyway, I will focus on features, the community loves (I assume, they love and need them, because there are tested).  

Enough bla bla. Here is the overall comparison of develop againt this PR as of now. For your convenience I marked the benchmarks where develop is better with a x. Benchmarks where this PR hopelessly outperforms develop are marked with a !. The one entry which is marked x! got high priority for me. Can't comprehend currently. But I did not have a look at the invokables-machine yet.

 		$ vendor\bin\phpbench report --file=all.develop.xml --file=all.setAlias.xml --report=hein
		benchmark: FetchCachedServicesBench
		+----------------------------------+------------------- +------------------+
		| subject                          | suite:develop:mean | suite:PR221:mean |
		+----------------------------------+--------------------+------------------+
		| benchFetchFactory1               | 0.393µs            | 0.382µs          |
	x!      | benchFetchInvokable1             | 0.394µs            | 0.937µs          |
		| benchFetchService1               | 0.391µs            | 0.382µs          |
		| benchFetchAlias1                 | 0.392µs            | 0.385µs          |
		| benchFetchRecursiveAlias1        | 0.392µs            | 0.384µs          |
		| benchFetchRecursiveAlias2        | 0.390µs            | 0.383µs          |
		| benchFetchAbstractFactoryService | 2.396µs            | 2.366µs          |
		+----------------------------------+--------------------+------------------+

		benchmark: FetchNewServiceUsingConfigAbstractFactoryAsFactoryBench
		+-------------------------------------+--------------------+------------------+
		| subject                             | suite:develop:mean | suite:PR221:mean |
		+-------------------------------------+--------------------+------------------+
		| benchFetchServiceWithNoDependencies | 4.615µs            | 4.387µs          |
		| benchBuildServiceWithNoDependencies | 4.138µs            | 4.141µs          |
		| benchFetchServiceDependingOnConfig  | 5.256µs            | 5.114µs          |
		| benchBuildServiceDependingOnConfig  | 4.851µs            | 4.720µs          |
		| benchFetchServiceWithDependency     | 5.249µs            | 4.976µs          |
		| benchBuildServiceWithDependency     | 4.711µs            | 4.659µs          |
		+-------------------------------------+--------------------+------------------+

		benchmark: FetchNewServiceUsingReflectionAbstractFactoryAsFactoryBench
		+-------------------------------------+--------------------+------------------+
		| subject                             | suite:develop:mean | suite:PR221:mean |
		+-------------------------------------+--------------------+------------------+
		| benchFetchServiceWithNoDependencies | 3.684µs            | 3.359µs          |
		| benchBuildServiceWithNoDependencies | 3.141µs            | 3.100µs          |
		| benchFetchServiceDependingOnConfig  | 7.070µs            | 6.767µs          |
	x	| benchBuildServiceDependingOnConfig  | 6.503µs            | 6.871µs          |
		| benchFetchServiceWithDependency     | 8.360µs            | 8.090µs          |
		| benchBuildServiceWithDependency     | 7.861µs            | 7.828µs          |
		+-------------------------------------+--------------------+------------------+

		benchmark: FetchNewServiceViaConfigAbstractFactoryBench
		+-------------------------------------+--------------------+------------------+
		| subject                             | suite:develop:mean | suite:PR221:mean |
		+-------------------------------------+--------------------+------------------+
		| benchFetchServiceWithNoDependencies | 5.297µs            | 4.990µs          |
		| benchBuildServiceWithNoDependencies | 4.660µs            | 4.567µs          |
		| benchFetchServiceDependingOnConfig  | 5.867µs            | 5.655µs          |
		| benchBuildServiceDependingOnConfig  | 5.324µs            | 5.181µs          |
		| benchFetchServiceWithDependency     | 5.850µs            | 5.600µs          |
		| benchBuildServiceWithDependency     | 5.238µs            | 5.139µs          |
		+-------------------------------------+--------------------+------------------+

		benchmark: FetchNewServiceViaReflectionAbstractFactoryBench
		+-------------------------------------+--------------------+------------------+
		| subject                             | suite:develop:mean | suite:PR221:mean |
		+-------------------------------------+--------------------+------------------+
		| benchFetchServiceWithNoDependencies | 3.333µs            | 3.100µs          |
		| benchBuildServiceWithNoDependencies | 2.741µs            | 2.731µs          |
		| benchFetchServiceDependingOnConfig  | 6.833µs            | 6.566µs          |
	x	| benchBuildServiceDependingOnConfig  | 6.168µs            | 6.452µs          |
		| benchFetchServiceWithDependency     | 8.211µs            | 7.988µs          |
		| benchBuildServiceWithDependency     | 7.543µs            | 7.519µs          |
		+-------------------------------------+--------------------+------------------+

		benchmark: FetchNewServicesBench
		+----------------------------------+--------------------+------------------+
		| subject                          | suite:develop:mean | suite:PR221:mean |
		+----------------------------------+--------------------+------------------+
	x	| benchFetchFactory1               | 2.620µs            | 2.627µs          |
		| benchBuildFactory1               | 2.165µs            | 2.129µs          |
		| benchFetchInvokable1             | 3.025µs            | 2.878µs          |
		| benchBuildInvokable1             | 2.331µs            | 2.327µs          |
		| benchFetchService1               | 0.392µs           	| 0.383µs          |
		| benchFetchFactoryAlias1          | 2.152µs            | 2.141µs          |
	x	| benchBuildFactoryAlias1          | 2.151µs            | 2.162µs          |
		| benchFetchRecursiveFactoryAlias1 | 2.158µs            | 2.145µs          |
	x	| benchBuildRecursiveFactoryAlias1 | 2.145µs            | 2.147µs          |
	x	| benchFetchRecursiveFactoryAlias2 | 2.148µs            | 2.158µs          |
		| benchBuildRecursiveFactoryAlias2 | 2.156µs            | 2.147µs          |
	x	| benchFetchAbstractFactoryFoo     | 2.300µs            | 2.314µs          |
		| benchBuildAbstractFactoryFoo     | 1.860µs            | 1.836µs          |
		+----------------------------------+--------------------+------------------+

		benchmark: HasBench
		+-------------------------+--------------------+------------------+
		| subject                 | suite:develop:mean | suite:PR221:mean |
		+-------------------------+--------------------+------------------+
		| benchHasFactory1        | 0.444µs            | 0.442µs          |
		| benchHasInvokable1      | 0.448µs            | 0.443µs          |
		| benchHasService1        | 0.445µs            | 0.444µs          |
	x	| benchHasAlias1          | 0.453µs            | 0.456µs          |
	x	| benchHasRecursiveAlias1 | 0.456µs            | 0.457µs          |
		| benchHasRecursiveAlias2 | 0.459µs            | 0.455µs          |
		| benchHasAbstractFactory | 0.792µs            | 0.780µs          |
	x	| benchHasNot             | 0.825µs            | 0.831µs          |
		+-------------------------+--------------------+------------------+

		benchmark: SetNewServicesBench
		+------------------------+--------------------+------------------+
		| subject                | suite:develop:mean | suite:PR221:mean |
		+------------------------+--------------------+------------------+
		| benchSetFactory        | 4.238µs            | 1.224µs          |
	!	| benchSetAlias          | 11.501µs           | 3.011µs          |
	!	| benchSetAliasOverrided | 36.141µs           | 3.001µs          |
		| benchSetService        | 1.924µs            | 0.655µs          |
		+------------------------+--------------------+------------------+

dantleech:
> This is a showcase where it would help that the comparison report would pull suite::context (as documented) instead of suite::uuid. I had to manually reformat. Not much fun

It uses UUID because the `context` might (and usually isn't) set. But it is relatively easy to change the behavior `--report='extends: compare, compare: [ context ]'` or similar.

Could you create an issue on PHPBench with the exact format of the report that you're after?

muglug:
A 970-word essay imploring the maintainers to accept your PR (before they've added a single comment) seems like the wrong approach.

fhein:
Accepted. Will not happen again.

Mit IBM Verse von meinem iPhone aus gesendet

Am 06.01.2018 um 16:57:33 hat notifications@github.com folgendes geschrieben:

fhein:
Essays removed. I felt somewhat treated like McStupid in the first time (other PRs). Did not understand why, And with a bottle of wine input I tend to get verbous.

weierophinney:
@fhein We perform code review for a number of reasons:

- To review correctness of the patch: does it do what it states?
- To point out inconsistencies in coding style. The components become harder to maintain when they are a patchwork of different coding practices. As such, you, as the patch author, might disagree with a change requested, but the _reviewer_ is often requesting a change to make the patch consistent with other code in the component or the other components we maintain.

If we are not pointing out the _why_ behind a request, ask. But do so in a collegial way. For instance, "I disagree with that request; why should I do that?" is combative and defensive, while "Could you explain the rationale behind that request? Is it part of a coding style? If so, can you provide some examples for me?" indicates a collaborative approach indicating you're willing to learn. If the reviewer cannot provide a rationale, then you can point out your own, so that _we_ can learn.

Additionally, in some cases, one reviewer will ask another to jump in. In a previous patch of yours, @Ocramius was asking for _my_ opinion on a potential future change regarding aliases. This may not affect the patch you're making, but if a future change might make the need for your patch obsolete, there's no point in us making a change we do not plan to maintain. Having a discussion around _what we choose to maintain_ is important.

Finally, I read your original essay, and I want to point out something: your aspersions at "how could such code even make it into v3" are totally out of line. Why? Because in a year or two, somebody else is going to come along with a huge performance patch, and ask the same question, and now _you_ are the target of that. Asking such questions is in no way helpful. Software evolves. The language evolves. Our understanding of the capabilities of the language evolves. As such, treat everyone as you would ask us to treat you. When in doubt, refer to our [Code of Conduct](https://github.com/zendframework/zend-servicemanager/blob/master/docs/CODE_OF_CONDUCT.md).

fhein:
I understand, that there can be rationales behind requests to change a PR, which may not be obvious for the contributor. And that the contributor should ask then.

But please allow a review of that particular request, which literally was 'Variable is redundant'.

My code was like

    public function textXYZ() {
        $config = [ ... ];
        $sm =new ServiceManager($config);
        // following code did not use $config
    }    

I did it the same way the other test functions do to be compliant with the coding style. 
In `ServiceManagerTest.php` please review 

     testSetAliasShouldWorkWithRecursiveAlias()
     testAliasToAnExplicitServiceShouldWork()
     testSharedServicesReferencingAliasShouldBeHonored()
     testSharedServicesReferencingInvokableAliasShouldBeHonored()
     testMapsNonSymmetricInvokablesAsAliasPlusInvokableFactory()
     testMapsOneToOneInvokablesAsInvokableFactoriesInternally()
     testShareability($sharedByDefault, $serviceShared, $serviceDefined, $shouldBeSameInstance)
     testCanWrapCreationInDelegators()

It actually did not even come to my mind to ask  "Could you explain the rationale behind that request? Is it part of a coding style? If so, can you provide some examples for me?" What came to my mind, was "What variable do you mean? $config?" and that's what I asked.

I did as requested. Now my test looks different from the others. I still do not understand the rationale behind that. Could you please explain?

Ocramius:
I generally review code from my phone while walking or in a coffee break,
so I only have an overview on the diff displayed, not the entire test.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

fhein:
Am I supposed to know that?

Ocramius:
No, but you asked:

 > I did as requested. Now my test looks different from the others. I still
do not understand the rationale behind that. Could you please explain?

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

fhein:
Thanks for the rationale. It's very personal. 

Regarding the optimizations of the 'setters'. I consider it as bad practice to discard caller supplied information and regather it again. This was done in all setters. What I mean is

    function setService($name, $service) 
    {
        // here we know already that the user wants to set a service
        $this->configure( ['services' => [$name, $service]]); // here we drop that information
        // configure then has to find out again, what (in cases routed from setXYZ) the
        // the user wanted via isset($config['services'])
    }

Is that really a matter of software evolution and PHP language evolution? I don't think so. But maybe I do not have the whole picture. Please explain.

Ocramius:
The setters are not hot path - besides `configure()`, the setters are
extremely edge case usage scenario, and should generally be avoided.
Re-using configure() internally is just to make it simple and nice to read.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

fhein:
Nicer to read than the performance optimized versions? That's how they look now. 

    public function setService($name, $service)
    {
        $this->validate($name);
        $this->services[$name] = $service;
    }

Performance and readibility don't mutually exclude each other.

weierophinney:
> I consider it as bad practice to discard caller supplied information and regather it again.

`configure()` was introduced in v3 as we were striving to make the container immutable once configured. Additionally, we do a few bits of work in there that affect configuration:

- invokables become factory entries, pointing a service name at the `InvokableFactory`, and, in some cases, alias entries.
- alias resolution (to ensure validity)
- override detection

This last part is fairly important; after the container is configured initially, we do not want somebody altering it such that a different service is returned for two subsequent requests to the same service name. This can only be done if all setters use the same logic, and `configure()` is the correct place.

As @Ocramius notes, the various setter methods _should generally not be used_ unless you are building the container service entries programmatically. Typical usage (see the zend-mvc and Expressive skeletons) uses a configuration-driven approach instead, where all services are injected during instantiation.

fhein:
As of my PR the setters and configure utilize the same implemenations to address your bullets. There is no code duplication. `validate` guards against overriding. private `doSetAlias` does all of the alias resolution inclduding cycle detection. Not more required to handle all of this than

    private function doSetAlias($alias, $target)
    {
        $this->aliases[$alias] = $target;
        $this->resolvedAliases[$alias] =
            isset($this->resolvedAliases[$target]) ? $this->resolvedAliases[$target] : $target;

        if ($alias === $this->resolvedAliases[$alias]) {
            throw CyclicAliasException::fromAliasesMap([$alias]);
        }

        if (in_array($alias, $this->resolvedAliases)) {
            $r = array_intersect($this->resolvedAliases, [ $alias ]);
            foreach ($r as $name => $service) {
                $this->resolvedAliases[$name] = $target;
            }
        }
    }

And even this is not matter of software and language evolution. See [Knuth, The Art of Computer Programming](https://en.wikipedia.org/wiki/The_Art_of_Computer_Programming).  

If you review the code, you will find that all setters use the same logic as configure. A setter uses the logic for the particular single request. configure uses said logic to handle all of the settings (requests) provided.

This PR also maintains the current handling of Invokables as you described it.

Please explain the rationale behind deprecating Invokables. Please see #223.

Ocramius:
Invokables provide an alternative way to define an implicit factory. The
fact that two separate config keys exists for invokables and factories
leads to hard-to-override configuration as well as more confusion (from a
user perspective) on priority in service resolution.

In most scenarios, in a ZF app, what happens is that an invokable defined
in one module requires more elaborate configuration needed by consumers,
and adding a factory leads to a service definition collision.

Making the implicit factory explicit simplifies things.

fhein:
Thanks. I will think about all the implications that has. Nothing for a quick shot reply.

Anyway, this PR does not touch how Invokables are handled. Semantics at all is not touched by this PR. It's about performance only.

I opened issues to discuss feature requests and changes that could change behaviour . (#222, #223, #224, #225).

fhein:
> The setters are not hot path

The setters are on my path, at least. And on the path of @malinink also. See #112.

src/Exception/CyclicAliasException.php
```diff
@@ -7,141 +7,37 @@
...
-use function array_values;
-use function implode;
-use function reset;
-use function serialize;
-use function sort;
 use function sprintf;
 
 class CyclicAliasException extends InvalidArgumentException
 {
+    public static $cycle;
```
Ocramius: `private`

src/Exception/CyclicAliasException.php
```diff
@@ -7,141 +7,37 @@
...
     /**
+     * @param string   conflicting alias key
      * @param string[] $aliases map of referenced services, indexed by alias name (string)
      *
      * @return self
      */
-    public static function fromAliasesMap(array $aliases)
+    public static function fromCyclicAlias($alias, $aliases)
```
Ocramius: This is a BC break

src/Exception/CyclicAliasException.php
```diff
@@ -7,141 +7,37 @@
...
-    /**
-     * Retrieves the cycle detected for the given $alias, or `null` if no cycle was detected
-     *
-     * @param string[] $aliases
-     * @param string   $alias
-     *
-     * @return array|null
-     */
-    private static function getCycleFor(array $aliases, $alias)
+    public function getCycle()
```
Ocramius: `private`

src/ServiceManager.php
```diff
@@ -429,7 +422,8 @@ private function configureAliases(array $aliases)
      */
     public function setAlias($alias, $target)
     {
-        $this->configure(['aliases' => [$alias => $target]]);
+        $this->validate($alias);
```
Ocramius: The naming `validate()` is too shallow - need to roughly express what it does in the method name

src/ServiceManager.php
```diff
@@ -248,21 +253,26 @@ public function build($name, array $options = null)
      */
     public function has($name)
     {
-        $name  = $this->resolvedAliases[$name] ?? $name;
```
Ocramius: When rebasing on `develop`, you will notice that this change leads to a test failure

src/ServiceManager.php
```diff
@@ -519,59 +515,52 @@ public function setService($name, $service)
      */
     public function setShared($name, $flag)
     {
-        $this->configure(['shared' => [$name => (bool) $flag]]);
+        $this->validate($name);
+        $this->shared[$name] = (bool) $flag;
```
Ocramius: We should just add `bool` to the argument, but that may indeed be a BC break for subclasses and consumers being too lax. Possibly an issue for `4.0.0` :-\

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -18,137 +18,86 @@ class CyclicAliasExceptionTest extends TestCase
     /**
      * @dataProvider aliasesProvider
      *
+     * @param string   conflicting alias key
```
Ocramius: Changes in this file should be completely reverted

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+     *     provided service name is invalid.
      */
-    private function validateOverrides(array $config)
+    private function validate($service)
     {
-        if ($this->allowOverride || ! $this->configured) {
+        // Important: Next three lines must kept equal to the three
+        // lines of validateArray (see below) which are marked as code
+        // duplicate!
+        if (! isset($this->services[$service]) ?: $this->allowOverride) {
```
weierophinney: While technically correct, the above is confusing syntax. Please change it to:


```php
if (! isset($this->services[$service] || $this->allowOverride) {
```

This produces exactly the same results, but uses a more readable and readily understandable conditional construct.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+     *
+     * @param string[] $services
+     * @param string $type Type of service being checked.
+     * @throws ContainerModificationsNotAllowedException if any
+     *     array keys is invalid.
+     */
+    private function validateArray(array $services)
+    {
+        $keys = \array_keys($services);
+        foreach ($keys as $service) {
```
weierophinney: Inline the `array_keys` operation, please:

```php
foreach (\array_keys($services) as $service) {
```

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+            // This is a code duplication from validate (see above).
+            // validate is almost a one liner, so we reproduce it
+            // here for the sake of performance of aggregated service
+            // manager configurations (we save the overhead the function
+            // call would produce)
+            //
+            // Important: Next three lines MUST kept equal to the first
+            // three lines of validate!
+            if (! isset($this->services[$service]) ?: $this->allowOverride) {
```
weierophinney: Please change as noted above.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+                return;
```
weierophinney: Don't you need to run this against _every_ service name? As such, this should be a `continue` statement, not a `return`...

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+            }
+            throw new ContainerModificationsNotAllowedException(sprintf(
+                'The container does not allow to replace/update a service'
+                . ' with existing instances; the following '
+                . 'already exist in the container: %s',
+                $service
+            ));
```
weierophinney: Considering this duplicates what is present in the `validate()` method above, please create a named constructor in the associated exception to generate an exception with this message.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+     */
+    private function validateArray(array $services)
+    {
+        $keys = \array_keys($services);
+        foreach ($keys as $service) {
+            // This is a code duplication from validate (see above).
+            // validate is almost a one liner, so we reproduce it
+            // here for the sake of performance of aggregated service
+            // manager configurations (we save the overhead the function
+            // call would produce)
```
weierophinney: Did you benchmark the differences between the two approaches? Were the differences measurable? I'm fine with code duplication, but would like to see a 10% or greater improvement to justify it.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
-        if (isset($config['factories'])) {
-            $this->validateOverrideSet(\array_keys($config['factories']), 'factory');
-        }
+        $sections = ['services', 'aliases', 'invokables', 'factories', 'delegators', 'shared'];
 
-        if (isset($config['delegators'])) {
-            $this->validateOverrideSet(\array_keys($config['delegators']), 'delegator');
+        foreach ($sections as $section) {
+            if (isset($config[$section])) {
+                $this->validateArray($config[$section]);
```
weierophinney: Why not the following instead?

```php
foreach (\array_keys($config) as $section) {
    if (! \in_array($section, $sections, true)) {
        continue;
    }
    $this->validateArray($config[$section]);
}
```

`$sections` could even be assigned as a class property to ensure it is cached and does not need to be redefined on each invocation.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
 
-        if (isset($config['lazy_services']['class_map'])) {
-            $this->validateOverrideSet(\array_keys($config['lazy_services']['class_map']), 'lazy service');
+    /**
+     * Assuming that the alias name is valid (see above) resolve/add it.
+     *
+     * @param string $alias
+     * @param string $target
+     */
+    private function doSetAlias($alias, $target)
```
weierophinney: I am really not a fan of the "prefix original method with 'do'" approach. Let's give these methods better names. I'd rename this one to `mapAliasToTarget()`.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+    {
+        // $target is either an alias or something else
+        // if it is an alias, resolve it
+        $this->aliases[$alias] =
+            isset($this->aliases[$target]) ? $this->aliases[$target] : $target;
```
weierophinney: Use null coalesce here, as we do elsewhere:

```php
$this->aliases[$alias] = $this->aliases[$target] ?? $target;
```

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+
+        // a self-referencing alias indicates a cycle
+        if ($alias === $this->aliases[$alias]) {
+            throw CyclicAliasException::fromCyclicAlias($alias, $this->aliases);
+        }
+
+        // finally we have to check if existing incomplete alias definitions
+        // exist which can get resolved by the new alias
+        if (in_array($alias, $this->aliases)) {
```
weierophinney: Make this a strict comparison, please:

```php
if (in_array($alias, $this->aliases, true)) {
```

This ensures type-safety. Additionally, PHP performs strict checks faster.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+            // found some, resolve them
+            foreach ($r as $name => $service) {
+                $this->aliases[$name] = $target;
+            }
         }
     }
 
     /**
-     * Determine if one or more services already exist for a given type.
+     * Instantiate abstract factories for to avoid checks during service construction.
```
weierophinney: "for to" is not grammatically correct. Let's try "Instantiate abstract factories in order to avoid checks during service construction".

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+     * @param string[]|Factory\AbstractFactoryInterface[] $abstractFactories
      *
-     * @param string[] $services
-     * @param string $type Type of service being checked.
-     * @throws ContainerModificationsNotAllowedException if any services
-     *     provided already have instances available.
+     * @return void
      */
-    private function validateOverrideSet(array $services, $type)
+    private function doAddAbstractFactory($abstractFactory)
```
weierophinney: Please rename this method to `resolveAbstractFactoryInstance()`.

src/ServiceManager.php
weierophinney: Also, `array_keys` and `in_array` should be imported, just like all other global functions.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
     {
-        $detected = [];
-        foreach ($services as $service) {
-            if (isset($this->services[$service])) {
-                $detected[] = $service;
+        if (\is_string($abstractFactory) && \class_exists($abstractFactory)) {
```
weierophinney: If these functions are not imported already, please do. Then drop the global resolution, as they'll already be imported.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+            //Cached string
```
weierophinney: Space between a comment opener and the comment text, please.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
+            if (! isset($this->cachedAbstractFactories[$abstractFactory])) {
+                $this->cachedAbstractFactories[$abstractFactory] = new $abstractFactory();
             }
+
+            $abstractFactory = $this->cachedAbstractFactories[$abstractFactory];
         }
 
-        if (empty($detected)) {
+        if ($abstractFactory instanceof Factory\AbstractFactoryInterface) {
+            $abstractFactoryObjHash = \spl_object_hash($abstractFactory);
```
weierophinney: `spl_object_hash` is already imported, so no need to globally qualify it.

src/ServiceManager.php
```diff
@@ -888,86 +802,180 @@ private function createFactoriesForInvokables(array $invokables)
...
-        ));
+        // Error condition; let's find out why.
+
+        // If we still have a string, we have a class name that does not resolve
+        if (\is_string($abstractFactory)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'An invalid abstract factory was registered; resolved to class "%s" ' .
+                    'which does not exist; please provide a valid class name resolving ' .
+                    'to an implementation of %s',
```
weierophinney: Put concatenation operators on the following lines, please.

Also, inline the `sprintf` with the line calling `new`, in order to reduce a level of indentation.

Personally, I'd likely create named constructors for this and the following throw statements, in order to make the code here simpler: `throw InvalidArgumentException::forInvalidAbstractFactoryClassName($abstractFactory);`

src/ServiceManager.php
fhein: Did not do those benchmarks requested. I know from other sources I benchmarked, that function call overhead is significant at least up to PHP 5,6. I will do the benchmarks, but I am pretty sure, that this will not achieve 10%.

Ocramius: > at least up to PHP 5,6

Please do check only latest PHP versions (7.2 or 7.3/master) - optimising performance without a new engine is pretty much useless :S

src/ServiceManager.php
fhein: Ok, I will check the import of all functions. But that's not code I touched. I think, in_array is the only one I introduced.

src/ServiceManager.php
fhein: Ok again. But I did not write that.

src/ServiceManager.php
fhein: I did not write that either. I'll correct.

src/ServiceManager.php
fhein: I'll do. But I did not touch that either. :)

src/AbstractPluginManager.php
```diff
@@ -120,6 +120,15 @@ public function configure(array $config)
...
     }
 
+    /**
+     * Override setService for additional plugin validation.
+     */
+    public function setService($name, $service)
+    {
+        $this->validate($service);
+        parent::setService($name, $service);
+    }
```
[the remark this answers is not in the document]
fhein: To explain: AbstractPluginManager relied on ServiceManager::setService to call configure. Therefore APM implemented configure to do plugin validation. This broke with the new setService implementation of SM. So I had to catch setService in APM.

src/ServiceManager.php
fhein: This PR is based on develop. Please explain a little bit more. All tests pass locally.

fhein:
Currently one test on travis fails in line 294 of ServiceManagerTest.php:
ZendTest\ServiceManager\ServiceManagerTest::testAbstractFactoryShouldBeCheckedForResolvedAliasesInsteadOfAliasNameName

My ServiceManagerTest.php does not contain named test end ends with line 287. I compared several branches, but I could not find a version of this file containing that test. Please help.

fhein:
Thank you for the thorough review. Highly appreciated. In particular, thanks @weierophinney for the pointer to configure() performance.

There is a benchmark already in place (`FetchNewServiceManagerBench.php`) for configure. 

I run my tests usually with the whole suite --revs=100000. Because FetchNewServiceManager can not stand that much revs (or better, my machine can't stand it) I left that out the last days. Shame on me. :(

I was shocked to see that configure() slowed down dramatically. I tracked that down locally to array_intersect which is horribly expensive.

To proof that configure() can accelerated significantly I adapted Tarjan's Strongly connected components algorithm to the alias resolution and cycle detection problem.

Please see the comparison against develop following in the next note.

fhein:
Here are the benchmark results. The test creates a SM with 1,000 factories, invokables, services and aliases each.

	$ vendor\bin\phpbench report --file=FetchNewServiceManagerBench.develop.xml --file=FetchNewServiceManagerBench.221.xml --report=hein
	benchmark: FetchNewServiceManagerBench
	+----------------------------------+--------------------+------------------+--------+
	| subject                          | suite:develop:mean | suite:PR221:mean | factor |
	+----------------------------------+--------------------+------------------+--------+
	| benchFetchServiceManagerCreation | 871.500µs          | 561.000µs        | 1.55x  |
	+----------------------------------+--------------------+------------------+--------+

fhein:
May I please ask again for the `ServiceManagerTest.php` which is different from my local version.

Ocramius:
You need to perform a git rebase

fhein:
To upstream develop?

Ocramius:
Yes, your previous patches were merged, so:

 1. git checkout develop
 2. git pull --ff-only origin develop
 3. git checkout your-branch
 4. git rebase develop

fhein:
There is a bit of work left around aliases. 

Examine if the new algo is satisfying for setAlias also, unification if so, and some minor things more.


I‘ll finish that by tomorrow, I think.

src/ServiceManager.php
fhein: Measurements show that strict in_array is much(!) slower than non-strict in_array. I'd recommend to  type check when inserting to the arrays, so we can be sure we act on strings later. Currently type checking does not occur at all. What do you think? IMO explicit type checks are not necessary. People know what to insert.

fhein:
I finalized alias optimization. I found that the algo suitable for mass updates (configure()) is not suitable to add a single alias and vice versa. 

In the setAlias case, the cost of in_array (non-strict) and array_intersect is affordable. But configure() gets 15 times slower with that. setAlias on the other hand with the mass algo gets 10 times slower.

configure() with the new mass algo in place gets about 33% faster than develop.
setAlias with the new single algo in place gets about 20 times faster than develop.

I optimized the mass algo somewhat. Removed the usage of a stack, because I found with a real huge number of aliases added it could blow the script (memory allocation) where develop still could handle that. Difficult to recognize the Tarjan any longer. It's a simple depth first algo now.

The new version can handle any number of aliases added that develop can.

More tuning was possible if I'd know, what number of aliases can be expected. I don't think, there are users out there, which have a thousand of aliases defined. But there is a bench for that. So I took that as a goal.

Additionally, configure could decide based on the number of aliases provided in $config, whether to use the single or the mass algo.

Anyway, as it is know, it is already a lot faster than develop. Next topic. I am not done yet. :)

weierophinney:
> I finalized alias optimization. I found that the algo suitable for mass updates (configure()) is not suitable to add a single alias and vice versa.

Very interesting!

Could you put a note at the top of `mapAliasToTarget()` indicating this fact? Something along the lines of:

```php
/**
 * Map a single alias to its target.
 * 
 * This is done differently from bulk mapping aliases for performance reasons, as the
 * algorithms for mapping a single item efficiently are different from those of mapping
 * many.
 *
 * @see mapAliasesToTargets()
 */
```

This will help later maintainers understand why the routines are different.

Also, you mention

> Difficult to recognize the Tarjan any longer. It's a simple depth first algo now.

If that's the case, maybe drop the reference to that algorithm in the description?

Otherwise, looks good!

src/Exception/ContainerModificationsNotAllowedException.php
```diff
@@ -14,4 +14,14 @@
  */
 class ContainerModificationsNotAllowedException extends DomainException implements ExceptionInterface
 {
+
+    public function __construct($service)
```
Ocramius: This is a BC break due to constructor changing signature - consider using a named constructor instead (the cyclic dependency detection does that, IIRC)

test/ServiceManagerTest.php
```diff
@@ -288,8 +288,8 @@ public function testAbstractFactoryShouldBeCheckedForResolvedAliasesInsteadOfAli
         $abstractFactory
             ->expects(self::once())
             ->method('canCreate')
-            ->with(self::anything(), 'ServiceName')
-            ->willReturn(true);
+            ->with($this->anything())
+            ->willReturn($this->equalTo('ServiceName'));
```
Ocramius: This test change seems incorrect - the `canCreate` should return a boolean, while the `->with()` is variadic

fhein:
Latest bench of configure().


	| benchmark                   | subject                          | groups | params | revs | its | mem_peak    | best      | mean      | mode      | worst     | stdev   | rstdev | diff  |
	+-----------------------------+----------------------------------+--------+--------+------+-----+-------------+-----------+-----------+-----------+-----------+---------+--------+-------+
	| FetchNewServiceManagerBench | benchFetchServiceManagerCreation |        | []     | 100  | 20  | 16,812,632b | 500.000µs | 511.500µs | 510.020µs | 520.000µs | 5.723µs | 1.12%  | 1.00x |
	+-----------------------------+----------------------------------+--------+--------+------+-----+-------------+-----------+-----------+-----------+-----------+---------+--------+-------+

src/Exception/ContainerModificationsNotAllowedException.php
fhein: Oh, ok. I was told to use a named constructor already. But me stupid was not aware that that‘s a the name of a design pattern. Oh my! :) Now I know.

fhein:
I have still one to three ideas left regarding alias resolutions. Such as preparing tests to find out the point of balance between the two algos regarding the number of aliases. 

I will return to that topic when it comes to fine tuning. For now I am fine with the performance. :)

fhein:
Today I spent two hours or so on a shallow dive into the sources of PHP itself. I currently do not completely comprehend, why an array_intersect can get slower then foreach loops, as it's the case with our alias resolution.

I found that the core code is somewhat inhomogene. Code efficiency depends on the (third party) libraries used and style of the particular author. I do not currently want to engage in PHP core development, but it's good to have the source as a reference and guidance to efficient PHP implementations. 

Regarding alias resolution the combo of in_array/array_intersect is fast for a low number of items you have to in_array and intersect. If you have many these can be handled more efficiently by a loop.

The main limiting factor to the performance of alias resolution currently is,that we have to take care of what I call backwards resolution. That means, that we may have alias definitions, on which has would return false when asked, because the aliased item is neither a service, a factory nor is it creatable by any of the registered abstract factories.

Alias resolution could be unified and could get handled with further increased performance for both many and few, if we define that aliases can only alias existing things, i.e. real services and factories and abstract factory producible things. So far, so easy.

But if we do not want to require a particular ordering of the aliases in a config file, we need to maintain the ability of backwards resolution to some degree, i.e. to backwards resolve the newly added aliases before 'standard resolving' them with the already resolved aliases array. There would remain some, but not all of the performance benefit which could get realized, if correct ordering of aliases was demanded.

I discuss that a bit more in #222 (Resolve Trouble with Aliases).

fhein:
Btw, if there are discussions about this PR somewhere off this line, please share your thoughts.

fhein:
If there are usage scenarios out there, where service manager configurations are reused by several acting SMs, maybe it would make sense to outsource the preparation of a working configuration (with resolved alias, etc.) from a raw configuration (config file) from the service manager core. 

As far as my current understanding goes are aggregated service manager configurations currently built via array_merge (more or less). If there are scenarios. where sm1 takes config1 to provide services, and sm2 takes config2 derived from config1 to deliver services, then sm1 and sm2 currently do redundant work to transform the intersection of config1 and config2 from the  raw config to it's working config. Do such scenarios exist? If so, this could be made faster, too. I do not know if the case of practical relevance. Please let me know.

fhein:
If we would decide to provide a distinct SM configuration transformation stage, we could leave it up to the user to decide if he/she wants to provide a carefully defined working configuration, that SMs can immediately act on, or a raw config (like the current), where they can define aliases referencing aliases, aliases referencing unknown things, aliases , that backwards resolve aliases, deprecated invokables (btw, I stand for undeprectating them (with a syntax like abstract factories (no keys, define an alias if you need a distinct key, resolution precedence considered and handled user friendly)).

If users decide to carefully design their config, they'll achieve tremendous performance improvements. And my current work on SM performance would not be 'on the hot path' any longer, if they would get used to provide clean configs. Hey, chance to get rid off me! If that isn't motivating, what else could be? ;)

fhein:
@Ocramius; What is the plan with this PR? Will you continue to silently merge it partly to develop? Where's the conversation around that? If you actually like the changes so far, you could simply accept this PR as of now. The title is getting to long meanwhile, anyway. I'll open a new PR for the rest of the work on my list. I want a 100% performance increase against develop. But the last 20% will get hard. Too hard, maybe. We'll see.

fhein:
@Ocramius; If I may issue one wish; Please don't inject buggy test procedures to my working set as long as you think fixing the bugs you inject are my task to fix. I am not used to use phpunits mock stuff. Meanwhile my bug fix works (really), but I'm not an expert regarding those mocks and I would like to avoid to become one. Not every problem this world has is mine.

fhein:
Could you please consider this PR to be rebased to master? IMO this one is slowly but surely growing to become a benefit everybody should participate of. Does not depend on ZF milestones IMO. I personally see this more as a bugfix than an enhancement meanwile, I'd grant support, if you want to  support the idea.

A full regression test should be done against this PR before going on to do so. Are there full regression tests in place? 

Even if you decide against a rebase to master, maybe somebody could run this full  regression test just for the fun of it. Results would be very(!) important input for me. If shared.

Is there a documentation somewhere on how I could do a full regression test locally?

Ocramius:
Improvements ALWAYS go to next minor or major, never to patch releases, no
exceptions.

Ocramius:
> Please don't inject buggy test procedures to my working set

That test was added to `develop` and simply exposed a regression: no buggy test.

 > What is the plan with this PR

So far, I haven't been able to review it due to time constraints and also because it keeps changing.

I strongly suggest stopping with additions if this is in a mergeable state. The past PRs were merged manually exactly because of that: a macro-PR was forming while smaller units of work were already "done".

fhein:
> That test was added to develop and simply exposed a regression: no buggy test.

This test assumed that canCreate was called only once. I don't think, that a test should not make such assumptions.  Because of that assumption the bug fix I applied to has() failed to test ok, allthough it was. I think, a mock should mimic the mocked object as far as possible. 

Hey, I just don't like to look into travis red eyes when code is ok. :)

fhein:
> I strongly suggest stopping with additions if this is in a mergeable state. The past PRs were merged manually exactly because of that: a macro-PR was forming while smaller units of work were already "done".

Next optimization often relies on the previous. It would be difficult for me to open a new PR for the next todos based on develop. I would have to branch it from this instead. Would that be ok? Please advice.

Ocramius:
> > That test was added to develop and simply exposed a regression: no buggy test.

 > This test assumed that canCreate was called only once. I don't think, that a test should not make such assumptions.  Because of that assumption the bug fix I applied to has() failed to test ok, allthough it was. I think, a mock should mimic the mocked object as far as possible. 

The test just verifies existing `ServiceManager` behavior, which is to first replace aliases with their actual names, and then ask an `AbstractFactory` about that. The mock fully mimicks an `AbstractFactory` behavior just to prevent BC issues. If we want to break BC there, that's fine, but it needs discussion.

 > It would be difficult for me to open a new PR for the next todos based on develop. I would have to branch it from this instead. Would that be ok? Please advice.

Yes, the suggestion would be to open PRs on your local repository and migrate them here as single parts are merged one by one. That would simplify the review process as well as discussion, which is widening at each commit. As it stands, I didn't even get to read half the patch.

fhein:
>Yes, the suggestion would be to open PRs on your local repository and migrate them here as single parts are merged one by one. That would simplify the review process as well as discussion, which is widening at each commit. As it stands, I didn't even get to read half the patch.

I see. Sorry. This PR is currently in a mergable state. The two change requests shown open are addressed, but not shown as outdated. I will not add further commits to this one.  I will consecutively open new ones based on the one before each.

Would it help if I'd squash all the commits of this PR to just one? That would unmess the diffs, which are currently quite useless.

Ocramius:
@fhein the final diff is what is mostly interesting. I'd simply hold it in the current state and work on a new branch for further improvements

fhein:
Ok. So you can review this one now. I created a new branch based in this branch and will continue there.
If I get you right, you don't want that squash of the thousands of commits in here?

weierophinney:
Correct. It's often instructive to understand how functionality evolved and
to see the precise reasons for a change. Squashing something as large as
this makes those smaller changes harder to understand.

Ocramius:
@fhein many of the commits here are extremely descriptive and useful for somebody jumping in 2 years down the line and trying to understand what was done: please don't squash :-)

src/Exception/CyclicAliasException.php
```diff
@@ -19,6 +19,28 @@
 
 class CyclicAliasException extends InvalidArgumentException
 {
+    /**
+     * @param string   conflicting alias key
```
webimpress: Missing variable name `$alias`

src/Exception/InvalidArgumentException.php
```diff
@@ -8,10 +8,32 @@
...
 class InvalidArgumentException extends SplInvalidArgumentException implements ExceptionInterface
 {
+    public static function fromInvalidInitializer($initializer)
+    {
+        return new self(sprintf(
+            'An invalid initializer was registered. Expected a callable, or an instance of '
+            . '(or valid class name or function name resolving to) "%s", '
+            . 'but "%s" was received',
+            InitializerInterface::class,
+            (is_object($initializer) ? get_class($initializer) : gettype($initializer))
```
webimpress: parenthesis around the expression are useless

src/Exception/InvalidArgumentException.php
```diff
@@ -8,10 +8,32 @@
...
+        ));
+    }
+
+    public static function fromInvalidAbstractFactory($abstractFactory)
+    {
+        return new self(sprintf(
+            'An invalid abstract factory was registered. Expected an instance of or a '
+            . 'valid class name resolving to an implementation of "%s", but "%s" was received.',
+            AbstractFactoryInterface::class,
+            (is_object($abstractFactory) ? get_class($abstractFactory) : gettype($abstractFactory))
```
webimpress: as above, useless parenthesis around the expression

src/ServiceManager.php
```diff
@@ -21,21 +21,18 @@
...
+use function \get_class;
+use function \gettype;
+use function \is_callable;
+use function \is_object;
+use function \is_string;
+use function \spl_autoload_register;
+use function \spl_object_hash;
+use function \sprintf;
+use function \trigger_error;
+use function \in_array;
```
webimpress: We shouldn't have `\` before function names. The same when we importing namespace, we never use `\` at the beginning.

src/ServiceManager.php
```diff
@@ -191,43 +183,50 @@ public function getServiceLocator()
...
-        // service, use it.
-        if ($requestedName !== $name
-            && (! isset($this->shared[$requestedName]) || $this->shared[$requestedName])
-            && isset($this->services[$name])
-        ) {
-            $this->services[$requestedName] = $this->services[$name];
-            return $this->services[$name];
+        // We achieve better performance if we can let all alias
+        // considerations out
+        if (empty($this->aliases)) {
```
webimpress: We can have here `! $this->aliases`

fhein:
Thanks. :) Appreciation appreciated. ;) I'll keep the next PRs as descriptive, but lean.

src/ServiceManager.php
```diff
@@ -842,132 +724,198 @@ private function createLazyServiceDelegatorFactory()
...
+            }
+        }
+    }
+
+    /**
+     * Instantiate abstract factories in order to avoid checks during service construction.
+     *
+     * @param string[]|Factory\AbstractFactoryInterface[] $abstractFactories
+     *
+     * @return void
```
webimpress: `return void` can be removed when function does not return any values

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -15,6 +15,101 @@
  */
 class CyclicAliasExceptionTest extends TestCase
 {
+
+    /**
+     * @dataProvider cyclicAliasProvider
+     *
+     * @param string   conflicting alias key
```
webimpress: Missing variable name `$alias`

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -15,6 +15,101 @@
...
+     * @param string[] $aliases
+     * @param string   $expectedMessage
+     *
+     * @return void
```
webimpress: As above, this can be removed.

test/ServiceManagerTest.php
```diff
@@ -272,7 +272,12 @@ public function testSetAliasShouldWorkWithRecursiveAlias()
         self::assertSame($service, $headAlias);
     }
 
-    public function testAbstractFactoryShouldBeCheckedForResolvedAliasesInsteadOfAliasNameName()
+    public function emulateHas($name)
```
webimpress: Hm... where this method is used? Can't see it.

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -15,6 +15,101 @@
...
+
+    /**
+     * Test data provider for testFromCyclicAlias
+     *
+     * @return string[][]|string[][][]
+     */
+    public function cyclicAliasProvider()
+    {
+        return
+        [
```
webimpress: Please to move the opening array bracket to the line above with `return` statement.

src/ServiceManager.php
fhein: Ok. But only if I guard against empty array in configure(). And, wouldn't isset be faster than '!' ?
I'll check.

test/Exception/CyclicAliasExceptionTest.php
fhein: Ok. Not my code.

src/ServiceManager.php
fhein: Ok. But not my code.

fhein:
@webimpress: Thanks for your review. Appreciated.

fhein:
> That test was added to develop and simply exposed a regression: no buggy test.

@Ocramius: My comment regarding that test was bullshit. You are perfectly right. The test was not buggy. When I change implementation I have to take care that all of the assumptions the according tests make still hold. Sorry.

src/Exception/CyclicAliasException.php
```diff
@@ -19,6 +19,28 @@
 
 class CyclicAliasException extends InvalidArgumentException
 {
+    /**
+     * @param string   $alias conflicting alias key
+     * @param string[] $aliases map of referenced services, indexed by alias name (string)
+     *
+     * @return self
+     */
+    public static function fromCyclicAlias($alias, $aliases)
```
webimpress: Looks like we can have typehint `array` on the second parameter

src/Exception/InvalidArgumentException.php
```diff
@@ -8,10 +8,32 @@
 namespace Zend\ServiceManager\Exception;
 
 use InvalidArgumentException as SplInvalidArgumentException;
+use Zend\ServiceManager\Initializer\InitializerInterface;
+use Zend\ServiceManager\AbstractFactoryInterface;
```
webimpress: Alphabetical order of imports please.

src/ServiceManager.php
```diff
@@ -21,21 +21,18 @@
...
 use function gettype;
-use function implode;
 use function is_callable;
 use function is_object;
 use function is_string;
 use function spl_autoload_register;
 use function spl_object_hash;
 use function sprintf;
 use function trigger_error;
+use function in_array;
```
webimpress: Alphabetical order of imported function please.

fhein:
@weierophinney: Thank you very much. I feel honoured that this piece of work made it.  :)

src/ServiceManager.php
fhein: I addressed this requests in #227 by mistake. But because #227 is more or less work missing here, that might be ok. Please tell me, if not.

weierophinney:
Thanks, @fhein; this one is now merged to develop.

weierophinney:
@fhein Also, be sure to rebase  your other open PRs based on this one against current develop; I did a number of minor consistency changes when merging that may cause conflicts.

fhein:
@weierophinney; Please be aware, that this PR still has that nasty bug I discovered while adding tests to increase code coverage. None of the current tests fails on this bug. A test and a bug fix is provided in #227.
What to do?

fhein:
@weierophinney; Minimum action required: Copy and replace mapAliasesToTargets member function from #227 to develop. This will not include the newly introduced test, but it will restore a stable state for develop. Alternatively you may consider to accept #227. This will add the test which fails on current develp also.

Ocramius:
Can you try and open a new PR just with the test and the proposed fix?
Patching without tests is a no-go

PR #226, opened 2018-01-12
Performance - Abstract factory request cache

fhein:
Work in Progress ...

This patch provides caching for abstract factory requests, so canCreate will be called only once on consecutive requests for the same service.

This feature can be turned on and off via config or anytime during the life time of the service manager. If turned off the cache gets unset. 

Recommended implementations of abstract factories should have an invariant canCreate method. Same request, same answer the life time through. But if users want to employ abstract factories where canCreate answers can change during life time, this feature has to be turned off.

There is currently no benchmark in place which provides meaningful information about the real life performance gain. Such a benchmark should assume some cost for canCreate. I will prepare.

What you can see using the current `FetchNewServiceViaConfigAbstractFactoryBench` is the small amout of overhead this feature introduces for cache administration.


fhein:
Reconsidered this one. The idea is good. The implementation is bullshit. We already have a 'hasCache'. It's called $this->factories. On the other side, the hasNotCache applies to all unknown services, not only for services, which abstract factories do not know.

fhein:
Abstract factories wouldn't have to be idempotent. Idempotency is a funky word. As I see it here, the service manager would get idempotent in a sense that a request which can get resolved to an abstract factory would be resolved to the same abstract factory when the same name gets requested again.

If the abstract factories are designed to produce the same service for a particular name requested (i.e, are idempotent)  this feature would have significantly positive performance impact. I checked several real life scenarios. Most of them use this 'pattern'. 

I'd not recommend to use abstract factories which change behaviour at runtime at all. Problems which may arise can be quite difficult to track down.

On the other hand, the service manager explicitly supports (there is a test for it) to register the same abstract factory twice. I do not completely understand how it would be accomplished to let the the first reference's canCreate return false and the second return true. But that this feature is supported and tested suggests somehow, that there are use cases for abstract factories, which are not idempotent. I find it difficult to imagine what these use cases are.

I'd vote for integration of the abstract factory cache and an option to turn it off for bc and for particular use cases relying on non-idempotent abstract factories.

fhein:
I leave this open for discussion. Implementation is outdated. Too much things to manually merge. If the feature can be agreed on, I'll reimplement.

Ocramius:
Agree with "wouldn't recommend" on mutable factory behaviour, bit that's
still a change in how the system behaves, and someone out there will have
to adapt their code to this patch, which means BC break.

fhein:
Not if configurable and default is bc.

Ocramius:
> Not if configurable and default is bc.

Let's just introduce a BC break, document it and delay for `4.0.0` instead. A configuration flag just complicates things.

fhein:
Do you think so? 'enableAbstractFactoryCache' => true, if people want it, does not seem to be to overcomplicate to me. (?)

We could get some performance measures from practical use to justify further support of that feature.
If it is not worth it, it could be removed with 4.0.0 without bc break. But I believe, the feature will proof its valuable.

Ocramius:
@fhein it is kinda problematic to introduce a flag that just serves as a filler until the next release is out.

fhein:
If we can agree on the feature and it would get default on without configuration option (abstract factory idempotence), then the flag would not harm if present in implementation.

We could issue a deprecation note stating that the flag is ignored.

fhein:
@Ocramius: All other factory types are required to be produce the same service for a given name. Otherwise you could not share any service.

If abstract factories are allowed to change behaviour, then in consequence it must be considered as a bug, that the service manager allows to share a service produced by an abstract factory.

I understand the feature that abstract factory services can be shared as an implicit requirement that abstract factories have to be idempotent just like  any other factory.

The docs do not say a word at all about the (obvious?) requirement of idempotency of factories.

My conclusion: The feature proposed here simply exploits a property all factories (for shared services) are required to have. Or sharing abstract factory services is a bug.

fhein:
This one is obsolete.

PR #230, opened 2018-01-18
Fix alias resolution #221

fhein:
> Can you try and open a new PR just with the test and the proposed fix?

Of course. This PR provides tests and a bug fix for mapAliasesToTargets(). The fix does not cost performance. :)

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -868,7 +868,13 @@ class ServiceManager implements ServiceLocatorInterface
      *
      * This function maps $this->aliases in place.
      *
-     * This algorithm is fast for mass updates through configure().
+     * This algorithm is an adaptated version of Tarjans Strongly
+     * Connected Components. Instead of returning the strongly
+     * connected components (i.e. cycles in our case), we throw.
+     * If nodes are not strongly connected (i.e. resolvable in
+     * our case), they get resolved.
+     *
+     * This algorithm fast for mass updates through configure().
      * It is not appropriate if just a single alias is added.
      *
      * @see mapAliasToTarget above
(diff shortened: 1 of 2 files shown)


src/ServiceManager.php
```diff
@@ -881,17 +887,40 @@ private function mapAliasesToTargets()
...
+                    throw CyclicAliasException::fromCyclicAlias($alias, $this->aliases);
+                }
                 $aCursor = $tCursor;
                 $tCursor = $this->aliases[$tCursor];
-                if ($aCursor === $tCursor) {
+            }
+
+            $tagged[$aCursor] = true;
+
+            if (! isset($stack)) {
```
webimpress: I would set empty array to `$stack` before `while` loop:
```php
$stack = [];
```

and then we can remove this condition and also `unset` in line 923.

src/ServiceManager.php
```diff
@@ -881,17 +887,40 @@ private function mapAliasesToTargets()
...
+                continue;
+            }
+
+            foreach ($stack as $_ => $alias) {
```
webimpress: If we are not using indexes from the array we can skip them in the `foreach`:
```php
foreach ($stack as $alias) {
```

fhein: '$_ => $alias' is faster.

src/ServiceManager.php
fhein: unset is faster.

src/ServiceManager.php
froschdesign: @weierophinney and @webimpress 
`$_` looks weird and it has no meaning. Is this allowed in our coding standard?

fhein: At least as of PHP 5.x foreach with explicit $key => $value is always faster.

fhein: As of 7.x it doesn't seem to make sifnificant difference.

fhein: Not addressed by the coding standards, it seems.

froschdesign: We are humans and we want / must read the code. This type of variable name says nothing and there is also no comment to explain the current usage.
I'm not against a performance boost, but I'm against useless variable names.

webimpress: @froschdesign 
> `$_` looks weird and it has no meaning. Is this allowed in our coding standard?

with new zend-coding-standard we will get an error:
> Variable `_` is not in valid camel caps format

Glad we got rid of this useless variable 😄

src/ServiceManager.php
webimpress: @fhein so you'd like to say that `isset` + `unset` is faster than setting empty array and foreach loop on empty array?

src/ServiceManager.php
fhein: And what about `foreach( $key as $_)` ? 

The current usage of $_ is to indicate that this variable will not get accessed in the loop.

Unused, yes. Useless, no.

src/ServiceManager.php
fhein: Yes. `empty($var)` is an equivalent for  `!isset($var) || $var == false`.

fhein: @fhein not sure if you understand me correctly. I know what is empty and what is the difference between empty and isset. My suggestion was to have:

    $stack = []; // <-- add this line  <- would be done on every iteration
    while (isset($this->aliases[$tCursor])) {
          // ... 
    }

    $tagged[$aCursor] = true;

    foreach ($stack as $alias) {
          // ...
    }
    instead of:

    while (isset($this->aliases[$tCursor])) {
        // ... 
    }
 
    $tagged[$aCursor] = true;

    if (! isset($stack)) { // <-- must be replaced with if ( empty($stack)
      continue;
    }

    foreach ($stack as $alias) {
       // ...
    }

   unset($stack); // <-- remove this line

webimpress: @fhein 
```
if (! isset($stack)) { // <-- must be replaced with if ( empty($stack)
```
no, can be removed at all, because then we will have `foreach` on an empty array, and `unset` will be no needed.

fhein: No performance impact following your suggestion on PHP 7.x. I'll do.

weierophinney:
Thanks, @fhein; merged to develop.

Issue #222, opened 2018-01-08
Discussion: Resolve trouble with aliases.


fhein:
Aliases are a very opinionated topic. They need to be used carefully. I thought about aliases and see two main issues with aliases which are important causes for the trouble aliases can make.

### Incomplete alias definitions

1.  Currently incomplete alias definitions are accepted. By incomplete I mean definitions which can not get resolved to a service or factory using the currently known service manager configuration.
Those dangling aliases can get 'satisfied' by later calls to SetService, SetFactory or even SetAlias. 
There are ways to exploit und meaningfully use that behaviour, but they are rare, I believe, and I don't consider them to be good practice. 
In particular if service manager configurations get aggregated over several modules, often including third-party, this behavoiur can easily introduce side effects which are hard to track down later.
Semantically the name alias suggests, that the aliased thing really exists.
I propose, to disallow incomplete alias definitions, i.e. to require that each alias definition can get resolved to a service or factory already defined when the alias gets defined. This could be accomplished more or less by leaving out backwards resolution of aliases and a check of service presence. 
In order not to require a particular order of alias entries in the servicemanager configuration, I propose to treat a 'mass update' of aliases via array as atomic operation regarding that new behaviour. Alias arrays will get resolved as before, but afterwards all aliases must point to known services.
If required, though currently I would not recommend that, a new service type could be introduced
called future or such, which intentiously expects later resolution. But is there really a demand for that? Please let me know, what you think.

### Alias resolution precedence

2. Resolution precedence changes for an aliased service in the current implementation. It gets higher. If you have a `factory` key in the `factories` array and define an alias called `factory` pointing to something else, then your factory gets unreachable. It is effectively overriden by the alias.
I assume, that in the most cases this happens by mistake. But it can be used intentionally to override a  service. But, in my view, this is more than an alias should do. I think, it is usefule to attach an alternative name for a service, and maybe even have a different `shared` setting. But disabling services IMHO is definitely out of the scope of an alias.
I propose to disable that behaviour. This could get accomplished simply changing get(), so that aliased services will be looked up and resolved last. Doing so, it will still be possible to define the alias as it was before, but a service, factory or even abstract factory provided service would override the alias definition. And not vice versa.

### Conclusion

Obviously both changes would break bc. I am not a fan of bc breaks, but in this case I think it is justifiable, even almost required. Last but not least, some additional performance gain (see #221) could get achieved.

[...]

PR #231, opened 2018-01-18
Bugfixes as of #242 applied to develop + more

fhein:
This PR initially delivered the rest of #221. As several bugs were unrevealed in the process I created according tests and fixes for master with #242.

This PR applies the same fixes to develop recognizing that #221 and #230 were already merged. The issuese fixed are in particular 

1. Preconfigured abstract factories are not resolved.
2. Preconfigured initializers are not resolved.
3. Preconfigured invokables are not resolved.
4. Invokables defined as name => InvokableClass can override delegators and factories.

A performance gain of more than 100% in FetchNewServiceManagerBench.php is a welcome side effect. 

Btw, I should have named this branch initially 'Rest-Of-#221'. 211 was wrong right from the beginning. :(

PR #232, opened 2018-01-19
Performance optimize invokables (bug fix also)

fhein:
Current master and develop branches provide a special treatment for invokables within sm configuration phase. This treatment requires to loop through the invokables array and perform actions and create an InvokableFactory entry for each Invokable and an alias if the name of the invokable is different from the invokable class name. This is a major bottleneck for configuration speed. This treatment alone takes about 30% of overall configuration time.

Here is the benchmark comparing master to this PR.

    benchmark: FetchNewServiceManagerBench
	+----------------------------------+-------------------+------------------+--------+
	| subject                          | suite:master:mean | suite:PR232:mean | factor |
	+----------------------------------+-------------------+------------------+--------+
	| benchFetchServiceManagerCreation | 872.000µs         | 218.000µs        | 4x     |
	+----------------------------------+-------------------+------------------+--------+

The replacement of an invokable by a factory and an alias can also be considered as a bug. Current service resolution precedence is:

1. Services
2. Aliases
3. Delegators
4. Factories
5. Abstract Factories

As @Ocramius said, there are use cases, where users want to provide a factory for an invokable class, because the particular application wants to apply configuration to the invokable object.

By defining an alias for the invokable class, this factory gets disabled efffectively.

This PR changes the treatment of invokables. They are stored in an $invokables array like the other items.
Now invokables get resolved to an InvokableFactory in getFactory now. The new resolution precedence is

1. Services
2. Aliases
3. Delegators
4. Factories
5. Invokables
6. Abstract Factories

I tried to discuss the problems of alias resolution precedence in #222. I do not know if there are discussions around that topic actually. At least nobody wanted to share his/her thoughts since I published that article weeks ago. Trouble with aliases could get eliminated by reducing them to aliases. That would mean to change the resolution precedence to 

1. Services
2. Delegators
3. Factories
4. Invokables
5. Abstract Factories
6. Aliases

### DIFF
--- a/benchmarks/BenchAsset/DelegatorFactoryFoo.php
+++ b/benchmarks/BenchAsset/DelegatorFactoryFoo.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace ZendBench\ServiceManager\BenchAsset;
+
+use Interop\Container\ContainerInterface;
+use Zend\ServiceManager\Factory\DelegatorFactoryInterface;
+
+class DelegatorFactoryFoo implements DelegatorFactoryInterface
+{
+    /**
+     * {@inheritDoc}
+     * @see \Zend\ServiceManager\Factory\DelegatorFactoryInterface::__invoke()
+     */
+    public function __invoke(ContainerInterface $container, $name, callable $callback, array $options = null)
+    {
+        $foo = call_user_func($callback);
+        return $foo;
+    }
+}
--- a/benchmarks/BenchAsset/DelegatorFactoryFoo.php
+++ b/benchmarks/BenchAsset/DelegatorFactoryFoo.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
- * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @link      https://github.com/zendframework/zend-servicemanager for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
--- a/benchmarks/SetNewServicesBench.php
+++ b/benchmarks/SetNewServicesBench.php
@@ -52,12 +52,12 @@ class SetNewServicesBench
             $config['invokables']['invokable_$i'] = BenchAsset\Foo::class;
             $config['delegators']['delegator_$i'] = [ DelegatorFactoryFoo::class ];
         }
-        
+
         $this->initializer = new BenchAsset\InitializerFoo();
         $this->abstractFactory = new BenchAsset\AbstractFactoryFoo();
         $this->sm = new ServiceManager($config);
     }
-        
+
 
     public function benchSetService()
     {
@@ -90,52 +90,51 @@ class SetNewServicesBench
 
         $sm->setAlias('recursiveFactoryAlias1', 'factory1');
     }
-    
+
     public function benchSetInvokableClass()
     {
-        
+
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
         $sm->setInvokableClass(BenchAsset\Foo::class, BenchAsset\Foo::class);
     }
-    
+
     public function benchAddDelegator()
     {
-        
+
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
         $sm->addDelegator(BenchAsset\Foo::class, DelegatorFactoryFoo::class);
     }
-    
+
     public function benchAddInitializerByClassName()
     {
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
         $sm->addInitializer(BenchAsset\InitializerFoo::class);
     }
-    
+
     public function benchAddInitializerByInstance()
     {
-        
+
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
         $sm->addInitializer($this->initializer);
     }
-    
+
     public function benchAddAbstractFactoryByClassName()
     {
-        
+
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
         $sm->addAbstractFactory(BenchAsset\AbstractFactoryFoo::class);
     }
-    
+
     public function benchAddAbstractFactoryByInstance()
     {
-        
+
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
         $sm->addAbstractFactory($this->abstractFactory);
     }
-    
 }
--- a/benchmarks/SetNewServicesBench.php
+++ b/benchmarks/SetNewServicesBench.php
@@ -11,6 +11,7 @@ use PhpBench\Benchmark\Metadata\Annotations\Iterations;
 use PhpBench\Benchmark\Metadata\Annotations\Revs;
 use PhpBench\Benchmark\Metadata\Annotations\Warmup;
 use Zend\ServiceManager\ServiceManager;
+use ZendBench\ServiceManager\BenchAsset\DelegatorFactoryFoo;
 
 /**
  * @Revs(1000)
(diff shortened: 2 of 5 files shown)


benchmarks/BenchAsset/DelegatorFactoryFoo.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
```
weierophinney: On this and other file-level docblocks for new files introduced by this patch, please do the following:

- Use https scheme for all URLs
- Date is 2018

benchmarks/SetNewServicesBench.php
```diff
@@ -80,11 +83,58 @@ public function benchSetAlias()
...
         // @todo @link https://github.com/phpbench/phpbench/issues/304
         $sm = clone $this->sm;
 
         $sm->setAlias('recursiveFactoryAlias1', 'factory1');
     }
+
+    public function benchSetInvokableClass()
+    {
+
+        // @todo @link https://github.com/phpbench/phpbench/issues/304
```
weierophinney: On this and the following new methods:

- Move the `@todo` annotation to a method-level docblock
- Remove the blank line at the top of the method

src/ServiceManager.php
```diff
@@ -413,11 +404,11 @@ public function configure(array $config)
...
     {
-        if (isset($this->services[$alias]) && ! $this->allowOverride) {
-            throw ContainerModificationsNotAllowedException::fromExistingService($alias);
+        if (! isset($this->services[$alias]) || $this->allowOverride) {
+            $this->mapAliasToTarget($alias, $target);
+            return;
         }
-
-        $this->mapAliasToTarget($alias, $target);
+        throw ContainerModificationsNotAllowedException::fromExistingService($alias);
```
weierophinney: Please revert this change; exceptions should be thrown within conditionals for internal consistency and with other components.

src/ServiceManager.php
```diff
@@ -431,11 +422,11 @@ public function setAlias($alias, $target)
...
     {
-        if (isset($this->services[$name]) && ! $this->allowOverride) {
-            throw ContainerModificationsNotAllowedException::fromExistingService($name);
+        if (! isset($this->services[$name]) || $this->allowOverride) {
+            $this->invokables[$name] = $class ?? $name;
+            return;
         }
-
-        $this->createAliasesAndFactoriesForInvokables([$name => $class ?? $name]);
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
```
weierophinney: Invert the conditional and throw from the conditional, please.

src/ServiceManager.php
```diff
@@ -449,23 +440,27 @@ public function setInvokableClass($name, $class = null)
...
-        if (isset($this->services[$name]) && ! $this->allowOverride) {
-            throw ContainerModificationsNotAllowedException::fromExistingService($name);
+        if (! isset($this->services[$name]) || $this->allowOverride) {
+            $this->factories[$name] = $factory;
+            return;
         }
-
-        $this->factories[$name] = $factory;
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
```
weierophinney: Please revert this change; exceptions should be thrown within conditionals for internal consistency and with other components.

src/ServiceManager.php
```diff
@@ -449,23 +440,27 @@ public function setInvokableClass($name, $class = null)
...
      *
-     * @param string $name Service name to map
-     * @param null|string $class Class to which to map; if not provided, $name
-     *     will be used for the mapping.
+     * @param string $name Service name
+     * @param string|callable|Factory\DelegatorFactoryInterface $factory Delegator
+     *     factory to assign.
      */
-    public function mapLazyService($name, $class = null)
+    public function addDelegator($name, $factory)
```
weierophinney: Why did you switch the positions of these two methods? Please revert, and also revert the changes within each.

src/ServiceManager.php
```diff
@@ -511,10 +511,11 @@ public function addInitializer($initializer)
...
     {
-        if (isset($this->services[$name]) && ! $this->allowOverride) {
-            throw ContainerModificationsNotAllowedException::fromExistingService($name);
+        if (! isset($this->services[$name]) || $this->allowOverride) {
+            $this->services[$name] = $service;
+            return;
         }
-        $this->services[$name] = $service;
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
```
weierophinney: Please revert this change; exceptions should be thrown within conditionals for internal consistency and with other components.

src/ServiceManager.php
```diff
@@ -527,11 +528,11 @@ public function setService($name, $service)
...
-        if (isset($this->services[$name]) && ! $this->allowOverride) {
-            throw ContainerModificationsNotAllowedException::fromExistingService($name);
+        if (! isset($this->services[$name]) || $this->allowOverride) {
+            $this->shared[$name] = (bool) $flag;
+            return;
         }
-
-        $this->shared[$name] = (bool) $flag;
+        throw ContainerModificationsNotAllowedException::fromExistingService($name);
     }
```
weierophinney: Please revert this change; exceptions should be thrown within conditionals for internal consistency and with other components.

src/ServiceManager.php
```diff
@@ -868,6 +858,12 @@ private function mapAliasToTarget($alias, $target)
      *
      * This function maps $this->aliases in place.
      *
+     * This algorithm is an adaptated version of Tarjans Strongly
+     * Connected Components. Instead of returning the strongly
+     * connected components (i.e. cycles in our case), we throw.
+     * If nodes are not strongly connected (i.e. resolvable in
+     * our case), they get resolved.
+     *
```
weierophinney: The changes in this method appear to be from #230; please rebase, so that this patch only contains the changes you propose.

src/ServiceManager.php
```diff
@@ -899,25 +915,26 @@ private function mapAliasesToTargets()
      * Instantiate abstract factories in order to avoid checks during service construction.
      *
      * @param string[]|Factory\AbstractFactoryInterface[] $abstractFactories
-     *
-     * @return void
      */
-    private function resolveAbstractFactoryInstance($abstractFactory)
+    private function resolveAbstractFactories($abstractFactories)
```
weierophinney: The changes in this method appear to be from a previously merged commit; please rebase, so that this patch only contains the changes you propose.

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -855,4 +854,28 @@ public function testCrashesOnCyclicAliases()
...
+            ],
+            'aliases' => [
+                'alias1' => 'alias2',
+                'alias2' => 'alias3',
+                'alias3' => stdClass::class,
+            ],
+        ]);
+        $this->assertSame($sm->get('alias1'), $sm->get('alias2'));
+        $this->assertSame($sm->get(stdClass::class), $sm->get('alias1'));
+    }
```
weierophinney: These two changes are from #230; please rebase.

src/ServiceManager.php
fhein: My idea behind that is, that by default a configuration should get assumed to be error free (it finally is after all errors got fixed). Assumed the configuration is error free, then  the `||` clause (with appropriate ordering,  which I am not quite sure yet what the best is) can pass successfully (`true`) after evaluation of just the first expression of the `||` clause at a high probability. 

I implemented it in a way, that one and worst case two evaluations are necessary to determine, that operation can continue, assuming that to continue occurs much more often than failure in production scenarios.

Inverting the condition means, that one maybe two evalutions get necessary to determine not to throw. The assumption here is pessimistic. Because working configs will be error free, I optimized assuming that case. It does not matter, how long it takes to decide to throw.

What do you think? Did I miss to consider something important?  

I hope my English is good enough to express the idea.

Anyway, this is fine tuning.

weierophinney: There are two points I was making:

- Consistency. Throughout the framework, we generally (as in greater than 99% of cases) _throw from a conditional_. The method body is where the work of the method is done. In these cases, you're setting an alias, setting a factory, etc. That's the work of the methods. The conditional should be used to flag _exceptional cases that prevent that work from happening_.

- Common use case. The primary use case for zend-servicemanager is to _configure once_. These setters are not the primary path. As such, having them readable and consistent with the code base is more important than micro-optimizations from conditionals.

If you are worried about the conditional performance, due to needing to evaluate two conditions, put the one most likely to be `false` first — which is how these were done (i.e., checking to see if the service/alias/factory/etc. has already been set). If the service has not been previously set, only the first part of the condition will be evaluated, and we'll then move on to the next statement.

On top of that, boolean checks (`&& ! $this->allowOverride`) are quite fast, and inverting the method workflow for this is truly a micro-optimization at the expense of consistency and readability.

fhein: Ok.

src/ServiceManager.php
fhein: I suggest I rebase to #231, if it be accepted. With #231 I did much you requested.

test/CommonServiceLocatorBehaviorsTrait.php
fhein: I suggest I rebase to #231, if that can get accepted. With #231 I did much you requested.

fhein:
Closed because this primarily a bug fix for invokable handling. See #242.

Issue #236, opened 2018-01-24
Test suite does not cover multiple idempotency scenarios around `ServiceManager#has()` and `ServiceManager#get()`


Ocramius:
From the original issue by @fhein (#235):

 > I got aware of this issue, when I created a feature branch, applied some changes, ran the tests witch result OK. Then I ran the benchmarks and I got errors not related to phpbench but related to the ServiceManager.

 > Tracking this down I found, that the benchmarks do things that are not tested by phpunit. Example: `get()` the same service twice. Or `build()` a service, then `get()` the same service. Or `has()` a service, then `get()` it. Just to mention the trivial cases.

 > ServiceManager can (and probably will) change state when any of this APIs (`build()`,`get()`, `has()`) gets called. There are no tests present to cover that state changes. Not completely true, there are some, but their intention is not to track the state changes. But fortunately they do.

fhein:
See #237.

Ocramius:
@fhein awesome, thanks! Will try prioritising it in the morning 👍

PowerKiKi:
This issue has been solved in #237 and should probably be closed now.

PR #237, opened 2018-01-24
Adds test for issue #236

fhein:
Adds a test addressing the issue described in #236. 

    /**
     * The ServiceManager can change internal state on calls to get,
     * build or has, latter not currently. Possible state changes
     * are caching a factory, registering a service produced by
     * a factory, resolving an alias and so on.
     *
     * This tests performs three consecutive calls to build/get for
     * each registered service to push the service manager through
     * all internal states, thereby verifying that build/get/has
     * remain stable through the internal states.
     */

The test does not make any assumptions about particular internal states the ServiceManager has or might have. The ServiceManager is treated as a blackbox in order to make sure, that the test is agnostic to possible future changes to the implementation of internal states.

### DIFF
--- a/test/CommonServiceLocatorBehaviorsTrait.php
+++ b/test/CommonServiceLocatorBehaviorsTrait.php
@@ -29,6 +29,7 @@ use ZendTest\ServiceManager\TestAsset\SimpleAbstractFactory;
 use function call_user_func_array;
 use function restore_error_handler;
 use function set_error_handler;
+use ZendTest\ServiceManager\TestAsset\SampleFactory;
 
 trait CommonServiceLocatorBehaviorsTrait
 {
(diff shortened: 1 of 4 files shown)


fhein:
Some older tests will become redundant if this will be accepted. I hesitate to cleanup such redundancies, because I didn't implement them and I don't want that people miss their tests.

I offer to care about those redundant tests and clean up, if you think that this is necessary.

Ocramius:
Changed base branch to `develop`. If this needs backporting, we'll probably have to redo the patch via cherry-picking

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+     * from all other objects, and asserts that the services got
+     * reference the same object.
+     *
+     * @see testConsistencyOverInternalStates
+     *
+     * @param ContainerInterface $smTemplate
+     * @param string $name
+     * @param array[] string $test
+     */
+    protected function checkConsistencyOverInternalStates($smTemplate, $name, $test)
```
Ocramius: Make it `private` and rename it as `assert...`: that also remoes the need for the first docblock sentence

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+        $names = array_merge(
+            $config['factories'],
+            $config['invokables'],
+            $config['services'],
+            $config['aliases']
+        );
+        $names[stdClass::class] = true;
+
+        foreach ($names as $name => $_) {
+            foreach ($tests as $test) {
```
Ocramius: I think you intended to use a data provider approach - are you familiar with `@dataProvider`? It would make the tests easier to verify...

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+                    'abstractFactoryAlias'  => stdClass::class
+                ],
+                'abstract_factories' => [
+                    SimpleAbstractFactory::class,
+                ]
+                ];
+
+        $smTemplate = $this->createContainer($config);
+
+        // produce all 3-tuples of 'build' and 'get'
```
Ocramius: So, if I get it correctly, you are taking the current configuration, then creating a number of permutations (triplets) of these two methods to then be called in a sequence?

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+        // $names is not hard coded
+        $names = array_merge(
+            $config['factories'],
+            $config['invokables'],
+            $config['services'],
+            $config['aliases']
+        );
+        $names[stdClass::class] = true;
+
+        foreach ($names as $name => $_) {
```
Ocramius: `foreach (array_keys($names) as $name) {`

test/CommonServiceLocatorBehaviorsTrait.php
Ocramius: Yeah, after inspection, this becomes a data provider where the key is `implode('-', $test) . ' with service ' . $name`

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+     * This tests performs three consecutive calls to build/get for
+     * each registered service to push the service manager through
+     * all internal states, thereby verifying that build/get/has
+     * remain stable through the internal states.
+     *
+     * @see testConsistencyOverInternalStates below
+     *
+     * @param ContainerInterface $smTemplate
+     * @param string $name
+     * @param array[] string $test
```
Ocramius: `string[]`

test/CommonServiceLocatorBehaviorsTrait.php
Ocramius: Errata: if this becomes the test, and the test method is turned into a data provider, this becomes much easier to understand.

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+        foreach ($test as $method) {
+            $object[$method][] = $sm->$method($name);
+            $this->assertTrue($sm->has($name));
+        }
+
+        // if there is more than one object in $object['get']
+        // compare all to the first
+        $nrShared = count($object['get']);
+        for ($i = 1; $i < $nrShared; $i++) {
+            $this->assertSame($object['get'][0], $object['get'][$i]);
```
Ocramius: Use a `foreach` instead

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+        }
+        // objects from object['build'] have to be different
+        // from all other objects
+        foreach ($object['build'] as $idx1 => $nonSharedObj) {
+            foreach ($object['get'] as $sharedObj) {
+                $this->assertNotSame($nonSharedObj, $sharedObj);
+            }
+            foreach ($object['build'] as $idx2 => $nonSharedObj2) {
```
Ocramius: An `assertCount(count($object['build']) - 1, array_diff($object['build'], [$nonSharedObj]))` would also work (although unsure, as I can't remember if diffing is strict). Still, that would allow putting the description of this check in the assertion

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+        // compare all to the first
+        $nrShared = count($object['get']);
+        for ($i = 1; $i < $nrShared; $i++) {
+            $this->assertSame($object['get'][0], $object['get'][$i]);
+        }
+        // objects from object['build'] have to be different
+        // from all other objects
+        foreach ($object['build'] as $idx1 => $nonSharedObj) {
+            foreach ($object['get'] as $sharedObj) {
+                $this->assertNotSame($nonSharedObj, $sharedObj);
```
Ocramius: `assertNotContains()` replaces the entire foreach here

Ocramius:
> Some older tests will become redundant if this will be accepted.

We should try removing them here after the comments above have been addressed/discussed, or else this will pass under the radar again. Also, that is a good chance to figure out whether the older tezts actually cover more "edgy" scenarios.

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+
+        // To allow changes to the config above
+        // $names is not hard coded
+        $names = array_merge(
+            $config['factories'],
+            $config['invokables'],
+            $config['services'],
+            $config['aliases']
+        );
+        $names[stdClass::class] = true;
```
Ocramius: Is this just to make an abstract factory worke? In that case, make the abstract factory recognise a hardcoded string, so that the service name highlights that it should be built via abstract factory

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -879,4 +880,128 @@ public function testCoverageDepthFirstTaggingOnRecursiveAliasDefinitions()
...
+                'aliases' => [
+                    'serviceAlias'          => 'service',
+                    'invokableAlias'        => 'invokable',
+                    'factoryAlias'          => 'factory',
+                    'abstractFactoryAlias'  => stdClass::class
+                ],
+                'abstract_factories' => [
+                    SimpleAbstractFactory::class,
+                ]
+                ];
```
Ocramius: Should a delegator scenario also be added?

test/CommonServiceLocatorBehaviorsTrait.php
fhein: 'get','get','get'
'get'.'get.'build'
...
'build'.'build'.'build'

These sequences are run then.

test/CommonServiceLocatorBehaviorsTrait.php
fhein: I did not know `assertNotContains`. Thanks. :)

test/CommonServiceLocatorBehaviorsTrait.php
fhein: But I don't want [0]. ?

fhein:
Isn't that idempotency thing a general issue? Wouldn't it be good to have all applicable tests wrapped into this one?

fhein:
The current version of the test unreveals a possible bug: delegators can produce shared services allthough `sharedByDefault` is `false`.

Could you please review to make sure, that it's not the test which is buggy. I should write a test for the test. :)

fhein:
Anything left for me to do about this?

PR #238, opened 2018-01-25
Closing the first package of configuration speed improvements. 

fhein:
1, If derived classes define an $abstractFactories member, this setting is ignored.
2. If derived classes define an $initializers member, this setting is ignored.

I will provide the test cases for hard coded configurations in derived classes within this PR.


fhein:
Closed in favour of #231

PR #241, opened 2018-01-27
Improved tests aliases with abstract factories.

jaapio:
There was an error in the testcase `testAbstractFactoryShouldBeCheckedForResolvedAliasesInsteadOfAliasName`
where the returnValueMap of phpunit was used wrong. This caused a false positive. The test was returning an
array which is interpreted as `TRUE`.

I discovered this when I added the `testResolvedAliasFromAbstractFactory`, which should do the opposite form the
earlier mentioned testcase.

Beside that I found a way that the servicemanager was returning `null` instead of returning false when the last fallback
on abstract factories failed.

I'm not sure if the `testAbstractFactoryShouldBeCheckedForResolvedAliasesInsteadOfAliasName` is still doing what was intended to do. So that might be a point of discussion.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -278,6 +278,8 @@ class ServiceManager implements ServiceLocatorInterface
                 return true;
             }
         }
+
+        return false;
     }
 
     /**
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -285,18 +285,15 @@ class ServiceManagerTest extends TestCase
             ],
         ]);
 
-        $valueMap = [
-            ['Alias', false],
-            ['ServiceName', true],
-        ];
-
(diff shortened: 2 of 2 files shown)


fhein:
You are right. It was me ‚fixed‘ this test and left it broken with #221. I am sorry. 

The issue is fixed in #231.

Ocramius:
@fhein btw, no need to be sorry - the original patch was legitimate, we simply missed it 👍

Ocramius:
Manually merged via 54f5a52d113adaa33222bfcbf807e2b001d5d94c

jaapio:
Cool, thanks for merging

fhein:
:+1: . I'am somwhat embarrased to have introduced this hack without knowing what I did. :)

PR #244, opened 2018-01-30
FIX: Make tests platform agnostic

fhein:
Standardizes asserts comparing (file based) strings to "\n" line delimiter. Without this patch the test suite reports nine failures on Windows.

Currently I have to phpcbf in order to make the tests pass. Problem with that is, that after the phpcbf run git sees almost all files changed.


fhein:
Ok, there is git config setting called `autocrlf`. This seems to be set `true` by default on Win installations. Turn that off and make sure your editors use "\n" instead of "\r\n", then all works without this patch.

Nevertheless I propose to apply this patch. If a user opened a generated file with an arbitrary editor, which silently converts line delimiters to native, corrected the generated file manually to find if the test would pass with that corrections, he would get disappointed.

Ocramius:
Files are supposed to maintain their `\n` line endings, and that is also
what you will get from distribution composer packages, so I don't think the
patch should be applied. `\r\n` or `\n\r` should never ever be generated.
If any code generator does that, we need to instead adapt generator to
produce `\n`

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -28,7 +28,9 @@ public function testFromAliasesMap(array $aliases, $expectedMessage)
         $exception = CyclicAliasException::fromAliasesMap($aliases);
 
         self::assertInstanceOf(CyclicAliasException::class, $exception);
-        self::assertSame($expectedMessage, $exception->getMessage());
+        $expectedMessage = str_replace(PHP_EOL, "\n", $expectedMessage);
+        $message = str_replace(PHP_EOL, "\n", $exception->getMessage());
+        self::assertSame($expectedMessage, $message);
```
weierophinney: A better approach would be to update the exception to use `PHP_EOL` instead of `\n` in these cases.

That said, glancing at both the class files and the test files, the `\n` characters you are correcting for here are committed as `\n`, and any changes would be due to local configuration of git or your IDE, both of which should then show that there are _repository changes_. As such, I agree with @Ocramius that this is not a change we want to bring in.

fhein: Ok. Windows users could do a `git config --global core.autocrlf false` and make sure to that the editors use unix line delimiters.

PR #187, opened 2017-03-28
Replace container-interop with PSR-11 container interface

michaelmoussa:
This PR replaces use of the `container-interop` package with the PSR-11 container interface. This is obviously a BC break and would need to be released in ServiceManager 4.0 along with a migration guide.

Since `zend-eventmanager` comes into play when ProxyManager is used for lazy instantiation of services (via its own dependency on `zend-code`, which in turn depends on `zend-eventmanager`), I'll be submitting an equivalent PR on that repository shortly and will link it here when it's up.

PR #245, opened 2018-01-31
update to psr/container and remove interop/container

samsonasik:
probably can be initial work for servicemanager v4 which uses `psr-container`, should can be cherry-picked when needed.


webimpress:
@samsonasik Have you seen [dev-4.0 branch](https://github.com/zendframework/zend-servicemanager/tree/dev-4.0)? I think this is already done there...

samsonasik:
latest changes are develop branch, and dev-4.0 seems not yet rebased to develop

weierophinney:
Duplicates #187

PR #248, opened 2018-02-02
Added more test for shared services (aliases, factories, invokables)

webimpress:
During the work on expressive @weierophinney set shared service to `false`  for `ResponseInterface` service. It means every call of the `get` method on the service manager will create new instance of `ResponseInterface`. Please see:
https://github.com/zendframework/zend-expressive/pull/543#discussion_r165378017

In this PR I'd like to check if `shared` services work correctly with all types of services and then provide the same functionality for other containers:
- [zend-pimple-config](https://github.com/zendframework/zend-pimple-config) - see: https://github.com/zendframework/zend-pimple-config/pull/6
- [zend-auradi-config](https://github.com/zendframework/zend-auradi-config)

to have exactly the same behaviour with different container managers.

This is just basics: I'm testing here factories and invokables and both with aliases. Aliases complicates all a bit, because then we can have many cases, let say:

```php
`aliases` => [
    'alias' => 'service',
],
'invokables' => [
    'service' => MyService::class,
],
'shared' => [
    // all different combinations of these:
    // 'alias' => false,
    // 'service' => true,
],
```

And if we include also `shared_by_default` I'm not sure what should be expected results in all of these cases.

Tests in this PR covers part from: https://github.com/zendframework/zend-servicemanager/blob/9f35a104b8d4d3b32da5f4a3b6efc0dd62e5af42/test/CommonServiceLocatorBehaviorsTrait.php#L38-L100
but these are just for services created through `factories`.

Probably we should add more tests, but I don't know what should be the behaviour, for example with config:
```php
'services' => [
    'service' => new MyService(),
],
'shared' => [
    'service' => false,
],
```
should we return always cloned instance of `service` or `shared` does not work with `services` only with factories/invokables/abstract_factories/delegators... ?

**Update:**
Maybe we should extract these tests to separate repository and use them for all containers integration? I guess we'd like to have the same behaviour of containers with the same configuration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,8 @@
         "ocramius/proxy-manager": "^2.1.1",
         "phpbench/phpbench": "^0.13.0",
         "phpunit/phpunit": "^6.4.4",
-        "zendframework/zend-coding-standard": "~1.0.0"
+        "zendframework/zend-coding-standard": "~1.0.0",
+        "zendframework/zend-container-config-test": "^0.2.1"
     },
     "provide": {
         "psr/container-implementation": "^1.0"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "bba5f6d05be1ae91bc424cd3510f5bbc",
+    "content-hash": "f280e96d0b05401d973f1fdd15dc0c90",
     "packages": [
         {
             "name": "psr/container",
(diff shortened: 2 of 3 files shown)


fhein:
We should throw, if there is no factory to create an instance of 'service' defined also. I proposed cloning some time ago, and got the answer, that this was not appriciated, because it would be dangerous.

If your interested, I'll find out, what the particular discussion was.

Danger is an interesting term in a context, where developers completely control, what's going on.

webimpress:
I've created repo (WIP) with common container test cases:
https://github.com/webimpress/zend-container-test

We would be able to use it with zend-servicemanager, zend-pimple-config and zend-auradi-config at least right now.

Ocramius:
@webimpress this looks like an improvement, since the newly introduced tests do not fail. I suggest we move it to `develop`

webimpress:
@Ocramius It's now ready. I've changed branch target to `develop` and used [`zendframework/zend-container-config-test`](https://github.com/zendframework/zend-container-config-test) library instead, as there we have common tests for SM compatible containers.
All these tests I've implemented here before are now in that separate library in separate trait I used here.

Ocramius:
Thanks @webimpress!

PR #242, opened 2018-01-28
BUGFIXES: Processing of preconfigured settings and invokable resolution

fhein:
This PR adds tests and fixes for several issues around configuration and resolution of Invokables.

1. Preconfigured abstract factories are not resolved.
2. Preconfigured initializers are not resolved.
3. Preconfigured invokables are not resolved.
4. Invokables defined as `name => InvokableClass` can override delegators and factories.

A performance gain more than 100% in FetchNewServiceManagerBench.php is a welcome side effect.

PR #253, opened 2018-02-08
BUGFIX: Fix bugs documented in #242 (deployable without conflict to develop also, see #254)

fhein:
This PR fixes some severe bugs. Service manager does not process preconfiguration provided by child classes. Please see #242 for details.

I'd highly appreciate to get some qualified feedback on this. Same for #254.

All it would take to invalidate all the work done and time spent here was to merge something else upfront, as happened before by merging replace-this-with-self-for-no-particular-reason-and-no-measurable-benefit before this one. I would not consider such proceeding as being serious and responsible.


Ocramius:
Same resolution as https://github.com/zendframework/zend-servicemanager/pull/254#issuecomment-364026562

PR #254, opened 2018-02-08
BUGFIX: Fix bugs documented in #242 (see #253 also, which is the same PR against master)

fhein:
This PR fixes some severe bugs. Service manager does not process preconfiguration provided by child classes. 

This  PR is literally the same as #253 but with develop branch as target. Both PRs address the problems described with #242. For details please review the documentation of #242.

Both PRs (this and #253) can be merged automatically to their respectice branches.

Assuming the bugfixes are important and would get applied, I would revoke (close) #231 and reimplement it based on those eventually accepted #253 and #254. Without additional changes develop will not feature all of the  performance improvement #231 provides.

I'd highly appreciate to get some qualified feedback on this.

All it would take to invalidate all the work done and time spent here was to merge something else upfront, as happened before by merging something like replace-this-with-self-for-no-particular-reason-and-no-particular-benefit before this one. I would not consider such proceeding as being serious and responsible.


Ocramius:
Closing here: I don't welcome your patches anymore, no matter how useful they may be.

Between having a faster ServiceManager and not having to discuss things being constantly under the assumption that the @zendframework/community-review-team is acting in malice, I prefer not having your patches at all.

Please do feel free to contribute to other OSS projects instead.

PR #258, opened 2018-02-10
Updated configuring service manager documentation - invokables

webimpress:
Invokables are used in zend-expressive a lot and these were missing in the docs.

I know, these were deprecated with v3 but there are still used in zend-expressive and these are not even marked as deprecated, so imho docs should be up to date.

### DIFF
--- a/docs/book/configuring-the-service-manager.md
+++ b/docs/book/configuring-the-service-manager.md
@@ -4,6 +4,9 @@ The Service Manager component can be configured by passing an associative array 
 constructor. The following keys are:
 
 - `services`: associative array that maps a key to a service instance.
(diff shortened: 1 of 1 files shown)


docs/book/configuring-the-service-manager.md
```diff
@@ -130,6 +148,16 @@ class MyObjectFactory implements FactoryInterface
     }
 }
 
+// or without implementing the interface:
+class MyObjectFactory
+{
+    public function __invoke(ContainerInterface $container, $requestedName)
+    {
+        $dependency = $container(Dependency::class);
```
vaclavvanik: should be `$dependency = $container->get(Dependency::class);`

weierophinney:
Thanks, @webimpress!

PR #261, opened 2018-02-28
Detect cyclic dependency using reflection abstract factory

asgrim:
This is a new feature whereby cyclic dependencies are detected in a bit of an edge case. Example illustrated here: https://gist.github.com/asgrim/72f3ee4c25b901ffed58501657ea98da

I've written the test which demonstrates what I'm trying to do, but the implementation I feel is really sub-par however. My plan to implement was simply to check if `$requestedName === $type` (i.e. the requested service name is the same as the parameter type being used), but in practice this doesn't work because `$requestedName` is actually the *resolved* name:

https://github.com/zendframework/zend-servicemanager/blob/develop/src/ServiceManager.php#L209-L222

On L209 the alias is resolved, and on L222 `doCreate` is called with the `$resolvedName` so we never have the actual originally requested service name.

Therefore the only solution I could come up with for now was to check if the `$container` is a `ServiceManager` and reverse-engineer the service name to the original alias. Really not keen on this though, but I'm lacking on any better ideas. Feedback welcome here! :+1:

It's worth adding that in real terms, this is a user error, but I just wanted to catch this situation and provide some useful feedback (instead of something unclear `Maximum function nesting level of '500' reached, aborting!`), so if there's not a practical way to solve this problem, it simply may not be worth the effort.

### DIFF
--- a/src/AbstractFactory/ReflectionBasedAbstractFactory.php
+++ b/src/AbstractFactory/ReflectionBasedAbstractFactory.php
@@ -10,8 +10,11 @@ namespace Zend\ServiceManager\AbstractFactory;
 use Psr\Container\ContainerInterface;
 use ReflectionClass;
 use ReflectionParameter;
+use ReflectionProperty;
+use Zend\ServiceManager\Exception\InvalidServiceException;
 use Zend\ServiceManager\Exception\ServiceNotFoundException;
 use Zend\ServiceManager\Factory\AbstractFactoryInterface;
+use Zend\ServiceManager\ServiceManager;
 
 use function array_map;
 use function class_exists;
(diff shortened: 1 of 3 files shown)


weierophinney:
@asgrim Is this still a WIP? I've scheduled for 3.4.0, but if it's not ready, I'll wait...

Ocramius:
Given the hacky nature of this, I wouldn't merge it at all, to be honest...

weierophinney:
This repository has been closed and moved to laminas/laminas-servicemanager; a new issue has been opened at https://github.com/laminas/laminas-servicemanager/issues/13.

weierophinney:
This repository has been moved to laminas/laminas-servicemanager. If you feel that this patch is still relevant, please re-open against that repository, and reference this issue. To re-open, we suggest the following workflow:
 - Squash all commits in your branch (`git rebase -i origin/{branch}`)
 - Make a note of all changed files (`git diff --name-only origin/{branch}...HEAD
 - Run the laminas/laminas-migration tool on the code.
 - Clone laminas/laminas-servicemanager to another directory.
 - Copy the files from the second bullet point to the clone of laminas/laminas-servicemanager.
 - In your clone of laminas/laminas-servicemanager, commit the files, push to your fork, and open the new PR.
 We will be providing tooling via laminas/laminas-migration soon to help  automate the process.

PR #263, opened 2018-03-04
remove invokable redundant condition

pine3ree:
In `createFactoriesForInvokables` when `$name === $class` in line 869 then in line 870
```php
$factories[$name] = Factory\InvokableFactory::class;
//is equivalent to
$factories[$class] = Factory\InvokableFactory::class;
```

so there is no need for a conditional assignment since we always use the `$class` name as service key (and in cases where `$name !== $class`, aliases are added via `createAliasesForInvokables`)

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -866,11 +866,6 @@ class ServiceManager implements ServiceLocatorInterface
     {
         $factories = [];
         foreach ($invokables as $name => $class) {
-            if ($name === $class) {
-                $factories[$name] = Factory\InvokableFactory::class;
-                continue;
-            }
-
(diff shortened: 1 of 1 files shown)


Ocramius:
@pine3ree as this is an improvement, please target `develop`

pine3ree:
@Ocramius 
I'm closing this! I just checked the develop branch and a pull with the same effect is already there (along with a useful merging of the two methods and the optional 2nd parameter for setInvokableClass, changes that I also wished for)
sorry and thanks

Ocramius:
@pine3ree no need to be sorry: thanks for looking into this!

PR #268, opened 2018-05-08
ReflectionBasedAbstractFactory should not be expected to instantiate private constructors

asgrim:
I noticed that when using the `ReflectionBasedAbstractFactory`, it chokes when classes with `private` constructors are used. These types of classes should be ignored from the start, so I've updated `canCreate` to return `false` where any class has a `private` constructor.

### DIFF
--- a/src/AbstractFactory/ReflectionBasedAbstractFactory.php
+++ b/src/AbstractFactory/ReflectionBasedAbstractFactory.php
@@ -137,7 +137,7 @@ class ReflectionBasedAbstractFactory implements AbstractFactoryInterface
         return class_exists($requestedName) && $this->canCallConstructor($requestedName);
     }
 
-    private function canCallConstructor(string $requestedName) : bool
+    private function canCallConstructor($requestedName)
     {
         $constructor = (new ReflectionClass($requestedName))->getConstructor();
 
--- a/src/AbstractFactory/ReflectionBasedAbstractFactory.php
+++ b/src/AbstractFactory/ReflectionBasedAbstractFactory.php
@@ -1,8 +1,8 @@
 <?php
 /**
- * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
+ * @see       https://github.com/zendframework/zend-servicemanager for the canonical source repository
  * @copyright Copyright (c) 2016-2018 Zend Technologies USA Inc. (http://www.zend.com)
- * @license   http://framework.zend.com/license/new-bsd New BSD License
+ * @license   https://github.com/zendframework/zend-servicemanager/blob/master/LICENSE.md New BSD License
  */
 
 namespace Zend\ServiceManager\AbstractFactory;
--- a/src/AbstractFactory/ReflectionBasedAbstractFactory.php
+++ b/src/AbstractFactory/ReflectionBasedAbstractFactory.php
@@ -141,11 +141,7 @@ class ReflectionBasedAbstractFactory implements AbstractFactoryInterface
     {
         $constructor = (new ReflectionClass($requestedName))->getConstructor();
 
-        if ($constructor === null) {
-            return true;
-        }
-
-        return $constructor->isPublic();
+        return $constructor === null || $constructor->isPublic();
     }
 
     /**
--- a/src/AbstractFactory/ReflectionBasedAbstractFactory.php
+++ b/src/AbstractFactory/ReflectionBasedAbstractFactory.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
- * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (http://www.zend.com)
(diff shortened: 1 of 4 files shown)


src/AbstractFactory/ReflectionBasedAbstractFactory.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
- * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Year range: `2016-2018`

src/AbstractFactory/ReflectionBasedAbstractFactory.php
```diff
@@ -135,10 +135,27 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
 
     /**
      * {@inheritDoc}
+     * @throws \ReflectionException
```
webimpress: Hm... I know it could be thrown, but I don't think we should have it here.

src/AbstractFactory/ReflectionBasedAbstractFactory.php
```diff
@@ -135,10 +135,27 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
...
      */
     public function canCreate(ContainerInterface $container, $requestedName)
     {
-        return class_exists($requestedName);
+        return class_exists($requestedName) && $this->canCallConstructor($requestedName);
+    }
+
+    /**
+     * @param string $requestedName
+     * @return bool
```
webimpress: Definitely we don't need two above doc-block comments, as these are duplicated with method signature.
The third one, yes, it's possible but as it is not throw explicitly I would remove it as well.

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
- * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Year range, please.

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -17,6 +17,8 @@
 
 class ReflectionBasedAbstractFactoryTest extends TestCase
 {
+    private $container;
```
webimpress: Please add docblock:
```php
/** @var ContainerInterface|ObjectProphecy $container */
```

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -38,6 +40,20 @@ public function testCanCreateReturnsFalseForNonClassRequestedNames($requestedNam
         self::assertFalse($factory->canCreate($this->container->reveal(), $requestedName));
     }
 
+    /**
+     * @throws \ReflectionException
```
webimpress: No point of having it here.

asgrim:
@webimpress thanks for review! updated

src/AbstractFactory/ReflectionBasedAbstractFactory.php
```diff
@@ -138,7 +138,18 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
...
+
+    private function canCallConstructor(string $requestedName) : bool
+    {
+        $constructor = (new ReflectionClass($requestedName))->getConstructor();
+
+        if ($constructor === null) {
+            return true;
+        }
+
+        return $constructor->isPublic();
```
weierophinney: Could these be boiled down to:

```php
return $constructor === null || $constructor->isPublic();
```

?

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -38,6 +42,17 @@ public function testCanCreateReturnsFalseForNonClassRequestedNames($requestedNam
         self::assertFalse($factory->canCreate($this->container->reveal(), $requestedName));
     }
 
+    public function testCanCreateReturnsFalseWhenConstructorIsPrivate() : void
```
weierophinney: It also looks like the new method would return true if _no_ constructor is defined, correct? If so, that deserves its own test case.

asgrim: Correct; test case added

GeeH:
Can this be merged now @weierophinney

Ocramius:
@GeeH @asgrim we need this to re-target `develop`, since it is a bug fix

weierophinney:
Um... Bugfixes target **master**, features/improvements target develop; can
you clarify what you meant please, @Ocramius?

PR #269, opened 2018-06-08
Undefined $factory notice fix

demiankatz:
When a plugin manager uses a static class method as a factory (i.e. `FactoryClass::factoryMethod`), as of release 2.7.10, it causes an undefined variable notice to be thrown due to an uncovered else case. This pull request provides a test case to demonstrate the problem as well as a fix. This problem does not appear to exist in the 3.x code, so it is presumably a side effect of backporting to 2.7. However, for projects depending on earlier versions of the service manager, it can be a significant problem and should be corrected -- that is why I am targeting this PR against the release-2.7 branch.

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -355,6 +355,8 @@ abstract class AbstractPluginManager extends ServiceManager implements ServiceLo
         } elseif (is_array($callable)) {
             // reset both rewinds and returns the value of the first array element
             $factory = reset($callable);
+        } else {
+            $factory = null;
         }
 
         if ($factory instanceof Factory\InvokableFactory) {
--- a/test/TestAsset/FooStaticFactory.php
+++ b/test/TestAsset/FooStaticFactory.php
@@ -1,0 +1,18 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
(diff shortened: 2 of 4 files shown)


src/AbstractPluginManager.php
```diff
@@ -355,6 +355,8 @@ protected function createServiceViaCallback($callable, $cName, $rName)
         } elseif (is_array($callable)) {
             // reset both rewinds and returns the value of the first array element
             $factory = reset($callable);
+        } else {
+            $factory = null;
```
malukenho: Can't it be initialized as `null`  first?

test/TestAsset/FooStaticFactory.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
```
malukenho: `2018`

src/AbstractPluginManager.php
demiankatz: Sure, just a question of style preference. I did it this way to avoid an unnecessary double-assignment, but if you prefer to trim a couple of lines by eliminating the else clause, that works fine too. I'll change it if you wish.

test/TestAsset/FooStaticFactory.php
demiankatz: Thanks -- fixed!

Ocramius:
Thanks @demiankatz! Released in `2.7.11` :-)

PR #275, opened 2018-11-17
Introduce psr to interop container decorator for plugin managers

Xerkus:
Enables plugin managers to accept non-interop psr containers as creation
context and transparently wrap as interop for backward compatibility

Provide a narrative description of what you are trying to accomplish:

- [ ] Are you fixing a bug?
  - [ ] Detail how the bug is invoked currently.
  - [ ] Detail the original, incorrect behavior.
  - [ ] Detail the new, expected behavior.
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.
  - [ ] Add a `CHANGELOG.md` entry for the fix.

- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?
    This PR enables plugin managers extending from abstract plugin manager to accept non-interop psr containers as creation context in a backward compatible manner. It is useful when alternative containers are used eg with zend-expressive.
    Decoration is necessary since FactoryInterface typehints against interop and won't work with pure PSR https://github.com/zendframework/zend-servicemanager/blob/0aea76faf40350852fd6570cb4c1eb60a18ac872/src/Factory/FactoryInterface.php#L10 https://github.com/zendframework/zend-servicemanager/blob/0aea76faf40350852fd6570cb4c1eb60a18ac872/src/Factory/FactoryInterface.php#L37
    This change is required since FactoryInterface typehints against interop container
  - [x] How will users use the new feature?
    Users now can pass non-interop psr containers as a constructor argument to plugin managers
  - [ ] Base your feature on the `develop` branch, and submit against that branch.
    Unable as `develop` branch have changes targeted at next major release
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [x] Add tests for the new feature.
  - [ ] Add documentation for the new feature.
    Change is internal, no documentation is needed.
  - [x] Add a `CHANGELOG.md` entry for the new feature.

- [ ] Is this related to quality assurance?
  

- [ ] Is this related to documentation?

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -8,6 +8,7 @@
 namespace Zend\ServiceManager;
 
 use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface as PsrContainerInterface;
 use Zend\ServiceManager\Exception\InvalidServiceException;
 
 /**
@@ -52,6 +53,11 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
      */
     public function __construct($configInstanceOrParentLocator = null, array $config = [])
     {
+        if ($configInstanceOrParentLocator instanceof PsrContainerInterface
+            && ! $configInstanceOrParentLocator instanceof ContainerInterface
+        ) {
+            $configInstanceOrParentLocator = new PsrContainerDecorator($configInstanceOrParentLocator);
+        }
         if (null !== $configInstanceOrParentLocator
             && ! $configInstanceOrParentLocator instanceof ConfigInterface
             && ! $configInstanceOrParentLocator instanceof ContainerInterface
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -56,6 +56,11 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
         if ($configInstanceOrParentLocator instanceof PsrContainerInterface
             && ! $configInstanceOrParentLocator instanceof ContainerInterface
         ) {
+            /**
+             * {@see \Zend\ServiceManager\Factory\FactoryInterface} typehints
+             * against interop container and as such cannot accept non-interop
+             * psr container. Decorate it as interop.
+             */
             $configInstanceOrParentLocator = new PsrContainerDecorator($configInstanceOrParentLocator);
         }
         if (null !== $configInstanceOrParentLocator
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -48,7 +48,7 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
      * factories; for $config, {@see \Zend\ServiceManager\ServiceManager::configure()}
      * for details on its accepted structure.
      *
-     * @param null|ConfigInterface|ContainerInterface $configInstanceOrParentLocator
+     * @param null|ConfigInterface|ContainerInterface|PsrContainerInterface $configInstanceOrParentLocator
      * @param array $config
      */
     public function __construct($configInstanceOrParentLocator = null, array $config = [])
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -1,8 +1,8 @@
 <?php
 /**
- * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
- * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
- * @license   http://framework.zend.com/license/new-bsd New BSD License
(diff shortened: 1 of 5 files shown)


src/AbstractPluginManager.php
```diff
@@ -47,11 +48,16 @@
...
+     * @param null|ConfigInterface|ContainerInterface|PsrContainerInterface $configInstanceOrParentLocator
      * @param array $config
      */
     public function __construct($configInstanceOrParentLocator = null, array $config = [])
     {
+        if ($configInstanceOrParentLocator instanceof PsrContainerInterface
+            && ! $configInstanceOrParentLocator instanceof ContainerInterface
+        ) {
+            $configInstanceOrParentLocator = new PsrContainerDecorator($configInstanceOrParentLocator);
+        }
```
weierophinney: zend-servicemanager requires container-interop/container-interop `^1.2`, and the `ContainerInterface` it ships _extends the PSR-11 `ContainerInterface`_. As such, you can simplify this tremendously by changing the check to test the container against `Psr\Container\ContainerInterface` instead. There's no reason to decorate in that instance.

src/PsrContainerDecorator.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
```
webimpress: remove empty line

src/PsrContainerDecorator.php
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @link      http://github.com/zendframework/zend-servicemanager for the canonical source repository
```
webimpress: usually we are using `@see` and https links please

src/PsrContainerDecorator.php
```diff
@@ -0,0 +1,69 @@
...
+ * @copyright Copyright (c) 2005-2018 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: As new file, just 2018 year

src/PsrContainerDecorator.php
```diff
@@ -0,0 +1,69 @@
...
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
```
webimpress: please use link to the LICENSE.md in the repository (see other repos for reference)

Xerkus: I am not sure if i should change dockblock in a single file. I think it should be project wide PR targeting develop

src/PsrContainerDecorator.php
Xerkus: Are those copyrights on per project or per file basis? Per project would need whole range.

Xerkus:
Those changes target 3.x only and will be undone before merge into develop for 4.0 btw.

src/PsrContainerDecorator.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Finds an entry of the container by its identifier and returns it.
+     *
+     * @param string $id Identifier of the entry to look for.
+     *
+     * @throws NotFoundExceptionInterface  No entry was found for **this** identifier.
+     * @throws ContainerExceptionInterface Error while retrieving the entry.
+     *
+     * @return mixed Entry.
```
weierophinney: This whole docblock can become:

```php
/**
 * {@inheritDoc}
 */
```

as it does not differ from the `ContainerInterface`.

(Same comment applies to the `has()` method.)

test/AbstractPluginManagerTest.php
```diff
@@ -60,6 +63,32 @@ public function testInjectCreationContextInFactories()
...
+        $invokableFactory->method('__invoke')
+            ->will($this->returnArgument(0));
+
+        $config = [
+            'factories' => [
+                'creation context container' => $invokableFactory,
+            ],
+        ];
+
+        $container     = $this->getMockBuilder(PsrContainerInterface::class)
```
charma:
```suggestion
        $container = $this->getMockBuilder(PsrContainerInterface::class)
```

webimpress: @charma Please see: https://discourse.zendframework.com/t/poll-zend-coding-standard-2-aligning-statements-arrays-and-comment-tags/729

charma: Thanks for the Info. I didnt know.

src/PsrContainerDecorator.php
weierophinney: Per file.

src/PsrContainerDecorator.php
weierophinney: My policy during review is:

- New files get the new format, listing the current year only in the date range.
- Existing files that have changed get updated to the new format, and have the date range from the previous docblock.

PR #276, opened 2018-11-21
Allow delegators to decorate the `$options` array

tux-rampage:
This PR allows delegator factories to decorate the `$options` array that is passed to the actual factory.

Use case:

```php

class MyService
{
    public function __construct(array $types)
    { /* ... */ }
}

$serviceConfig = [
    'factories' => [
        MyService::class => fuction($container, $name, $options) {
            return new MyService($options['types'] ?? []);
        }
    ],
    'delegators' => [
        MyService::class => [
            function($container, $name, $factory, $options) {
                $options['types'] = array_merge(
                    $options['types'] ?? [],
                    [ 'string' => 1 ]
                );
                $factory($options);
            },
        ],
    ],
]; 

```

- [ ] Are you fixing a bug?
  - [ ] Detail how the bug is invoked currently.
  - [ ] Detail the original, incorrect behavior.
  - [ ] Detail the new, expected behavior.
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.
  - [ ] Add a `CHANGELOG.md` entry for the fix.

- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?
  - [x] How will users use the new feature?
  - [x] Base your feature on the `develop` branch, and submit against that branch.
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [x] Add tests for the new feature.
  - [x] Add documentation for the new feature.
  - [x] Add a `CHANGELOG.md` entry for the new feature.

- [ ] Is this related to quality assurance?
  

- [ ] Is this related to documentation?

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -603,10 +603,10 @@ class ServiceManager implements ServiceLocatorInterface
      */
     private function createDelegatorFromName($name, array $options = null)
     {
-        $creationCallback = function () use ($name, $options) {
+        $creationCallback = function (array $decoratedOptions = null) use ($name, $options) {
             // Code is inlined for performance reason, instead of abstracting the creation
             $factory = $this->getFactory($name);
-            return $factory($this->creationContext, $name, $options);
+            return $factory($this->creationContext, $name, $decoratedOptions ?? $options);
         };
 
         foreach ($this->delegators[$name] as $index => $delegatorFactory) {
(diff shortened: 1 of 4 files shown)


src/ServiceManager.php
```diff
@@ -640,12 +640,22 @@ private function createDelegatorFromName($name, array $options = null)
...
+                    $this->creationContext,
+                    $name,
+                    $creationCallback,
+                    $decoratedOptions ?? $options
+                );
             };
         }
 
-        return $creationCallback($this->creationContext, $name, $creationCallback, $options);
+        return $creationCallback();
```
tux-rampage: `$creationCallback` never accepted any parameters before this PR. Passing parameters had no effect.

tux-rampage:
@weierophinney It was just an idea and maybe feature creep indeed. I had some idea in correlation with zend-di's auto wire factory, but I don't remeber it anymore tbh. Of course there are better solutions and I might have come up with a better one finally. 

Maybe it's better to drop the `$options` array for service manager altogether?

Feel free to close if nobody has a need for this use case.

weierophinney:
> Maybe it's better to drop the `$options` array for service manager altogether?

That's one of the things I'd like to do for zend-servicemanager v4.

I'll close this for now; if you run across the problem you were trying to solve, and find that this is the best way to solve it, re-open and update the PR to detail it. :smile:

PR #273, opened 2018-10-13
Added PHP 7.3 support

michalbundyra:
Updated Travis CI configuration to run all tests on PHP 7.3 with lowest/locked/latest dependencies

PR #277, opened 2018-12-20
Added PHP 7.3 support - updated Travis CI configuration

webimpress:


webimpress:
The build is still failing, because we do not have zend-code which support PHP 5.6 and 7.3 so we can't do it in easy way (without additional conditions in .travis.yml) to pass all tests with lowest/locked/latest deps. As discussed with @Xerkus closing in favour of previous PR #273 with target `develop`, where we support only PHP 7.1+.

weierophinney:
@Xerkus and @webimpress: Just a note: I want develop to target 3.4.0, as we have features that we should get into that release (including the preliminary support for PSR-11 in plugin managers). To that end, I willl be creating a release-4.0 branch soon so we can target a new major version bump for this component, and reverting the change in develop that bumped the PHP version.

Issue #279, opened 2019-02-15
Bug with array_merge_recursive


thomasvargiu:
I'm trying to do something like this:

```php
$config1 = [
    'factories`=> [
        // factories
    ],
    'lazy_services' => [
        'class_map' => [
            'Foo' => 'Foo',
        ],
    ],
];
```

```php
$config2 = [
    'factories`=> [
        // factories
    ],
    'lazy_services' => [
        'class_map' => [
            'Foo' => 'Foo',
        ],
    ],
];

$container->configure($config1);
$container->configure($config2);
```

This will result on internal `lazy_services` property the following:

```php
$this->lazy_services = [
    'class_map' => [
        'Foo' => [
            'Foo',
            'Foo',
        ],
    ].
];
```

The problem is the use of `array_merge_recursive()` function in [this line](https://github.com/zendframework/zend-servicemanager/blob/a1ed6140d0d3ee803fec96582593ed024950067b/src/ServiceManager.php#L360).

Same problem with `delegators` property.

I found this issue using `zend-mvc` and an already configured container, then `Zend\ModuleManager\Listener\ServiceListener` reconfigure it adding configuration from ModuleManager features.

weierophinney:
This repository has been closed and moved to laminas/laminas-servicemanager; a new issue has been opened at https://github.com/laminas/laminas-servicemanager/issues/4.

PR #280, opened 2019-02-15
Hotfix #279

thomasvargiu:
Fix issue #279


src/ServiceManager.php
```diff
@@ -34,6 +34,7 @@
 use function spl_object_hash;
 use function sprintf;
 use function trigger_error;
+use Zend\Stdlib\ArrayUtils;
```
samsonasik: this use class statement can be moved up to the collection of use class statements.

thomasvargiu: You're right. Done.

test/ServiceManagerTest.php
```diff
@@ -363,4 +364,34 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
...
+        $serviceManager->configure($config);
+
+        $ref = new ReflectionObject($serviceManager);
+        $delegatorsProperty = $ref->getProperty('delegators');
+        $delegatorsProperty->setAccessible(true);
+        $lazyServicesProperty = $ref->getProperty('lazyServices');
+        $lazyServicesProperty->setAccessible(true);
+
+        self::assertSame($config['delegators'], $delegatorsProperty->getValue($serviceManager));
+        self::assertSame($config['lazy_services'], $lazyServicesProperty->getValue($serviceManager));
```
webimpress: I am not convinced to that test, as we are using reflection to check internal state of the object. Is there something what we can notice "outside" that it's wrong configured? I'm afraid that test like that is not future-proof.

weierophinney: @thomasvargiu I agree with @webimpress here. What is the _side effect_ that occurs currently? The test should be checking to ensure that side effect does not occur with your patch.

With delegators, that's generally pretty easy - you can check that a class is not double-decorated after retrieval. With lazy services, I'm assuming that the fact that you get an array value in a class map will likely throw an error on retrieval — if that's the case, test that you do not get that exception.

test/ServiceManagerTest.php
```diff
@@ -363,4 +364,34 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
...
+    {
+        $config = [
+            'delegators' => [
+                'Foo' => 'FooDelegator',
+            ],
+            'lazy_services' => [
+                'class_map' => [
+                    'Foo' => 'Foo',
+                ],
+            ]
```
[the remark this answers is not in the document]
thomasvargiu: Fixed

test/ServiceManagerTest.php
thomasvargiu: @webimpress @weierophinney I changed it, configuration used in test was wrong.
But I also changed the implementation to merge delegators. Let me know.

weierophinney:
@thomasvargiu —

- https://travis-ci.org/zendframework/zend-servicemanager/jobs/508580256#L881

Run `composer cs-fix`, commit, and push, please. :smile:

thomasvargiu:
@weierophinney fixed it

src/ServiceManager.php
```diff
@@ -361,8 +361,8 @@ public function configure(array $config)
             $this->factories = $config['factories'] + $this->factories;
         }
 
-        if (isset($config['delegators'])) {
-            $this->delegators = array_merge_recursive($this->delegators, $config['delegators']);
+        if (isset($config['delegators']) && ! empty($config['delegators'])) {
```
dennybrandes: I think check for empty() is enough here. No need for isset.

webimpress:
@thomasvargiu It looks like a hotfix, so I think you should target master here.

thomasvargiu:
@webimpress I rewrite it on master base

test/ServiceManagerTest.php
```diff
@@ -283,4 +284,51 @@ public function testFactoryMayBeStaticMethodDescribedByCallableString()
...
+        ];
+
+        $serviceManager = new ServiceManager($config);
+        $serviceManager->configure($config);
+
+        /** @var InvokableObject $instance */
+        $instance = $serviceManager->get('Foo');
+
+        self::assertInstanceOf(InvokableObject::class, $instance);
+        self::assertSame(1, $instance->getOptions()['inc']);
```
dennybrandes: I think a single test should have only one assertion.

thomasvargiu: IMHO, I think it can be a valid assertion. Instead to throw an error in case of failure I'm asserting that the second assertion depends on the first assertion. If it's enough the phpunit error catch for you, I can remove the first assertion. Different point of view :)
Someonelse opinion?

webimpress: It's fine to have multiple assertions as long as they are connected, imho. Kinda "one logic assertion". I was checking what people say about it, and I see some approach to write test with single assertion. I see the advantages of it, but I think this approach would be more messy. You can see that in other tests usually we have more than one assertion, but these are connected - as you are saying, @thomasvargiu.

See: https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test

any my favourite comment there is:
> A single assert per unit test is a great way to test the reader's ability to scroll up and down.

🤣

weierophinney:
This repository has been closed and moved to laminas/laminas-servicemanager; a new issue has been opened at https://github.com/laminas/laminas-servicemanager/issues/3.

weierophinney:
This repository has been moved to laminas/laminas-servicemanager. If you feel that this patch is still relevant, please re-open against that repository, and reference this issue. To re-open, we suggest the following workflow:
 - Squash all commits in your branch (`git rebase -i origin/{branch}`)
 - Make a note of all changed files (`git diff --name-only origin/{branch}...HEAD
 - Run the laminas/laminas-migration tool on the code.
 - Clone laminas/laminas-servicemanager to another directory.
 - Copy the files from the second bullet point to the clone of laminas/laminas-servicemanager.
 - In your clone of laminas/laminas-servicemanager, commit the files, push to your fork, and open the new PR.
 We will be providing tooling via laminas/laminas-migration soon to help  automate the process.

